PR #65, opened 2023-05-24
Use `nvidia-container-runtime-hook` install path from runtime config

chiragjn:
- Set it during toolkit installation
- Read from it during runtime invocation to install the hook with this path

TODO:
- [ ] Fix and update tests


elezar:
Thanks for the contribution @chiragjn. I also created https://gitlab.com/nvidia/container-toolkit/container-toolkit/-/merge_requests/401 to add this functionalty shortly after I saw your issue this morning.

Note that we use GitLab to do our development, and as such the PR should be closed and an MR opened. It would be greate if you could look at the MR that I created and give any feedback you may have there.

chiragjn:
Thank you, closing this

Issue #143, opened 2023-11-05
CDI with podman when whole `/dev` is mounted fails on `nvidia-ctk hook chmod --mode "755" --path /dev/dri` hook


Jmennius:
Hello,

I'm trying to use (implement?) `toolbox` with nvidia GPUs via CDI.
`toolbox` is essentially specifically configured `podman` containers, in particular, they have `/dev` mounted.

The issue I am having can be easily reproduced with `podman`:
```
podman run --rm --device=nvidia.com/gpu=all --security-opt=label=disable --userns keep-id --privileged --user root:root --volume /dev:/dev:rslave ubuntu:latest nvidia-smi -L
Error: OCI runtime error: crun: {"msg":"error executing hook `/usr/bin/nvidia-ctk` (exit code: 1)","level":"error","time":"2023-11-05T12:05:27.216981Z"}
```
I have CDI spec generated with `sudo nvidia-ctk cdi generate --output=/etc/cdi/nvidia.yaml`.

I was able to trace back the issue to specific `createContainer` hook - the one which `chmod`s `/dev/dri`: `nvidia-ctk hook chmod --mode "755" --path /dev/dri`).
If I remove the hook - it works as expected.

Is there anything we can do at the hook level to improve the behavior in this scenario?

Thanks!

elezar:
That hook was added to workaround a bug in crun specifically for rootless containers. It seems that running mounting /dev into the container is causing this to fail 

We could definitely make this more robust on the hooks side. Is that something that you will be willing to spend some time on?

elezar:
A little out of scope for this issue, but since you're adding mounts etc to your command line anyway a longer term option may be for toolbox to generate it's own CDI spec and then request that CDI device from podman. We have a go API (still a little fragile) that can be used on the NVIDIA side.

Jmennius:
> That hook was added to workaround a bug in crun specifically for rootless containers. It seems that running mounting /dev into the container is causing this to fail
> 
> We could definitely make this more robust on the hooks side. Is that something that you will be willing to spend some time on?

Yes, I would, with some guidance of course :)
Which direction should we take? Check if directory already has required permissions or ignore errors maybe?

> A little out of scope for this issue, but since you're adding mounts etc to your command line anyway a longer term option may be for toolbox to generate it's own CDI spec and then request that CDI device from podman. We have a go API (still a little fragile) that can be used on the NVIDIA side.

[...]

PR #158, opened 2023-11-19
nvidia-ctk hook chmod: Improve permission error handling

Jmennius:
Sometimes setting file mode may fail although the system would work fine as is (see #143).

To avoid as much trouble as possible:
* do not `chmod()` paths that already have the correct mode
* ignore permission errors of `chmod()`

### DIFF
--- a/cmd/nvidia-ctk/hook/chmod/chmod.go
+++ b/cmd/nvidia-ctk/hook/chmod/chmod.go
@@ -17,16 +17,15 @@
 package chmod
 
 import (
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
-	"syscall"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
-	"github.com/NVIDIA/nvidia-container-toolkit/internal/lookup"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/oci"
 	"github.com/urfave/cli/v2"
 )
@@ -126,17 +125,16 @@ func (m command) run(c *cli.Context, cfg *config) error {
 		return nil
 	}
 
-	locator := lookup.NewExecutableLocator(m.logger, "")
-	targets, err := locator.Locate("chmod")
-	if err != nil {
-		return fmt.Errorf("failed to locate chmod: %v", err)
+	for _, path := range paths {
+		err = os.Chmod(path, desiredMode)
+		// in some cases this is not an issue (e.g. whole /dev mounted), see #143
+		if errors.Is(err, fs.ErrPermission) {
+			m.logger.Debugf("Ignoring permission error with chmod: %v", err)
+			err = nil
+		}
 	}
-	chmodPath := targets[0]
 
-	args := append([]string{filepath.Base(chmodPath), cfg.mode}, paths...)
-
-	//nolint:gosec // TODO: Can we harden this so that there is less risk of command injection
-	return syscall.Exec(chmodPath, args, nil)
+	return err
 }
 
 // getPaths updates the specified paths relative to the root.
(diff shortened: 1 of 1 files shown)


cmd/nvidia-ctk/hook/chmod/chmod.go
```diff
@@ -112,34 +113,44 @@ func (m command) run(c *cli.Context, cfg *config) error {
 		return fmt.Errorf("empty container root detected")
 	}
 
-	paths := m.getPaths(containerRoot, cfg.paths.Value())
+	desiredModeInt, err := strconv.ParseUint(cfg.mode, 8, 32)
```
elezar, line 116: Could we check this in `valdidateFlags` instead?

cmd/nvidia-ctk/hook/chmod/chmod.go
```diff
@@ -126,17 +125,16 @@ func (m command) run(c *cli.Context, cfg *config) error {
 		return nil
 	}
 
-	locator := lookup.NewExecutableLocator(m.logger, "")
-	targets, err := locator.Locate("chmod")
-	if err != nil {
-		return fmt.Errorf("failed to locate chmod: %v", err)
+	for _, path := range paths {
+		err = os.Chmod(path, desiredMode)
```
elezar, line 129: Just for my own information: What is the difference between this and `Exec`ing to `chmod`?

cmd/nvidia-ctk/hook/chmod/chmod.go
Jmennius: I was thinking about this, but we also need the parsed value here - so we either need to save it after parsing or parse twice.

cmd/nvidia-ctk/hook/chmod/chmod.go
Jmennius: I've tried to explain it in the commit message:

> Due to loosing control of the program after `exec()`-ing `chmod(1)` program
> and therefore not being able to handle errors -
> refactor to use `chmod(2)` syscall instead of `exec()` `chmod(1)` program.

Essentially, `nvidia-ctk` code stops executing and control is transferred fully to `chmod(1)` which will `exit(1)` once an error occurs.

Jmennius:
> Thanks for the contribution.
> 
> (Note that this should ideally be created as an MR against our GitLab repo: https://gitlab.com/nvidia/container-toolkit/container-toolkit).
> 
> I'm happy to pull the PR over after we finalize it here though.

I can open MR on Gitlab, no problem :)

Jmennius:
Opened https://gitlab.com/nvidia/container-toolkit/container-toolkit/-/merge_requests/496

cmd/nvidia-ctk/hook/chmod/chmod.go
elezar: Yes, I'm aware of that. I meant if there was any difference in behaviour between the different `chmods`? Assuming that they were each called with a single path?

Jmennius: chmod(1) does some checks and so on, but for our use-case it looks to be effectively the same.. or at least I hope so.
One thing I can think of: since 'mode' string was used as is - it could be in any form `chmod(1)` accepts (i.e. symbolic); but since we did not use that - should not be a problem.

cmd/nvidia-ctk/hook/chmod/chmod.go
Jmennius: I've pushed a commit where I've moved it into `valdidateFlags` - saving the result in additional cfg field. We can leave it or drop it - up to you :)

Jmennius:
Moved [to GitLab](https://gitlab.com/nvidia/container-toolkit/container-toolkit/-/merge_requests/496).

PR #307, opened 2024-01-28
Add GitHub Actions

ArangoGutierrez:


.gitlab-ci.yml
```diff
@@ -15,42 +15,8 @@
...
-unit-tests:
-  extends:
-    - .requires-build-image
-  stage: unit-tests
-  script:
-    - make coverage
-
 # Define the package build helpers
-.multi-arch-build:
+  .multi-arch-build:
```
elezar, line 19: Why was this whitespace inserted?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,215 @@
...
+    branches:
+      - main
+      - release-*
+  push:
+    branches:
+      - main
+      - release-*
+
+jobs:
+  ubuntu-packages:
```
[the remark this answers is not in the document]
elezar: It would also be good to skip some of these on PRs.

elezar (commented):
I would rather keep this transition "go-only" and rely on the internal pipelines to start with.

.github/workflows/image.yaml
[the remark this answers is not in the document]
ArangoGutierrez: I agree, it sounds like a lot of packaging (pun intended) .

Let's move all of these to `push` event. `pull_request` should be about checks and tests, not packaging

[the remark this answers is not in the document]
ArangoGutierrez: I have moved the package/image building to `push` event, and added the matrix as requested
PTAL

.github/workflows/image.yaml
```diff
@@ -0,0 +1,236 @@
...
+name: image
+
+on:
+  push:
+    branches:
+      - main
+      - release-*
+
+jobs:
+  ubuntu-packages:
```
elezar, line 25: Should just be `packages` an all `targets` can be specified. There is no reason to separate `centos` and `ubuntu`.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,236 @@
...
+      matrix:
+        arch: [amd64, arm64, ppc64le]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+        name: Check out code
+      - name: Calculate build vars
+        id: vars
+        run: |
+          echo "DIST_DIR=${{ github.workspace }}/toolkit-container-ubuntu18.04-${{ github.run_id }}" >> $GITHUB_ENV
```
elezar, line 37: If we're not using complex logic, we can just specify these directly.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,236 @@
...
+        with:
+          compression-level: 0
+          name: toolkit-container-centos7-${{ github.run_id }}
+          path: ${{ github.workspace }}/toolkit-container-centos7-${{ github.run_id }}/centos7/${{ matrix.dist }}
+
+  centos8-packages:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: [x86_64, aarch64, ppc64le]
```
elezar, line 97: `centos8` is only `ppc64le`.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,236 @@
...
+          echo "Building packages"
+          ./scripts/build-packages.sh centos8-${{ matrix.dist }}
+      - name: 'Upload Artifacts'
+        uses: actions/upload-artifact@v4
+        with:
+          compression-level: 0       
+          name: toolkit-container-centos7-${{ github.run_id }}
+          path: ${{ github.workspace }}/toolkit-container-centos8-${{ github.run_id }}/centos8/${{ matrix.dist }}
+
+  ubuntu20_04:
```
elezar, line 127:
```suggestion
  image:
```
?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,101 @@
...
+        id: vars
+        run: |
+          echo "COMMIT_SHORT_SHA=${GITHUB_SHA:0:8}" >> $GITHUB_ENV
+          echo "LOWERCASE_REPO_OWNER=$(echo "${GITHUB_REPOSITORY_OWNER}" | awk '{print tolower($0)}')" >> $GITHUB_ENV
+          REPO_FULL_NAME="${{ github.event.pull_request.head.repo.full_name }}"
+          echo "${REPO_FULL_NAME}"
+          echo "LABEL_IMAGE_SOURCE=https://github.com/${REPO_FULL_NAME}" >> $GITHUB_ENV
+
+          echo "PUSH_ON_BUILD=true" >> $GITHUB_ENV
+          echo "BUILD_MULTI_ARCH_IMAGES=true" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
ArangoGutierrez:
```bash
        GENERATE_ARTIFACTS="false"
          if [[ "${{ github.actor }}" == "dependabot[bot]" ]]; then
            GENERATE_ARTIFACTS="false"
          elif [[ "${{ github.event_name }}" == "pull_request" && "${{ github.event.pull_request.head.repo.full_name }}" == "${{ github.repository }}" ]]; then
            GENERATE_ARTIFACTS="true"
          elif [[ "${{ github.event_name }}" == "push" ]]; then
            GENERATE_ARTIFACTS="true"
          fi
```

since we moved the job to merge event `push`, the result will always be the same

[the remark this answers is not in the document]
elezar: Yes, I understand that. From my perspective it makes sense to keep these as similar as possible so as to minimize the drift between projects. This makes updating things later simpler.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,101 @@
...
+        with:
+          compression-level: 0
+          name: toolkit-container-${{ matrix.package }}-${{ github.run_id }}
+          path: ${{ github.workspace }}/dist/*
+
+  image:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: 
+        dist: [ubuntu20.04, centos7, ubi8]
```
elezar, line 62: there is also a `packaging` image that includes all the packages built in previous steps:
```suggestion
        dist:
        - ubuntu20.04
        - centos7
        - ubi8
        - packaging
```

See https://github.com/NVIDIA/nvidia-container-toolkit/blob/03cbf9c6cd26c75afef8a2dd68e0306aace80401/.gitlab-ci.yml#L207

.github/workflows/image.yaml
```diff
@@ -0,0 +1,101 @@
...
+# Run this workflow on pull requests
+name: image
+
+on:
+  push:
+    branches:
+      - main
+      - release-*
+    tags:        
+      - 'v*'
```
elezar, line 24: On second thought, let's remove this for now and only include it once we're actually pushing images from here.
```suggestion
```

.github/workflows/image.yaml
```diff
@@ -0,0 +1,108 @@
...
+          echo "LABEL_IMAGE_SOURCE=https://github.com/${REPO_FULL_NAME}" >> $GITHUB_ENV
+          
+          GENERATE_ARTIFACTS="false"
+            if [[ "${{ github.actor }}" == "dependabot[bot]" ]]; then
+              GENERATE_ARTIFACTS="false"
+            elif [[ "${{ github.event_name }}" == "pull_request" && "${{ github.event.pull_request.head.repo.full_name }}" == "${{ github.repository }}" ]]; then
+              GENERATE_ARTIFACTS="true"
+            elif [[ "${{ github.event_name }}" == "push" ]]; then
+              GENERATE_ARTIFACTS="true"
+            fi
```
elezar, line 81: The intentation is messed up.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,108 @@
...
+        with:
+          compression-level: 0
+          name: toolkit-container-${{ matrix.package }}-${{ github.run_id }}
+          path: ${{ github.workspace }}/dist/*
+
+  image:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: 
+        image-packaging: [ubuntu20.04, centos7, ubi8, packaging]
```
elezar, line 60: `image` or `dist` should be preferred here. Note that only the `build-packaging` target generates a "packaging" image. The other targets generate distribution-specific images for deploying the `contianer-toolkit` as an operand used in the GPU Operator.

elezar (commented):
I think my comments are mostly nits. 

One thing: Have we actually triggered this workflow. Does it make sense to enable this for PRs *once* so that we can verify that everything works as expected *before* merging?

.github/workflows/image.yaml
ArangoGutierrez: ummm yeah weird.... fixing...

.github/workflows/image.yaml
ArangoGutierrez: edited to `image`

ArangoGutierrez:
> One thing: Have we actually triggered this workflow. Does it make sense to enable this for PRs _once_ so that we can verify that everything works as expected _before_ merging?

This could be a followup.

elezar (approved):
Sure. Let's start with this and iterate.

ArangoGutierrez:
Mirror direction has been switched

PR #320, opened 2024-01-30
Remove centos8 jobs

elezar:
This change removes the centos8-x86_64 and centos8-aarch64 pipeline jobs.

These packages are no longer used since centos7 packages are used instead.

### DIFF
--- a/test/release/docker/centos8/Dockerfile
+++ b/test/release/docker/centos8/Dockerfile
@@ -25,7 +25,7 @@ RUN fpm -s empty \
     rm -f /tmp/docker.rpm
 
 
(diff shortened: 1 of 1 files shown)


ArangoGutierrez:
Should we remove also https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/.github/workflows/image.yaml#L35 ?

elezar:
> Should we remove also https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/.github/workflows/image.yaml#L35 ?

No. That is the `ppc64le` target. This PR removes the `x86_64` and `aarch64` jobs. I will update the description to make this clearer.

ArangoGutierrez (approved):

PR #321, opened 2024-01-30
Update libnvidia-container

elezar:


elezar:
This is required to address the build failures on `main` since `libnvidia-container` was updated.

ArangoGutierrez (approved):
Fixes https://github.com/NVIDIA/nvidia-container-toolkit/actions/runs/7713175260/job/21022465795

PR #326, opened 2024-02-01
Create .so symlinks for driver libraries in container

elezar:
This change explicitly creates .so and SONAME symlinks for injected driver libraries. This means that the `create-soname-symlinks` hook that was added in #947 is not longer required and is therefore removed.

```
$ docker run --rm -ti --runtime=nvidia --gpus all busybox  ls -al /usr/lib/x86_64-linux-gnu/
total 263564
drwxr-xr-x    2 root     root          4096 Jul 21 10:00 .
drwxr-xr-x    3 root     root          4096 Jul 21 10:00 ..
lrwxrwxrwx    1 root     root            12 Jul 21 10:00 libcuda.so -> libcuda.so.1
lrwxrwxrwx    1 root     root            21 Jul 21 10:00 libcuda.so.1 -> libcuda.so.570.133.20
-rw-r--r--    1 root     root      71365624 Apr 13 04:46 libcuda.so.570.133.20
lrwxrwxrwx    1 root     root            20 Jul 21 10:00 libcudadebugger.so -> libcudadebugger.so.1
lrwxrwxrwx    1 root     root            29 Jul 21 10:00 libcudadebugger.so.1 -> libcudadebugger.so.570.133.20
-rw-r--r--    1 root     root      10244640 Apr 13 04:17 libcudadebugger.so.570.133.20
lrwxrwxrwx    1 root     root            17 Jul 21 10:00 libnvidia-ml.so -> libnvidia-ml.so.1
lrwxrwxrwx    1 root     root            26 Jul 21 10:00 libnvidia-ml.so.1 -> libnvidia-ml.so.570.133.20
-rw-r--r--    1 root     root       2217912 Apr 13 04:23 libnvidia-ml.so.570.133.20
lrwxrwxrwx    1 root     root            19 Jul 21 10:00 libnvidia-nvvm.so -> libnvidia-nvvm.so.4
lrwxrwxrwx    1 root     root            28 Jul 21 10:00 libnvidia-nvvm.so.4 -> libnvidia-nvvm.so.570.133.20
-rw-r--r--    1 root     root      81978912 Apr 13 05:03 libnvidia-nvvm.so.570.133.20
lrwxrwxrwx    1 root     root            21 Jul 21 10:00 libnvidia-opencl.so -> libnvidia-opencl.so.1
lrwxrwxrwx    1 root     root            30 Jul 21 10:00 libnvidia-opencl.so.1 -> libnvidia-opencl.so.570.133.20
-rw-r--r--    1 root     root      65758768 Apr 13 04:46 libnvidia-opencl.so.570.133.20
-rw-r--r--    1 root     root         10176 Apr 13 04:21 libnvidia-pkcs11-openssl3.so.570.133.20
-rw-r--r--    1 root     root         10168 Apr 13 04:21 libnvidia-pkcs11.so.570.133.20
lrwxrwxrwx    1 root     root            29 Jul 21 10:00 libnvidia-ptxjitcompiler.so -> libnvidia-ptxjitcompiler.so.1
lrwxrwxrwx    1 root     root            38 Jul 21 10:00 libnvidia-ptxjitcompiler.so.1 -> libnvidia-ptxjitcompiler.so.570.133.20
-rw-r--r--    1 root     root      38251952 Apr 13 04:33 libnvidia-ptxjitcompiler.so.570.133.20
```

And:
```
$ docker run --rm -ti --runtime=nvidia --gpus all ubuntu ldconfig -p | grep libcuda
        libcudadebugger.so.1 (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libcudadebugger.so.1
        libcudadebugger.so (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libcudadebugger.so
        libcuda.so.1 (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libcuda.so.1
        libcuda.so (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libcuda.so
```
showing that the .so and .so.1 entries are present in the ldcache.


elezar (commented):
I think some of the issues were because this was out of date.

internal/config/features.go
```diff
@@ -31,6 +36,9 @@ type features struct {
 	MOFED    *feature `toml:"mofed,omitempty"`
 	NVSWITCH *feature `toml:"nvswitch,omitempty"`
 	GDRCopy  *feature `toml:"gdrcopy,omitempty"`
+	// DotSoSymlinks allows for the creation of .so symlinks to .so.1 driver
+	// files to be opted out of.
```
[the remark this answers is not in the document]
elezar: You're right. I initially had this as "on-by-default" with opt-out logic, but rolled that back and didn't update the comment. Will update.

ArangoGutierrez:
@elezar does #906 supersede this one?

elezar:
> @elezar does #906 supersede this one?

No. This is not related to the CUDA Forward Compat libraries at all. This is about creating `.so` -> `SONAME` symlinks for driver libraries in the container. #906 is about ensuring that the CUDA Forward compat libraries are properly detectable in a container if required.

cmd/nvidia-ctk/info/info.go
```diff
@@ -40,7 +44,27 @@ func (m command) build() *cli.Command {
...
+	}
+	defer lib.Close()
+
+	sonames, err := lib.DynString(elf.DT_SONAME)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("sonames=%v\n", sonames)
+	return nil
+}
```
klueska: What is going on here? What is this hard-coded
```
filename := "/usr/lib/x86_64-linux-gnu/libcuda.so.570.133.20"
```

elezar: Ah ... sorry ... this was me testing the behaviour on an actual driver. It should not have been included in the PR.

internal/lookup/root/root.go
```diff
@@ -53,6 +62,50 @@ func (r *Driver) Libraries() lookup.Locator {
...
+func (r *Driver) Version() (string, error) {
+	r.Lock()
+	defer r.Unlock()
+	if r.version != "" {
+		return r.version, nil
+	}
+
+	libCudaPaths, err := cuda.New(
+		r.Libraries(),
+	).Locate(".*.*")
```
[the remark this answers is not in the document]
elezar: No longer relevant.

pkg/nvcdi/driver-nvml.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/lookup/root"
 )
 
-// NewDriverDiscoverer creates a discoverer for the libraries and binaries associated with a driver installation.
+// newDriverDiscoverer creates a discoverer for the libraries and binaries associated with a driver installation.
```
klueska: mistake?

internal/discover/symlinks.go
```diff
@@ -110,3 +130,50 @@ func (d additionalSymlinks) isDriverLibrary(libraryName string, filename string)
...
+
+	if sonameLinkPath != "" {
+		sonameLinkPathExt := filepath.Ext(sonameLinkPath)
+		soLinkPath := strings.TrimSuffix(sonameLinkPath, sonameLinkPathExt)
+		s := Symlink{
+			target: filepath.Base(sonameLinkPath),
+			link:   soLinkPath,
+		}
+		soSymlinks = append(soSymlinks, s.String())
+	}
```
klueska: So you walk through all sonames, and add a symlik for them.

And at the same time you see if there was at least one soname that matched, and if there was, you store it in `sonameLinkPath` so that you can strip off the final version number from it and just additionally include the base `.so` symlink.

Is that what is going on here?

klueska: Actually, I'm confused as to why there is a loop over *multiple* sonames to begin with -- I thought you could only have one per library.

elezar: Yes, for a given library, we extract the sonames and create symlinks for them (if they don't match the original library name). For the first soname we additionally strip off the numbered suffix and create a `.so` symlink too. 

One thing that we may want to do is explicitly match on `*.so` instead of making assumptions about the soname that we're linking to.

klueska: Meaning, would it make more sense to implement this as:
```
	sonames, err := lib.DynString(elf.DT_SONAME)
	if err != nil {
		return nil, err
	}

	if len(sonames) != 1 {
		return fmt.Errorf(...)
	}

	// Create the <library>.so.<version> symlink from SONAME
	linkPath := filepath.Join(filepath.Dir(libraryPath), soname[0])
	if soname[0] != libraryName {
		s := Symlink{
			target: libraryName,
			link:   linkPath,
		}
		soSymlinks = append(soSymlinks, s.String())
	}
	
	// Create the <library>.so symlink from the <library>.so.<version> symlink
	soLinkPath := strings.TrimSuffix(linkPath, filepath.Ext(linkPath))
	s := Symlink{
		target: filepath.Base(linkPath),
		link:   soLinkPath,
	}
	soSymlinks = append(soSymlinks, s.String())
```

klueska: I don't understand this part:
>  For the first soname we additionally strip off the numbered suffix and create a .so symlink too

Is it actually possible to have multiple SONAMEs set?

elezar: The [`elf.h` manpage](https://man7.org/linux/man-pages/man5/elf.5.html) states:

>               DT_SONAME
                     String table offset to name of shared object

Which refers to a singular. The issue is that the `DynString` function returns a slice to handle other case. I would therefore assume that it's unexpected if we have MULTIPLE sonames. Let me update the implementation to check the length specifically.

pkg/nvcdi/driver-nvml.go
elezar: Yes. Should not have been in this change set.

internal/discover/symlinks_test.go
```diff
@@ -334,3 +334,118 @@ func TestWithWithDriverDotSoSymlinks(t *testing.T) {
...
+				return "libcuda.so.1", nil
+			},
+			expectedError: nil,
+			expectedSymlinks: []string{
+				"libcuda.so.999.88.77::/usr/lib/libcuda.so.1",
+				"libcuda.so.1::/usr/lib/libcuda.so",
+			},
+		},
+		{
+			description: "empty soname is ingored",
```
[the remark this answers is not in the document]
ArangoGutierrez: @elezar ++

internal/discover/symlinks_test.go
```diff
@@ -334,3 +334,118 @@ func TestWithWithDriverDotSoSymlinks(t *testing.T) {
...
+			soLink := getSoLink(tc.input)
+
+			require.Equal(t, tc.expectedSoLink, soLink)
+		})
+	}
+}
+
+func setGetSoname(ovverride func(string) (string, error)) func() {
+	original := getSoname
+	getSoname = ovverride
```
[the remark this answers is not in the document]
ArangoGutierrez: ++

elezar:
> docker run --rm -ti --runtime=nvidia --gpus all busybox  ls -al /usr/lib/x86_64-linux-gnu/

Added in latest.

tests/e2e/nvidia-container-toolkit_test.go
ArangoGutierrez, line 1: LGTM

internal/discover/symlinks.go
```diff
@@ -60,7 +71,14 @@ func (d *additionalSymlinks) Hooks() ([]Hook, error) {
 		}
 		processedPaths[mount.Path] = true
 
-		for _, link := range d.getLinksForMount(mount.Path) {
+		linksForMount := d.getLinksForMount(mount.Path)
+		soSymlinks, err := d.getDotSoSymlinks(mount.HostPath)
+		if err != nil {
+			soSymlinks = nil
+		}
```
klueska: Why would soSymlinks be anything other than `nil` if this function returns an error? By conventions we shouldn't return a value if there is an error, it should be `nil` already.

internal/discover/symlinks.go
```diff
@@ -60,7 +71,14 @@ func (d *additionalSymlinks) Hooks() ([]Hook, error) {
 		}
 		processedPaths[mount.Path] = true
 
-		for _, link := range d.getLinksForMount(mount.Path) {
+		linksForMount := d.getLinksForMount(mount.Path)
+		soSymlinks, err := d.getDotSoSymlinks(mount.HostPath)
+		if err != nil {
```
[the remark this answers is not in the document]
klueska: I agree (with the sentiment, not with using a printf here). Why are we ignoring the error?

internal/discover/symlinks.go
```diff
@@ -110,3 +128,73 @@ func (d additionalSymlinks) isDriverLibrary(libraryName string, filename string)
...
+		return nil, nil
+	}
+
+	soname, err := getSoname(libraryPath)
+	if err != nil {
+		return nil, err
+	}
+	if soname == "" || soname == libraryName {
+		return nil, nil
+	}
```
klueska: Do we not still want the `.so` link even if the SONAME is the same as the libraryName?

internal/discover/symlinks.go
```diff
@@ -110,3 +128,73 @@ func (d additionalSymlinks) isDriverLibrary(libraryName string, filename string)
...
+	sonameLinkPath := filepath.Join(libraryDir, soname)
+	s := Symlink{
+		target: libraryName,
+		link:   sonameLinkPath,
+	}
+	soSymlinks = append(soSymlinks, s.String())
+
+	// Create the .so -> SONAME symlink
+	soLink := getSoLink(soname)
+	if soLink != "" && soLink != soname && soLink != libraryName {
```
klueska: I don't understand the need for all these special cases

tests/e2e/Makefile
```diff
@@ -23,6 +23,14 @@ GINKGO_BIN := $(CURDIR)/bin/ginkgo
 test: $(GINKGO_BIN)
 	$(GINKGO_BIN) $(GINKGO_ARGS) -v --json-report ginkgo.json ./tests/e2e/...
 
+# test-reinstalled runs the test cases against the version of the toolkit that
```
ArangoGutierrez:
```suggestion
# test-preinstalled runs the test cases against the version of the toolkit that
```

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,6 +261,91 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+				}
+
+				// Keep track of the symlink chain.
+				symlinks[soSymlink] = append(symlinks[soSymlink], parts[1:]...)
+
+				// We assume a regular file is the end of a symlink chain.
+				if strings.HasPrefix(permissionString, "-") {
+					soSymlink = ""
+				}
+
```
ArangoGutierrez:
```suggestion
```

internal/discover/symlinks.go
elezar: Yes, I think you're right. We want the `.so` link but not the soname link. I don't think this comes up in typical driver installations though, so we may want to skip this for now.

internal/discover/symlinks.go
elezar: You're right it will always be nil in our implementation. I was being overly cautious.

internal/discover/symlinks.go
elezar: If we get `""` this means that `soname` is not a `.so.*` file which one could argue is unexpected since we do filter on driver libraries at the start of this function. The other cases make sure that we don't try to create a link that links to itself. I think these are unlikely to occur in our driver installations, but are there for completenes.

internal/discover/symlinks.go
```diff
@@ -110,3 +132,70 @@ func (d additionalSymlinks) isDriverLibrary(libraryName string, filename string)
...
+	sonames, err := lib.DynString(elf.DT_SONAME)
+	if err != nil {
+		return "", err
+	}
+	if len(sonames) > 1 {
+		return "", fmt.Errorf("multiple SONAMEs detected for %v: %v", libraryPath, sonames)
+	}
+	if len(sonames) == 0 {
+		return filepath.Base(libraryPath), nil
+	}
```
klueska, line 227: So if no soname is set in the binary, we just assume it to be the same as the library name itself? That seems reasonable -- do we actually have cases where this happens or is this just being defensive?

internal/discover/symlinks.go
```diff
@@ -110,3 +132,70 @@ func (d additionalSymlinks) isDriverLibrary(libraryName string, filename string)
...
+
+	// Create the .so -> SONAME symlink
+	if soLink := getSoLink(soname); soLink != "" && soLink != soname {
+		s := Symlink{
+			target: soname,
+			link:   filepath.Join(libraryDir, soLink),
+		}
+		soSymlinks = append(soSymlinks, s.String())
+	}
```
klueska: I still think this could be a bit more readable (in terms of the conditional cases). Maybe just a comment on what the cases are / mean would be sufficient, e.g.:
```
// Create the .so -> SONAME symlink
// If soname does not have a '.so' extension or soname itself ends with '.so'
// we skip this (because we will have just created the same symlink above).
```

klueska:
A couple of minor comments, otherwise LGTM

internal/discover/symlinks.go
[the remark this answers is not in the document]
elezar: Because I didn't want to plumb the logger through. I have updated this in latest.

internal/discover/symlinks.go
elezar: The `libnvidia-pkcs11.so.550.120` library is an example of one that doesn't have a proper soname. The driver installation also doesn't include symlinks for this library.

internal/discover/symlinks.go
klueska: That's fine then. We can always revisit this if anyone complains.

klueska:
Looking good. Thanks.

klueska (approved):

ArangoGutierrez (approved):
LGTM

PR #327, opened 2024-02-01
Add config search path option to driver root

elezar:
This changes adds a `Configs()` method to the Driver root abstraction that can be used to locate config files.

In addition a `--config-search-path` option is added to the `nvidia-ctk cdi generate` command to allow this to be overridden.

### DIFF
--- a/cmd/nvidia-ctk/cdi/generate/generate.go
+++ b/cmd/nvidia-ctk/cdi/generate/generate.go
@@ -53,6 +53,7 @@ type options struct {
 	vendor               string
 	class                string
 
+	configSearchPaths  cli.StringSlice
 	librarySearchPaths cli.StringSlice
 
 	csv struct {
@@ -86,6 +87,11 @@ func (m command) build() *cli.Command {
 	}
 
 	c.Flags = []cli.Flag{
+		&cli.StringSliceFlag{
+			Name:        "config-search-path",
+			Usage:       "Specify the path to search for config files when discovering the entities that should be included in the CDI specification.",
+			Destination: &opts.configSearchPaths,
+		},
 		&cli.StringFlag{
 			Name:        "output",
 			Usage:       "Specify the file to output the generated CDI specification to. If this is '' the specification is output to STDOUT",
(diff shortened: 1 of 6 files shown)


internal/lookup/root/root.go
```diff
@@ -53,6 +55,24 @@ func (r *Driver) Libraries() lookup.Locator {
...
 }
 
+// Configs returns a locator for driver configs.
+// If configSearchPaths is specified, these paths are used as absolute paths,
+// otherwise, /etc and /usr/share are searched.
+func (r *Driver) Configs() lookup.Locator {
+	useRoot := r.Root
+	searchPaths := []string{"/etc", "/usr/share"}
+	if len(r.configSearchPaths) > 0 {
+		useRoot = ""
```
tariq1890, line 65: Instead of setting this as an empty string, I would prefer if we added the with `.WithRoot()` to the Options list when configSearchPaths is not set

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -86,6 +87,11 @@ func (m command) build() *cli.Command {
 	}
 
 	c.Flags = []cli.Flag{
+		&cli.StringSliceFlag{
+			Name:        "config-search-path",
+			Usage:       "Specify the path to search for config files when discovering the entities that should be included in the CDI specification.",
```
tariq1890, line 92: nit: "config files" is a bit vague here. It would be helpful if we specified what configuration files they are

internal/lookup/root/root.go
elezar: I took a stab at reworking this as:
```
diff --git a/internal/lookup/root/root.go b/internal/lookup/root/root.go
index bfd3ace4..f3176744 100644
--- a/internal/lookup/root/root.go
+++ b/internal/lookup/root/root.go
@@ -59,17 +59,22 @@ func (r *Driver) Libraries() lookup.Locator {
 // If configSearchPaths is specified, these paths are used as absolute paths,
 // otherwise, /etc and /usr/share are searched.
 func (r *Driver) Configs() lookup.Locator {
-	useRoot := r.Root
-	searchPaths := []string{"/etc", "/usr/share"}
-	if len(r.configSearchPaths) > 0 {
-		useRoot = ""
-		searchPaths = normalizeSearchPaths(r.configSearchPaths...)
+	options := []lookup.Option{
+		lookup.WithLogger(r.logger),
 	}
 
+	if len(r.configSearchPaths) == 0 {
+		options = append(options,
+			lookup.WithRoot(r.Root),
+			lookup.WithSearchPaths("/etc", "/usr/share"),
+		)
+	} else {
+		options = append(options,
+			lookup.WithSearchPaths(normalizeSearchPaths(r.configSearchPaths...)...),
+		)
+	}
 	return lookup.NewFileLocator(
-		lookup.WithLogger(r.logger),
-		lookup.WithRoot(useRoot),
-		lookup.WithSearchPaths(searchPaths...),
+		options...,
 	)
 }
 
```
but I find this more verbose and unclear as to what is actually being done here. If you don't mind, I will leave this as is -- possibly renaming some local variables.

cmd/nvidia-ctk/cdi/generate/generate.go
tariq1890: nvm, I guess cdi config files is more or less implied

PR #330, opened 2024-02-02
Support nvidia and nvidia-frontend names when getting device major

### LINKED ISSUE #324: Add support for >= 550.40.x drivers to the toolkit container
As part of generating management CDI specifications, the container-toolkit container creates control device nodes. This requires reading the device major number from `/proc/devices`. This fails in drivers from 550.40.x since the `nvidia-frontend` module was renamed `nvidia`.

See: https://www.nvidia.de/Download/driverResults.aspx/218124/en-us

> Changed the name visible in /proc/devices of NVIDIA devices and theNVIDIA control device from "nvidia-frontend" to "nvidia" and "nvidiactl".Scripts which parse /proc/devices (such as udev rules) may need to beupdated. Note that the conventional /dev device paths like /dev/nvidia0and /dev/nvidiactl remain unchanged.

This can be triggered by starting the toolkit container with CDI enabled to trigger the code path but will also affect the `nvidia-ctk system create-dev-char-symlinks` command.

Note that `nvidia-modprobe` hardcodes the major number: https://github.com/NVIDIA/nvidia-modprobe/blob/d6bce304f30b6661c9ab6a993f49340eafca7a7e/modprobe-utils/nvidia-modprobe-utils.c#L85

We should probably do the same as this will ensure that we remain compatible across driver versions.

tariq1890:
This PR fixes #324

It checks the NVIDIA Devices struct constructed from `/proc/devices` for both `nvidia` and `nvidia-frontend` to retrieve the major number.

This enables support for newer driver versions ( >= 550.40 drivers) where the module was renamed while still allowing this to detect the kernel module not being loaded.

### DIFF
--- a/internal/system/nvdevices/devices_test.go
+++ b/internal/system/nvdevices/devices_test.go
@@ -31,11 +31,25 @@ func TestCreateControlDevices(t *testing.T) {
 
 	nvidiaDevices := &devices.DevicesMock{
 		GetFunc: func(name devices.Name) (devices.Major, bool) {
-			devices := map[devices.Name]devices.Major{
+			devs := map[devices.Name]devices.Major{
 				"nvidia-frontend": 195,
 				"nvidia-uvm":      243,
 			}
-			return devices[name], true
+
+			// devs550_40 represents the device map from the nvidia gpu drivers >= 550.40.x
+			devs550_40 := map[devices.Name]devices.Major{
+				"nvidia":     195,
+				"nvidia-uvm": 243,
+			}
+
+			d, ok := devs[name]
+			if ok {
+				return d, ok
+			}
+
+			// if device d is not found, fallback to the second mock device map
+			d, ok = devs550_40[name]
+			return d, ok
 		},
 	}
 
@@ -46,6 +60,7 @@ func TestCreateControlDevices(t *testing.T) {
 		root          string
 		devices       devices.Devices
 		mknodeError   error
+		hasError      bool
 		expectedError error
 		expectedCalls []struct {
 			S  string
@@ -58,6 +73,7 @@ func TestCreateControlDevices(t *testing.T) {
 			root:        "",
 			devices:     nvidiaDevices,
 			mknodeError: nil,
+			hasError:    false,
 			expectedCalls: []struct {
 				S  string
 				N1 int
@@ -73,6 +89,7 @@ func TestCreateControlDevices(t *testing.T) {
 			description: "some root specified",
 			root:        "/some/root",
 			devices:     nvidiaDevices,
+			hasError:    false,
 			mknodeError: nil,
 			expectedCalls: []struct {
 				S  string
@@ -88,6 +105,7 @@ func TestCreateControlDevices(t *testing.T) {
 		{
 			description:   "mknod error returns error",
 			devices:       nvidiaDevices,
+			hasError:      true,
 			mknodeError:   mknodeError,
 			expectedError: mknodeError,
 			// We expect the first call to this to fail, and the rest to be skipped
@@ -106,7 +124,23 @@ func TestCreateControlDevices(t *testing.T) {
 					return 0, false
 				},
 			},
+			hasError:      true,
 			expectedError: errInvalidDeviceNode,
+		},
+		{
+			description: "nvidia device renamed from nvidia-frontend to nvidia",
+			devices:     nvidiaDevices,
+			hasError:    false,
+			expectedCalls: []struct {
+				S  string
+				N1 int
+				N2 int
+			}{
+				{"/dev/nvidiactl", 195, 255},
+				{"/dev/nvidia-modeset", 195, 254},
+				{"/dev/nvidia-uvm", 243, 0},
+				{"/dev/nvidia-uvm-tools", 243, 1},
+			},
 		},
 	}
 
@@ -126,9 +160,12 @@ func TestCreateControlDevices(t *testing.T) {
 			d.mknoder = mknode
 
 			err := d.CreateNVIDIAControlDevices()
-			require.ErrorIs(t, err, tc.expectedError)
+			if tc.hasError {
+				require.ErrorContains(t, err, tc.expectedError.Error())
+			} else {
+				require.Nil(t, err)
+			}
 			require.EqualValues(t, tc.expectedCalls, mknode.MknodeCalls())
 		})
 	}
-
 }
--- a/internal/system/nvdevices/devices_test.go
+++ b/internal/system/nvdevices/devices_test.go
@@ -29,29 +29,18 @@ import (
 func TestCreateControlDevices(t *testing.T) {
 	logger, _ := testlog.NewNullLogger()
 
-	nvidiaDevices := &devices.DevicesMock{
-		GetFunc: func(name devices.Name) (devices.Major, bool) {
-			devs := map[devices.Name]devices.Major{
-				"nvidia-frontend": 195,
-				"nvidia-uvm":      243,
-			}
-
-			// devs550_40 represents the device map from the nvidia gpu drivers >= 550.40.x
-			devs550_40 := map[devices.Name]devices.Major{
-				"nvidia":     195,
-				"nvidia-uvm": 243,
-			}
-
-			d, ok := devs[name]
-			if ok {
-				return d, ok
-			}
-
-			// if device d is not found, fallback to the second mock device map
-			d, ok = devs550_40[name]
-			return d, ok
-		},
-	}
+	nvidiaDevices := devices.New(
+		devices.WithDeviceToMajor(map[string]int{
+			"nvidia-frontend": 195,
+			"nvidia-uvm":      243,
+		}),
+	)
+	nvidia550Devices := devices.New(
+		devices.WithDeviceToMajor(map[string]int{
+			"nvidia":     195,
+			"nvidia-uvm": 243,
+		}),
+	)
 
 	mknodeError := errors.New("mknode error")
 
@@ -60,7 +49,6 @@ func TestCreateControlDevices(t *testing.T) {
 		root          string
 		devices       devices.Devices
 		mknodeError   error
-		hasError      bool
 		expectedError error
 		expectedCalls []struct {
 			S  string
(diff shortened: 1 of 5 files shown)


elezar (changes requested):
Thanks for this @tariq1890.

If we were to do the check for `nvidia` or `nvidia-frontend` in the `devices.Get` method as below:
```
diff --git a/internal/info/proc/devices/devices.go b/internal/info/proc/devices/devices.go
index 95430428..65f3108f 100644
--- a/internal/info/proc/devices/devices.go
+++ b/internal/info/proc/devices/devices.go
@@ -32,8 +32,12 @@ const (
 	NVIDIACTLMinor      = 255
 	NVIDIAModesetMinor  = 254
 
+	// NVIDIADeviceMajor is hardcoded as NV_MAJOR_DEVICE_NUMBER in nvidia-modprobe:
+	// https://github.com/NVIDIA/nvidia-modprobe/blob/d6bce304f30b6661c9ab6a993f49340eafca7a7e/modprobe-utils/nvidia-modprobe-utils.c#L85
+	NVIDIADeviceMajor = 195
+
 	NVIDIAFrontend = Name("nvidia-frontend")
-	NVIDIAGPU      = NVIDIAFrontend
+	NVIDIAGPU      = Name("nvidia")
 	NVIDIACaps     = Name("nvidia-caps")
 	NVIDIAUVM      = Name("nvidia-uvm")
 
@@ -61,12 +65,15 @@ var _ Devices = devices(nil)
 
 // Exists checks if a Device with a given name exists or not
 func (d devices) Exists(name Name) bool {
-	_, exists := d[name]
+	_, exists := d.Get(name)
 	return exists
 }
 
 // Get a Device from Devices
 func (d devices) Get(name Name) (Major, bool) {
+	if name == NVIDIAGPU || name == NVIDIAFrontend {
+		return NVIDIADeviceMajor, true
+	}
 	device, exists := d[name]
 	return device, exists
 }
```
The change set seems a little easier to reason about. This also doesn't change the interface to the lower level package at all.

The general cleanups that we can make can be made as additional commits on top. 

Is there a way that we can verify the failure of the existing code and the fix in a unit test?

klueska:
> Thanks for this @tariq1890.
> 
> If we were to do the check for `nvidia` or `nvidia-frontend` in the `devices.Get` method as below:
> 
> ```
> diff --git a/internal/info/proc/devices/devices.go b/internal/info/proc/devices/devices.go
> index 95430428..65f3108f 100644
> --- a/internal/info/proc/devices/devices.go
> +++ b/internal/info/proc/devices/devices.go
> @@ -32,8 +32,12 @@ const (
>  	NVIDIACTLMinor      = 255
>  	NVIDIAModesetMinor  = 254
>  
> +	// NVIDIADeviceMajor is hardcoded as NV_MAJOR_DEVICE_NUMBER in nvidia-modprobe:
> +	// https://github.com/NVIDIA/nvidia-modprobe/blob/d6bce304f30b6661c9ab6a993f49340eafca7a7e/modprobe-utils/nvidia-modprobe-utils.c#L85
> +	NVIDIADeviceMajor = 195
> +
>  	NVIDIAFrontend = Name("nvidia-frontend")
> -	NVIDIAGPU      = NVIDIAFrontend
> +	NVIDIAGPU      = Name("nvidia")
>  	NVIDIACaps     = Name("nvidia-caps")
>  	NVIDIAUVM      = Name("nvidia-uvm")
>  
> @@ -61,12 +65,15 @@ var _ Devices = devices(nil)
>  
>  // Exists checks if a Device with a given name exists or not
>  func (d devices) Exists(name Name) bool {
> -	_, exists := d[name]
> +	_, exists := d.Get(name)
>  	return exists
>  }
>  
>  // Get a Device from Devices
>  func (d devices) Get(name Name) (Major, bool) {
> +	if name == NVIDIAGPU || name == NVIDIAFrontend {
> +		return NVIDIADeviceMajor, true
> +	}
>  	device, exists := d[name]
>  	return device, exists
>  }
> ```
> 
> The change set seems a little easier to reason about. This also doesn't change the interface to the lower level package at all.
> 
> The general cleanups that we can make can be made as additional commits on top.
> 
> Is there a way that we can verify the failure of the existing code and the fix in a unit test?

This seems cleaner and more consistent given that the rest of the devices are all looked up using their major device file.

elezar:
Another option would be to inject the hardcoded name -> major mappings when we read the `/proc/devices` file. Something like:
```
diff --git a/internal/info/proc/devices/devices.go b/internal/info/proc/devices/devices.go
index 95430428..7379f032 100644
--- a/internal/info/proc/devices/devices.go
+++ b/internal/info/proc/devices/devices.go
@@ -89,7 +89,14 @@ func nvidiaDevices(devicesPath string) (Devices, error) {
 	}
 	defer devicesFile.Close()
 
-	return nvidiaDeviceFrom(devicesFile)
+	procDevices, err := nvidiaDeviceFrom(devicesFile)
+	if err != nil {
+		return nil, err
+	}
+	procDevices["nvida"] = 195
+	procDevices["nvidia-frontend"] = 195
+
+	return procDevices, nil
 }
 
 var errNoNvidiaDevices = errors.New("no NVIDIA devices found")
```

This would prevent us from forgetting to use the `.Get` or `.Exists` methods and allow direct element access to work as expected.

tariq1890:
Thanks for the reviews @klueska @elezar 

Agreed that this PR is more complex than it needs to be. My only concern is that we are returning Major number of the nvidia device at all times when passing in `nvidia/nvidia-frontend` as input to the Get function. Shouldn't we address the scenario where there many not be an entry for nvidia gpu device in `/proc/devices` ?

tariq1890:
> Another option would be to inject the hardcoded name -> major mappings when we read the `/proc/devices` file. Something like:
> 
> ```
> diff --git a/internal/info/proc/devices/devices.go b/internal/info/proc/devices/devices.go
> index 95430428..7379f032 100644
> --- a/internal/info/proc/devices/devices.go
> +++ b/internal/info/proc/devices/devices.go
> @@ -89,7 +89,14 @@ func nvidiaDevices(devicesPath string) (Devices, error) {
>  	}
>  	defer devicesFile.Close()
>  
> -	return nvidiaDeviceFrom(devicesFile)
> +	procDevices, err := nvidiaDeviceFrom(devicesFile)
> +	if err != nil {
> +		return nil, err
> +	}
> +	procDevices["nvida"] = 195
> +	procDevices["nvidia-frontend"] = 195
> +
> +	return procDevices, nil
>  }

This would not be an accurate representation of the `/proc/devices`. Ideally we shouldn't have to mutate the current state

tariq1890:
> Is there a way that we can verify the failure of the existing code and the fix in a unit test?

@elezar I have created a PR for this in the meantime #334

elezar:
> Thanks for the reviews @klueska @elezar
> 
> Agreed that this PR is more complex than it needs to be. My only concern is that we are returning Major number of the nvidia device at all times when passing in `nvidia/nvidia-frontend` as input to the Get function. Shouldn't we address the scenario where there many not be an entry for nvidia gpu device in `/proc/devices` ?

The fact that we're not maintaining behaviour for when the `/proc/devices` file does not contain any `nvidia-*` entries is a valid concern. If we modify the `.Get` method as follows:
```
diff --git a/internal/info/proc/devices/devices.go b/internal/info/proc/devices/devices.go
index 95430428..976417ea 100644
--- a/internal/info/proc/devices/devices.go
+++ b/internal/info/proc/devices/devices.go
@@ -33,7 +33,7 @@ const (
 	NVIDIAModesetMinor  = 254
 
 	NVIDIAFrontend = Name("nvidia-frontend")
-	NVIDIAGPU      = NVIDIAFrontend
+	NVIDIAGPU      = Name("nvidia")
 	NVIDIACaps     = Name("nvidia-caps")
 	NVIDIAUVM      = Name("nvidia-uvm")
 
@@ -61,14 +61,28 @@ var _ Devices = devices(nil)
 
 // Exists checks if a Device with a given name exists or not
 func (d devices) Exists(name Name) bool {
-	_, exists := d[name]
+	_, exists := d.Get(name)
 	return exists
 }
 
 // Get a Device from Devices
 func (d devices) Get(name Name) (Major, bool) {
-	device, exists := d[name]
-	return device, exists
+	for _, n := range name.getWithFallback() {
+		device, exists := d[n]
+		if exists {
+			return device, true
+		}
+	}
+	return 0, false
+}
+
+// getWithFallback returns a prioritised list of device names for a specific name.
+// This allows multiple names to be associated with a single name to support various driver versions.
+func (n Name) getWithFallback() []Name {
+	if n == NVIDIAGPU || n == NVIDIAFrontend {
+		return []Name{NVIDIAGPU, NVIDIAFrontend}
+	}
+	return []Name{n}
 }
 
 // GetNVIDIADevices returns the set of NVIDIA Devices on the machine
```
where we add a `getWithFallback()` method to `Name` that allows multiple names to be specified and query each of these. That means for: `nvidia` or `nvidia-frontend` we will query `[]string{nvidia, nvidia-frontend}` but still fail if none of these exist in the file.

Thanks for creating #334. Feel free to integrate those changes as part of this PR so that it does not require additional changes once this is merged.

tariq1890:
Thanks for your patience with this @elezar . I've made the changes and it hopefully addresses all of your concerns.

internal/system/nvdevices/devices_test.go
```diff
@@ -46,6 +60,7 @@ func TestCreateControlDevices(t *testing.T) {
 		root          string
 		devices       devices.Devices
 		mknodeError   error
+		hasError      bool
```
elezar: Ideally `require.ErrorIs` should be used. If we're wrapping errors correctly, then things should work as expected.

internal/system/nvdevices/devices_test.go
```diff
@@ -126,9 +160,12 @@ func TestCreateControlDevices(t *testing.T) {
 			d.mknoder = mknode
 
 			err := d.CreateNVIDIAControlDevices()
-			require.ErrorIs(t, err, tc.expectedError)
+			if tc.hasError {
```
elezar: `require.ErrorIs` should hanldle the case where `tc.expectedError` is `nil`. The need to do this may have been affected by us not wrapping an error correctly somewhere in the callchain.

elezar (commented):
Thanks @tariq1890.

I have added a commit on top with some refactoring that should simplify testing this.

internal/system/nvdevices/devices_test.go
```diff
@@ -31,11 +31,25 @@ func TestCreateControlDevices(t *testing.T) {
 
 	nvidiaDevices := &devices.DevicesMock{
```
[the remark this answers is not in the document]
tariq1890: Agreed

elezar (approved):
LGTM

PR #334, opened 2024-02-02
add unit test to replicate error nvidia device rename

tariq1890:
This PR adds a unit test case to replicate the nvidia-ctk failing to create device nodes when on gpu drivers with versions >= `550.40.x`

### DIFF
--- a/internal/system/nvdevices/devices_test.go
+++ b/internal/system/nvdevices/devices_test.go
@@ -39,6 +39,18 @@ func TestCreateControlDevices(t *testing.T) {
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


elezar (commented):
I would recommend adding the tests as part of #330 instead. With the workflow (in theory) being:
1. Add a test that fails for a `/proc/devices` file with contents matching the 550 driver.
2. Make the changes to support the rename of `nvidia-frontend` to `nvidia`
3. Ensure that the test passed.

Note that this also assumes that there's already a test in place for a `/proc/devices` file that contains `nvidia-frontend` and that this continues to work.

elezar:
CLosing this as the tests were added as part of #330.

PR #337, opened 2024-02-05
Golang 1.17 compat

elezar:
Some customers need to build the components of the NVIDIA Container Toolkit from source using Golang 1.17.

This MR defines some minor change that allow the project to be built using Go 1.17, with an `internal/errors` package being used to provide basic `errors.Join` functionality. This is the only feature we currently use from Go 1.20.

Migrated from https://gitlab.com/nvidia/container-toolkit/container-toolkit/-/merge_requests/516

### DIFF
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -1,0 +1,27 @@
+/**
+# Copyright 2023 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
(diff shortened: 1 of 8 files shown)


internal/errors/errors.go
```diff
@@ -0,0 +1,27 @@
+/**
+# Copyright 2023 NVIDIA CORPORATION
```
ArangoGutierrez, line 2: 24

internal/errors/errors_legacy.go
```diff
@@ -0,0 +1,36 @@
+//go:build !go1.20
+
+/**
+# Copyright 2023 NVIDIA CORPORATION
```
ArangoGutierrez, line 4: 24

internal/errors/errors_modern.go
```diff
@@ -0,0 +1,28 @@
+//go:build go1.20
+
+/**
+# Copyright 2023 NVIDIA CORPORATION
```
ArangoGutierrez, line 4: 24

ArangoGutierrez (commented):
only a nit, the rest works for me

debarshiray (commented):
These days there's also the `slices` package.  See: https://github.com/NVIDIA/nvidia-container-toolkit/issues/886

ArangoGutierrez:
@elezar Let's rebase this one to see how it behaves with the E2E tests, and we can move on with reviewing and merging this initiative

PR #342, opened 2024-02-08
Add spec-dir flag to nvidia-ctk cdi list command

elezar:


cmd/nvidia-ctk/cdi/list/list.go
```diff
@@ -29,7 +30,9 @@ type command struct {
 	logger logger.Interface
 }
 
-type config struct{}
+type config struct {
+	cdiSpecDirs cli.StringSlice
```
tariq1890, line 34: I don't think we should inject `cli.StringSlice` into a non-CLI package. When we create this config struct, we should probably convert the `cli.StringSlice` to a `[]string`, thoughts?

elezar: I don't quite follow. The changes are CLI-only. We convert this to a regular slice by calling .Value()

tariq1890: oh never mind, I see that this file is int the `cmd/` package :). Will resolve

tariq1890 (approved):

PR #345, opened 2024-02-09
Add --create-device-nodes option to toolkit config

### LINKED ISSUE #344: Allow control device node creation to be skipped in the toolkit container
See the failure reported here: https://github.com/NVIDIA/nvidia-container-toolkit/issues/209#issuecomment-1929013503

When the driver root does not contain device nodes.

elezar:
This change adds a --create-device-nodes option to the toolkit config CLI. Most noteably, this allows the creation of control devices to be skipped when CDI spec generation is enabled.

Currently values of "", "node", and "control" are supported and can be set via the command line flag or the CREATE_DEVICE_NODES environment variable.

The default value of CREATE_DEVICE_NODES=control will trigger the creation of control device nodes. Setting this envvar to include the (comma-separated) strings of "" or "none" will disable device node creation regardless of whether other supported strings are included.

Closes #344


tools/container/toolkit/toolkit.go
```diff
@@ -251,6 +260,29 @@ func validateOptions(c *cli.Context, opts *options) error {
...
+		log.Warning("Skipping CDI spec generation (no output directory specified)")
+		opts.cdiEnabled = false
+	}
+
+	isDisabled := false
+	for _, mode := range opts.createDeviceNodes.Value() {
+		if mode != "" && mode != "none" && mode != "control" {
+			return fmt.Errorf("invalid --create-device-nodes value: %v", mode)
+		}
+		if mode == "" || mode == "node" {
```
cdesiniotis, line 273:
```suggestion
		if mode == "" || mode == "none" {
```

cdesiniotis (approved):

Issue #209, opened 2024-01-20
Container Toolkit Support GKE/COS Platform and its test coverage 


Dragoncell:
For pods inside the GPU Operator,  after driver installation finished, they rely on container toolkit starts on the node for setting up the nvidia container runtime:  
Download nvidia container runtime, hooks, container ctk(nvidia-ctk) and copy over from container to the host /run/nvidia/toolkit. [[link](https://github.com/NVIDIA/nvidia-container-toolkit/tree/main/cmd)]
Update the containerd config file based on container runtime like nvidia or nvidia-cdi 
Generate the CDI spec for management containers if runtime is nvidia-cdi  

Toolkit is a necessary component on GPU Operator, to make it work on COS, we needs to:

Support necessary binary from container toolkit for the COS platform. So far,  the container runtime, hooks, and container CTK are not yet supported ([supported platform lists](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/supported-platforms.html)). 

Starting from COS109, the nvidia-ctk is pre-built in COS. However, in current state (intermediate CDI mode), it still requires nvidia container runtime (nvidia-cdi) binaries. For legacy mode support, nvidia container runtime(nvidia-legacy) and its hooks are also required

The goal is to achieve the same functionality of container toolkit on COS platform with custom installed driver, container runtime binaries path

elezar:
@Dragoncell I am not too familiar with COS. Is there anything specific to the OS that means that the `toolkit-container` (with some modification) cannot be used to install the components of the NVIDIA Container Toolkit? What it essentially does is:
1. Copy the pre-built binaries and libraries from the running container image to a location mounted from the host.
2. Update the config file (on the host through the mount) for this location and the location of the containerized driver.

bobbypage:
/cc

One thing that comes to mind is that COS filesystem is read only (https://cloud.google.com/container-optimized-os/docs/concepts/disks-and-filesystem), so only certain paths are mounted with exec. So I think the path on the host should be configurable, i.e. we can't use default of `/usr/bin` since it's readonly.

cdesiniotis:
[...]

PR #360, opened 2024-02-14
Add dev-root option to toolkit container

elezar:
This adds support for using the toolkit container in environments where the "driverRoot" does not represent a chrootable filesystem.

For example, if the NVIDIA driver libraries are available at `/opt/nvidia/driver` on the host, but the device nodes are created at `/dev` on the host. Assuming these are mounted readOnly as follows into the toolkit container:
* `/opt/nvidia/driver` -> `/driver-root` 
* `/` -> `/host`

Setting the following envvars:
* `NVIDIA_DRIVER_ROOT=/opt/nvidia/driver`
* `DRIVER_ROOT_CTR_PATH=/driver-root`
* `NVIDIA_DEV_ROOT=/`
* `DEV_ROOT_CRT_PATH=/host`

will allow the CDI specification generation to complete and generate a valid CDI spec.

See https://github.com/NVIDIA/nvidia-container-toolkit/issues/209#issuecomment-1943441883

### DIFF
--- a/pkg/nvcdi/transform/root/driver-root.go
+++ b/pkg/nvcdi/transform/root/driver-root.go
@@ -1,0 +1,99 @@
+/**
+# Copyright 2024 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package root
+
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/pkg/nvcdi/transform"
+)
+
+type DriverOption func(*driverOptions)
+
+func WithDriverRoot(root string) DriverOption {
+	return func(do *driverOptions) {
+		do.driverRoot = root
+	}
+}
+
+func WithTargetDriverRoot(root string) DriverOption {
+	return func(do *driverOptions) {
+		do.targetDriverRoot = root
+	}
+}
+
+func WithDevRoot(root string) DriverOption {
+	return func(do *driverOptions) {
+		do.devRoot = root
+	}
+}
+
+func WithTargetDevRoot(root string) DriverOption {
+	return func(do *driverOptions) {
+		do.targetDevRoot = root
+	}
+}
+
+type driverOptions struct {
+	driverRoot       string
+	targetDriverRoot string
+	devRoot          string
+	targetDevRoot    string
+}
+
+// NewDriverTransformer creates a transformer for transforming driver specifications.
+func NewDriverTransformer(opts ...DriverOption) transform.Transformer {
+	d := &driverOptions{}
+	for _, opt := range opts {
+		opt(d)
+	}
+	if d.driverRoot == "" {
+		d.driverRoot = "/"
+	}
+	if d.targetDriverRoot == "" {
+		d.targetDriverRoot = "/"
+	}
+	if d.devRoot == "" {
+		d.devRoot = d.driverRoot
+	}
+	if d.targetDevRoot == "" {
+		d.targetDevRoot = d.targetDriverRoot
+	}
+
+	var transformers []transform.Transformer
+
+	if d.targetDevRoot != d.targetDriverRoot {
+		devRootTransformer := New(
+			WithRoot(ensureDev(d.devRoot)),
+			WithTargetRoot(ensureDev(d.targetDevRoot)),
+		)
+		transformers = append(transformers, devRootTransformer)
+	}
+
+	driverRootTransformer := New(
+		WithRoot(d.driverRoot),
+		WithTargetRoot(d.targetDriverRoot),
+	)
+	transformers = append(transformers, driverRootTransformer)
+
+	return transform.Merge(transformers...)
+}
+
+func ensureDev(p string) string {
+	return filepath.Join(strings.TrimSuffix(filepath.Clean(p), "/dev"), "/dev")
+}
--- a/tools/container/toolkit/toolkit.go
+++ b/tools/container/toolkit/toolkit.go
@@ -48,7 +48,9 @@ const (
 
 type options struct {
 	DriverRoot        string
+	DevRoot           string
 	DriverRootCtrPath string
+	DevRootCtrPath    string
 
 	ContainerRuntimeMode     string
 	ContainerRuntimeDebug    string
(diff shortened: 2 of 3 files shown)


klueska:
Just to be clear, if I want to make sure that `/dev` is referenced as my dev directory, I want to set `NVIDIA_DEV_ROOT=/`?

klueska:
I'm not opposed to this change, I'm just curious what the advantage of this would be over just mounting whatever `/dev` directory you wanted to use from the host into `/driver-root/dev` in the container (and then continue to assume that's where the `dev` folder you needed is? Is it because you need the *actual* host path in order to generate the CDI spec properly?

elezar:
> Just to be clear, if I want to make sure that `/dev` is referenced as my dev directory, I want to set `NVIDIA_DEV_ROOT=/`?

Yes, that is correct. The `devRoot` is the folder at which `/dev` is located. In the same way that the `driverRoot` implies that the libraries etc. exist at well-known locations rooted at that folder.

tools/container/toolkit/toolkit.go
```diff
@@ -702,6 +716,7 @@ func generateCDISpec(opts *options, nvidiaCTKPath string) error {
 	cdilib, err := nvcdi.New(
 		nvcdi.WithMode(nvcdi.ModeManagement),
 		nvcdi.WithDriverRoot(opts.DriverRootCtrPath),
+		nvcdi.WithDevRoot(opts.DevRootCtrPath),
```
cdesiniotis, line 719: Question -- do we also need to use `opts.DevRootCtrPath` when creating control device nodes in the code preceding this?

tools/container/toolkit/toolkit.go
```diff
@@ -714,6 +729,21 @@ func generateCDISpec(opts *options, nvidiaCTKPath string) error {
...
+		log.Infof("Transforming device root %v to %v", opts.DevRootCtrPath, opts.DevRoot)
+		ensureDev := func(p string) string {
+			return filepath.Join(strings.TrimSuffix(filepath.Clean(p), "/dev"), "/dev")
+		}
+		err = transformroot.New(
+			transformroot.WithRoot(ensureDev(opts.DevRootCtrPath)),
+			transformroot.WithTargetRoot(ensureDev(opts.DevRoot)),
+		).Transform(spec.Raw())
+		if err != nil {
+			return fmt.Errorf("failed to transform driver root in CDI spec: %v", err)
```
cdesiniotis, line 743:
```suggestion
			return fmt.Errorf("failed to transform dev root in CDI spec: %v", err)
```

elezar: Done.

tools/container/toolkit/toolkit.go
elezar: Yes, we did. I have updated it.

deployments/static/custom-driver.yaml
```diff
@@ -0,0 +1,116 @@
...
+      tolerations:
+        - key: nvidia.com/gpu
+          operator: Exists
+          effect: NoSchedule
+      priorityClassName: system-node-critical
+      serviceAccountName: nvidia-container-toolkit
+      hostPID: true
+      containers:
+      - image: nvidia/container-toolkit:1.15.0-ubuntu20.04
+        command: ["/work/nvidia-toolkit"]
```
Dragoncell, line 31: Wondering why use `/work/nvidia-toolkit` instead of `/bin/bash -c` with below args ? Does the yaml from the operator requires this change or not ?

deployments/static/custom-driver.yaml
```diff
@@ -0,0 +1,116 @@
...
+        # Set the main arguments.
+        - name: ROOT
+          value: "/usr/local/nvidia"
+        - name: RUNTIME
+          value: "containerd"
+        # Set the toolkit installation arguments.
+        - name: NVIDIA_DRIVER_ROOT
+          value: /host/nvidia/driver
+        - name: DRIVER_ROOT_CTR_PATH
+          value: /driver-root
```
Dragoncell, line 47: Have you tested set it to /driver-root works to generate the correct CDI spec for lib path ? from my experiment, it seems like doesn't work well. 

From my understanding,  from the container, it has two places holding the driver:
1. `/driver-root`, as it is mounted to driver-install-path
2. `/host/</host/nvidia/driver>`, as it is mounted to </host/nvidia/driver> path 

Below code is doing the transform: https://github.com/NVIDIA/nvidia-container-toolkit/blob/973a6633b30a95e27aff4ef07772387aac961395/tools/container/toolkit/toolkit.go#L773 

If using the 2), it can generate the desired CDI spec path mount like `hostPath: "</host/nvidia/driver>/lib64/XXX", containerPath: "/lib64/XXX"`

if using the 1), it generate like `hostPath: "</host/nvidia/driver>/lib64/XXX", containerPath: "</host/nvidia/driver>/lib64/XXX"`, which is probably not what as we expected

deployments/static/custom-driver.yaml
elezar: This spec was something that I put together to test the deployment of the toolkit container on Kind -- independent of th GPU operator. I was not deploying the entrypoint config map in this case.

The GPU Operator may not need this change since it uses the entrypoint to set the driver envvars below. 

cc @cdesiniotis

deployments/static/custom-driver.yaml
elezar: In my emulation of a GKE-like driver installation using kind it seems to work as expected. In the container when generating the spec I see:
```
time="2024-05-08T11:07:42Z" level=info msg="Selecting /driver-root/usr/lib/x86_64-linux-gnu/libcuda.so.515.105.01 as /usr/lib/x86_64-linux-gnu/libcuda.so.515.105.01"
``` 
(for example)

And in the generated management spec I see:
```
 grep libcuda.so management.nvidia.com-gpu.yaml
  - containerPath: /usr/lib/x86_64-linux-gnu/libcuda.so.515.105.01
    hostPath: /host/nvidia/driver/usr/lib/x86_64-linux-gnu/libcuda.so.515.105.01
```

Showing that the container path `/driver-root` is correctly transformed to the host path: `/host/nvidia/driver`.

I think the confusion may be that in the case I'm testing here, the node has the driver installed at `/host/nvidia/driver` -- which has nothing to do with `/host` in the toolkit contianer. I will update the example to use something that is less ambiguous.

Dragoncell: makes sense. Tested this change with GPU Operator changes by @cdesiniotis. It works as expected now.

deployments/static/custom-driver.yaml
Dragoncell: Gotcha. Thanks!

tools/container/toolkit/toolkit.go
```diff
@@ -778,6 +808,14 @@ func generateCDISpec(opts *options, nvidiaCTKPath string) error {
 		return fmt.Errorf("failed to transform driver root in CDI spec: %v", err)
 	}
 
+	// The transformation above may change the required spec version.
+	// TODO: This should be integrated into the transformer.
+	minVersion, err := cdi.MinimumRequiredVersion(spec.Raw())
```
elezar, line 813: I have created #506 to make this change unnecessary.

pkg/nvcdi/transform/root/driver-root.go
```diff
@@ -0,0 +1,99 @@
...
+	if d.devRoot == "" {
+		d.devRoot = d.driverRoot
+	}
+	if d.targetDevRoot == "" {
+		d.targetDevRoot = d.targetDriverRoot
+	}
+
+	var transformers []transform.Transformer
+
+	if d.targetDevRoot != "" && d.targetDevRoot != d.targetDriverRoot && d.devRoot != "" {
```
cdesiniotis, line 80: nit: can this be simplified to `if d.targetDevRoot != d.targetDriverRoot` since we apply default values for all these fields above, so none of them will be empty strings?

cdesiniotis (approved):

pkg/nvcdi/transform/root/driver-root.go
elezar: Yes you're right.

PR #379, opened 2024-02-27
[R550 driver support] add fallback logic to device.Exists(name)

tariq1890:
We missed to add the same fallback logic to the Exists method as was added to the Get method.

The gpu-operator-validator container uses the Exists method to assert the presence of `/dev/(nvidia|nvidia-frontend)`

### DIFF
--- a/internal/info/proc/devices/devices.go
+++ b/internal/info/proc/devices/devices.go
@@ -67,7 +67,7 @@ func (d devices) Count() int {
 
 // Exists checks if a Device with a given name exists or not
 func (d devices) Exists(name Name) bool {
-	_, exists := d[name]
+	_, exists := d.Get(name)
 	return exists
 }
 


internal/info/proc/devices/devices.go
```diff
@@ -67,8 +67,11 @@ func (d devices) Count() int {
 
 // Exists checks if a Device with a given name exists or not
 func (d devices) Exists(name Name) bool {
-	_, exists := d[name]
```
elezar, line 70: I'm sure I had this in a local change. Seems I missed it with the back and forth.
We could just call:
```
_, exists := d.Get(name)
```
here.

elezar (changes requested):

internal/info/proc/devices/devices.go
tariq1890: Sure

cdesiniotis (approved):

elezar:
I have overridden the merge. The image build was failing because the image was built from a branch and not a fork.

PR #382, opened 2024-02-28
Remove centos7 container-toolkit image

elezar:
Centos7 is almost EOL and from the point of view of the container-toolkit, we already include the centos7 packages in the ubi8 image.

This change removes the centos7 images from the release process.

Note that we still use centos7 to build our RPM packages to ensure a wider compatibility.

### DIFF
--- a/.common-ci.yml
+++ b/.common-ci.yml
@@ -224,13 +224,6 @@ test-packaging:
         OUT_IMAGE_VERSION: "${DEVEL_RELEASE_IMAGE_VERSION}"
 
 # Define the release jobs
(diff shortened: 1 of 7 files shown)


elezar:
@cdesiniotis I am sure that this will have implications for the GPU Operator documenation where any usage of the `-centos7` images must be replaced with `-ubi8`.

ArangoGutierrez:
Should we hold this PR until `June 30, 2024, CentOS Linux 7 will reach End of Life (EOL)`

tariq1890 (approved):

PR #384, opened 2024-02-28
Fix dockerfile arch selection

tariq1890:


versions.mk
```diff
@@ -28,7 +28,7 @@ NVIDIA_DOCKER_VERSION := 2.14.0
 NVIDIA_CONTAINER_RUNTIME_VERSION := 3.14.0
 
 CUDA_VERSION := 12.3.1
-GOLANG_VERSION := 1.20.5
+GOLANG_VERSION := 1.21.7
```
tariq1890, line 31: Is this update okay? Not sure what the glibc implications are with this

versions.mk
```diff
@@ -28,7 +28,7 @@ NVIDIA_DOCKER_VERSION := 2.14.0
 NVIDIA_CONTAINER_RUNTIME_VERSION := 3.14.0
 
 CUDA_VERSION := 12.3.1
-GOLANG_VERSION := 1.20.5
+GOLANG_VERSION := 1.21.5
```
tariq1890, line 31: Let me know if we are okay to update this. I am assuming we might run into the glibc version skew here as well

elezar: Yes, we would have to see whether these work in older distributions. I already have issues with `1.20.5` on some systems.

build/container/Dockerfile.centos
```diff
@@ -31,7 +31,7 @@ RUN set -eux; \
     case "${arch##*-}" in \
         x86_64 | amd64) ARCH='amd64' ;; \
         ppc64el | ppc64le) ARCH='ppc64le' ;; \
-        aarch64) ARCH='arm64' ;; \
+        aarch64 | arm64) ARCH='arm64' ;; \
```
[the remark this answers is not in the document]
tariq1890: `TARGETARCH` is better assuming its value is derived from the `docker buildx --platform` flag

versions.mk
tariq1890: I can revert it then. We definitely need to think of a solution for this though given 1.20 is close to EOL. Are we dependent on the glibc that comes from the `cuda-base-image`?

elezar: We don't build in the CUDA base images. We build in `ubuntu:18.04` and `centos:7` (`centos:8` for ppc64le).

Let's drop the golang bump and do that as a follow-up.

build/container/Dockerfile.centos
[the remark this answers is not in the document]
elezar: Yes. When building with `BUILDKIT` they're set automatically.

elezar (approved):
I have updated the PR to remove the Golang version bump for now.

PR #397, opened 2024-03-08
Use D3DKMTEnumAdapters3 for adapter enumeration

jbujak:
D3DKMTEnumAdapters3 is required to enumerate MCDM compute-only adapters

### DIFF
--- a/internal/dxcore/dxcore.c
+++ b/internal/dxcore/dxcore.c
@@ -41,14 +41,17 @@ static const char * const dxcore_nvidia_driver_store_components[] = {
  */
 
 struct dxcore_enumAdapters2;
+struct dxcore_enumAdapters3;
 struct dxcore_queryAdapterInfo;
 
 typedef int(*pfnDxcoreEnumAdapters2)(struct dxcore_enumAdapters2* pParams);
+typedef int(*pfnDxcoreEnumAdapters3)(struct dxcore_enumAdapters3* pParams);
 typedef int(*pfnDxcoreQueryAdapterInfo)(struct dxcore_queryAdapterInfo* pParams);
 
 struct dxcore_lib {
         void* hDxcoreLib;
         pfnDxcoreEnumAdapters2 pDxcoreEnumAdapters2;
+        pfnDxcoreEnumAdapters3 pDxcoreEnumAdapters3;
         pfnDxcoreQueryAdapterInfo pDxcoreQueryAdapterInfo;
 };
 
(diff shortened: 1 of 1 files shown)


internal/dxcore/dxcore.c
```diff
@@ -239,7 +251,36 @@ static void dxcore_add_adapter(struct dxcore_context* pCtx, struct dxcore_lib* p
         log_infof("Adding new adapter via dxcore hAdapter:%x luid:%llx wddm version:%d", pAdapterInfo->hAdapter, *((unsigned long long*)&pAdapterInfo->AdapterLuid), wddmVersion);
 }
 
-static void dxcore_enum_adapters(struct dxcore_context* pCtx, struct dxcore_lib* pLib)
+static int dxcore_enum_adapters3(struct dxcore_context* pCtx, struct dxcore_lib* pLib)
+{
+        struct dxcore_enumAdapters3 params = {0};
+        unsigned int adapterIndex = 0;
+
+        params.Filter = ENUMADAPTER3_FILTER_COMPUTE_ONLY;
```
elezar, line 259: As a matter of interest, how does this interact with adapters that would have been found using the previous implementation?

elezar (commented):
Thanks @jbujak. I think the mechanics look good. I have one question on the filter applied for the `EnumAdapters3` implementation.

Also, is favouring the `EnumAdapters3` what we want -- under which conditions should this function *not* be used?

internal/dxcore/dxcore.c
jbujak: > under which conditions should this function not be used?

`EnumAdapters3` was introduced in Windows 10, version 2004. To not break earlier versions, if this functions doesn't exist, we need to fallback to `EnumAdapters2`.

> how does this interact with adapters that would have been found using the previous implementation

This returns all the adapters returned previously and, additionally, compute-only adapters (it will skip display-only adapters, but those are not interesting for CUDA and would get filtered out later anyway).

elezar: Ok. It wasn't clear that this would still return the OTHER adapters. Could we at least add a comment or something to indicate that this is the behaviour?

jbujak: Added comments to explain

elezar: Thanks!

elezar (approved):
Thanks @jbujak!

Could you add the same comment to https://github.com/NVIDIA/libnvidia-container/pull/247

PR #427, opened 2024-03-25
Use functional options to construct driver root

elezar:


internal/lookup/root/options.go
```diff
@@ -0,0 +1,39 @@
+/**
+# Copyright 2023 NVIDIA CORPORATION
```
tariq1890, line 2:
```suggestion
# Copyright 2024 NVIDIA CORPORATION
```

tariq1890 (approved):

PR #437, opened 2024-04-02
Cherry pick NVIDIA_NVSWITCH and NVIDIA_GDRCOPY gated modifiers

cdesiniotis:
This PR cherry-picks commits from `main` which add support for injecting NVSWITCH and GDRCopy devices.


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # NVIDIA Container Toolkit Changelog
 
+* Add support for injecting /dev/nvidia-nvswitch* devices if the NVIDIA_NVSWITCH=enabled envvar is specified.
```
elezar, line 3: Is this required or was it simpler to backport if this was also added?

elezar (approved):
Minor question regarding the NVSWITCH feature gate.

internal/modifier/mofed.go
elezar, line 1: Seems like you actually need to remove the file.

elezar (changes requested):
The one file needs to be removed.

CHANGELOG.md
cdesiniotis: It is not strictly required. I can omit this.

Upon a closer look, I also need to cherry-pick [this commit](https://github.com/NVIDIA/nvidia-container-toolkit/commit/833254fa5937e2c18c775811522ded6fe8ffa975 ), as `gated.go` depends on some of it's refactoring. So this cherry-pick is a bit more intrusive than I originally thought.

elezar: I don't think that commit is required. The `image.Getenv` conditionals such as:
```
if image.Getenv(nvidiaGDSEnvvar) == "enabled" {
```
can be replaced by:
```
if gds, _ := image[nvidiaGDSEnvvar]; gds == "enabled" {
```

We would need to backport the driverRoot vs devRoot split -- which may be more problematic. Knowing the nature of the change, it may be simpler to *backport* the functionality and trigger it in a similar way to how we're currently handling the GDS and MOFED features on this branch.

internal/runtime/runtime_factory.go
```diff
@@ -87,21 +87,15 @@ func newSpecModifier(logger logger.Interface, cfg *config.Config, ociSpec oci.Sp
 		return nil, err
 	}
 
-	gdsModifier, err := modifier.NewGDSModifier(logger, cfg, image)
-	if err != nil {
-		return nil, err
-	}
-
-	mofedModifier, err := modifier.NewMOFEDModifier(logger, cfg, image)
+	featureModifier, err := modifier.NewFeatureGatedModifier(logger, cfg, image)
```
elezar, line 90: As mentioned above, a simpler change for the purpose of this backport may be to add `modifier.NewGDRCopyModifier` call here instead of backporting the feature-gated modifier.

cdesiniotis:
Closing for now as discussed offline.

PR #440, opened 2024-04-04
Find libnvidia-ml in driver root

elezar:


pkg/nvcdi/lib.go
```diff
@@ -113,7 +113,16 @@ func New(opts ...Option) (Interface, error) {
 		lib = (*managementlib)(l)
 	case ModeNvml:
 		if l.nvmllib == nil {
-			l.nvmllib = nvml.New()
+			var nvmlOpts []nvml.Option
+			candidates, err := l.driver.Libraries().Locate("libnvidia-ml.so.1")
+			if err != nil {
+				l.logger.Warningf("Ignoring error in locating libnvidia-ml.so.1: %v", err)
```
cdesiniotis, line 119: Why ignore the error? If we can't find `libnvidia-ml.so.1` here then wouldn't we be failing at a later point anyways when we try to initialize NVML?

cdesiniotis (commented):
What is the context / motivation for this change?

pkg/nvcdi/lib.go
elezar: The primary reason is that I didn't want to change existing behaviour unintentionally.

elezar:
> What is the context / motivation for this change?

Without this change CDI spec generation fails when running with a driver container.

PR #457, opened 2024-04-11
bump golang version to 1.22.2

tariq1890:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/NVIDIA/nvidia-container-toolkit
 
-go 1.20
+go 1.22.2
```
tariq1890, line 3: Should I revert this change @elezar ?

elezar: We already specify `GOLANG_VERSION ?= 1.22.2` in `versions.mk`. We also don't actually depend on any go1.22 features.

I think something like:
```
go 1.21
toolchain 1.22.2
```
might be more applicable here.

elezar (commented):
I don't think the change is strictly required.

go.mod
tariq1890: sure

tariq1890:
Sure I'll close the MR

PR #463, opened 2024-04-16
Update maintenance dependabot rule

elezar:
This change also changes the maintenance branch to `release-1.15` over `release-1.14` since `v1.15.0` has been released.

If we decide to continue supporting `v1.14.x` with updates we can reactivate this.


ArangoGutierrez (approved):
as discussed offline

PR #474, opened 2024-04-24
move nvidia-ctk hook command into own binary

### LINKED ISSUE #435: Consider separating `nvidia-ctk hook` from other functions?
Would the team consider separating `nvidia-ctk hook` from other functions?

The rationale is as follows.

The `hook` behaviours - currently `chmod`, `create-symlinks`, `update-ldcache` - run standard behaviours. They do not depend on anything GPU-specific or `libnvidia-ml.so` or CUDA in any way. The action of generating a CDI for a particular device (preparation), and the act of performing `hook` functions (runtime) are distinct, and can be executed at different times. I might even use a corporate or other non-JP OS which is not Ubuntu-based, generate my CDI inside a container, save the CDI yaml, and then a different process (which does not have access to the correct versions glibc or libnvidia-ml.so or other dependencies) would run the hooks. 

This would allow me to separate those two functionalities. It also would make the `hook` tool build much simpler.

One final advantage is understandability. When I run `nvidia-ctk cdi generate`, I have the option `--nvidia-ctkl-path <path>`, which really means "what path to `nvidia-ctk` should I place in the CDI yaml?" This can be confusing, since, aren't I executing `nvidia-ctk` right now? What does it mean to change my path to... myself?

If we separate it, e.g. `nvidia-cdi-hook`, then it becomes easier to understand: `nvidia-ctk cdi generate --nvidia-cdi-hook-path /path/to/nvidia-cdi-hook`. 

I would be fine opening a PR for it, but not going to waste the time if the maintainers don't want it.

deitch:
Fixes #435 

This does two things.

First: creates a new binary called `nvidia-cdi-hook`, as suggested by @klueska in #435. For backwards compatibility and ease of migration, I kept the functionality in `nvidia-ctk` as well as in `nvidia-cdi-hook`, so the following are equal:

```bash
$ nvidia-ctk hook create-symlinks
$ nvidia-cdi-hook create-symlinks
```

To avoid duplication, all of the command code that was under `cmd/nvidia-ctk/hook/{create-symlinks,update-ldcache,chmod}/` was moved to `cmd/nvidia-cdi-hook/`, and `cmd/nvidia-ctk/hook/hook.go` references those. This keeps the same functionality in place for `nvidia-ctk`, while avoiding duplication of code.

Second, I changed references in `nvidia-ctk cdi generate` from putting hooks that call `nvidia-ctk hook {subcommand}` with `nvidia-cdi-generate {subcommand}`. I am not convinced I got it correctly, and kept properties that reference `nvidiaCTKPath` in addition to `nvidiaCDIHookPath`, just in case.

There are 16 places left where `nvidiaCTKPath` remains in the code, almost all of them properties or setting options, which means they probably could be removed. 

As far as I could tell, and I am _not_ positive, references to `nvidia-ctk` inside the code are used _only_ for creating hooks. But I am not 100% sure, hence leaving them around. Looking forward to comments.


deitch:
I also am concerned that I missed some of the tests, or that there is a config property that needs to be added.

elezar:
@deitch thanks for the contribution. I'll have a look at it early next week.

cmd/nvidia-cdi-hook/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Functionality
+
+The `nvidia-cdi-hook` CLI provides the following functionality:
+
+* `chmod` - Change the permissions of a file or directory inside the directory path to be mounted into a container.
+* `create-symlinks` - Create symlinks inside the directory path to be mounted into a container.
+* `update-ldcache` - Update the dynamic linker cache inside the directory path to be mounted into a container.
+
+The CDI itself is created separately, normally using the `nvidia-ctk cdi` command.
```
lahwaacz, line 14: Can you document how to actually use the hook? Is it triggered automatically or users need to configure something?

deitch: I don't quite understand. It doesn't change from current behaviour. It's inserted into the cdi yaml when you run `nvidia-ctk cdi generate`

lahwaacz: So, can you document the current behavior?

For me it does not work by default though, `nvidia-ctk cdi generate` creates a CDI spec where is used only in `hooks:` sections, but it is missing in the `mounts:` section. Then e.g. when I run `podman run --rm -it --gpus all nvidia/cuda:12.4.1-devel-ubuntu22.04`, `/usr/bin/nvidia-ctk` is not available in the container.

deitch: > So, can you document the current behavior?

Sure, on its way.

deitch: > For me it does not work by default though, nvidia-ctk cdi generate creates a CDI spec where nvidia-ctk is used only in hooks: sections, but it is missing in the mounts: section. Then e.g. when I run podman run --rm -it --gpus all nvidia/cuda:12.4.1-devel-ubuntu22.04, /usr/bin/nvidia-ctk is not available in the container.

Do you mind posting your CDI here? I have 2 Jetsons, but unfortunately no access to them for the next week, so I have nothing ti generate CDIs on, in order to tests it. I also don't have remote access to anything with proper GPUs, so I am stuck working this through here without proper testing.

deitch: > So, can you document the current behavior?

Done.

lahwaacz: > It shouldn't be. containerd (or whatever CDI-compliant high-level runtime you use) is supposed to run the various hooks before launching the container, at least as far as I understand it.

Oh, that clarifies it. I've checked properly and e.g. the `create-symlinks` and `update-ldcache` hooks were indeed run, the container has the relevant symbolic links.

deitch: > Oh, that clarifies it. 

Cool, I am glad it works.

Do you mind sharing your generated CDI yaml? It will give me a better basis to check the work?

lahwaacz: Sure, here it is: https://gist.github.com/lahwaacz/0beed267b76ddb1c6d634e5ca2a3a229

deitch: Thank you. Heh, it is so small, compared to when I ran it on my Jetson Xavier NX. Last run had 3367 lines 😬 

I see it didn't include `nvidia-cdi-hook` for any of them, but rather `nvidia-ctk`. Is this generated by `nvidia-ctk` from the latest commit in this PR? If so, I definitely missed something.

deitch: Hmm, the code looks right. Do you mind pulling latest commit down, recompiling, and running `nvidia-ctk cdi generate` again?

deitch: Also, do you know of any way to emulate GPUs in a VM or container, so I can run this without accessing an actual device?

lahwaacz: Err, sorry - that was generated with the 1.15.0 version. This one is generated with the code from your branch: https://gist.github.com/lahwaacz/e28cf9aab89cc5175eee0c57436acc4d

Note there is a difference especially in the last `devices:` section (I don't know if it may be related to the 1.15.0 version existing in the system as well). There were some warnings when I run the command:

```
$ ./build/nvidia-ctk cdi generate
WARN[0000] Failed to locate nvidia-cdi-hook: pattern nvidia-cdi-hook not found
INFO[0000] Auto-detected mode as "nvml"
INFO[0000] Selecting /dev/nvidia0 as /dev/nvidia0
WARN[0000] Failed to evaluate symlink /dev/dri/by-path/pci--card; ignoring
WARN[0000] Failed to evaluate symlink /dev/dri/by-path/pci--render; ignoring
INFO[0000] Using driver version 550.67
INFO[0000] Selecting /dev/nvidia-modeset as /dev/nvidia-modeset
INFO[0000] Selecting /dev/nvidia-uvm-tools as /dev/nvidia-uvm-tools
INFO[0000] Selecting /dev/nvidia-uvm as /dev/nvidia-uvm
INFO[0000] Selecting /dev/nvidiactl as /dev/nvidiactl
Vim: Reading from stdin...
INFO[0000] Selecting /usr/lib/libnvidia-egl-gbm.so.1.1.1 as /usr/lib/libnvidia-egl-gbm.so.1.1.1
INFO[0000] Selecting /usr/share/glvnd/egl_vendor.d/10_nvidia.json as /usr/share/glvnd/egl_vendor.d/10_nvidia.json
INFO[0000] Selecting /usr/share/vulkan/icd.d/nvidia_icd.json as /usr/share/vulkan/icd.d/nvidia_icd.json
WARN[0000] Could not locate vulkan/icd.d/nvidia_layers.json: pattern vulkan/icd.d/nvidia_layers.json not found
INFO[0000] Selecting /usr/share/vulkan/implicit_layer.d/nvidia_layers.json as /usr/share/vulkan/implicit_layer.d/nvidia_layers.json
INFO[0000] Selecting /usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json as /usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json
INFO[0000] Selecting /usr/share/egl/egl_external_platform.d/10_nvidia_wayland.json as /usr/share/egl/egl_external_platform.d/10_nvidia_wayland.json
INFO[0000] Selecting /usr/share/nvidia/nvoptix.bin as /usr/share/nvidia/nvoptix.bin
WARN[0000] Could not locate nvidia/xorg/nvidia_drv.so: pattern nvidia/xorg/nvidia_drv.so not found
INFO[0000] Selecting /usr/lib/nvidia/xorg/libglxserver_nvidia.so.550.67 as /usr/lib/nvidia/xorg/libglxserver_nvidia.so.550.67
WARN[0000] Could not locate X11/xorg.conf.d/10-nvidia.conf: pattern X11/xorg.conf.d/10-nvidia.conf not found
INFO[0000] Selecting /usr/lib/libEGL_nvidia.so.550.67 as /usr/lib/libEGL_nvidia.so.550.67
INFO[0000] Selecting /usr/lib/libGLESv1_CM_nvidia.so.550.67 as /usr/lib/libGLESv1_CM_nvidia.so.550.67
INFO[0000] Selecting /usr/lib/libGLESv2_nvidia.so.550.67 as /usr/lib/libGLESv2_nvidia.so.550.67
INFO[0000] Selecting /usr/lib/libGLX_nvidia.so.550.67 as /usr/lib/libGLX_nvidia.so.550.67
INFO[0000] Selecting /usr/lib/libcuda.so.550.67 as /usr/lib/libcuda.so.550.67
INFO[0000] Selecting /usr/lib/libcudadebugger.so.550.67 as /usr/lib/libcudadebugger.so.550.67
INFO[0000] Selecting /usr/lib/libnvcuvid.so.550.67 as /usr/lib/libnvcuvid.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-allocator.so.550.67 as /usr/lib/libnvidia-allocator.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-cfg.so.550.67 as /usr/lib/libnvidia-cfg.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-eglcore.so.550.67 as /usr/lib/libnvidia-eglcore.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-encode.so.550.67 as /usr/lib/libnvidia-encode.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-fbc.so.550.67 as /usr/lib/libnvidia-fbc.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-glcore.so.550.67 as /usr/lib/libnvidia-glcore.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-glsi.so.550.67 as /usr/lib/libnvidia-glsi.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-glvkspirv.so.550.67 as /usr/lib/libnvidia-glvkspirv.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-gpucomp.so.550.67 as /usr/lib/libnvidia-gpucomp.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-ml.so.550.67 as /usr/lib/libnvidia-ml.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-ngx.so.550.67 as /usr/lib/libnvidia-ngx.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-nvvm.so.550.67 as /usr/lib/libnvidia-nvvm.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-opticalflow.so.550.67 as /usr/lib/libnvidia-opticalflow.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-pkcs11-openssl3.so.550.67 as /usr/lib/libnvidia-pkcs11-openssl3.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-pkcs11.so.550.67 as /usr/lib/libnvidia-pkcs11.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-ptxjitcompiler.so.550.67 as /usr/lib/libnvidia-ptxjitcompiler.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-rtcore.so.550.67 as /usr/lib/libnvidia-rtcore.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-tls.so.550.67 as /usr/lib/libnvidia-tls.so.550.67
INFO[0000] Selecting /usr/lib/libnvoptix.so.550.67 as /usr/lib/libnvoptix.so.550.67
INFO[0000] Selecting /run/nvidia-persistenced/socket as /run/nvidia-persistenced/socket
WARN[0000] Could not locate /nvidia-fabricmanager/socket: pattern /nvidia-fabricmanager/socket not found
WARN[0000] Could not locate /tmp/nvidia-mps: pattern /tmp/nvidia-mps not found
INFO[0000] Selecting /lib/firmware/nvidia/550.67/gsp_ga10x.bin as /lib/firmware/nvidia/550.67/gsp_ga10x.bin
INFO[0000] Selecting /lib/firmware/nvidia/550.67/gsp_tu10x.bin as /lib/firmware/nvidia/550.67/gsp_tu10x.bin
INFO[0000] Selecting /usr/bin/nvidia-smi as /usr/bin/nvidia-smi
INFO[0000] Selecting /usr/bin/nvidia-debugdump as /usr/bin/nvidia-debugdump
INFO[0000] Selecting /usr/bin/nvidia-persistenced as /usr/bin/nvidia-persistenced
INFO[0000] Selecting /usr/bin/nvidia-cuda-mps-control as /usr/bin/nvidia-cuda-mps-control
INFO[0000] Selecting /usr/bin/nvidia-cuda-mps-server as /usr/bin/nvidia-cuda-mps-server
INFO[0000] Generated CDI spec with version 0.5.0
```

deitch: Two questions. 

1. Do you mind rerunning with 1.15.0 and posting the logs, so I can compare?
2. Is the format on the second one [here](https://gist.github.com/lahwaacz/e28cf9aab89cc5175eee0c57436acc4d) messed up? The yaml indentation is off; compare with the first one you posted [here](https://gist.github.com/lahwaacz/0beed267b76ddb1c6d634e5ca2a3a229). Is that a copy-paste error? Or is the original yaml off?

klueska:
```
podman run --rm -it --gpus all nvidia/cuda:12.4.1-devel-ubuntu22.04
```
This does not use CDI. The command you want is:
```
podman run --rm -it --device nvidia.com/gpu=all nvidia/cuda:12.4.1-devel-ubuntu22.04
```

lahwaacz: @deitch The logs with 1.15.0 are:
```
INFO[0000] Auto-detected mode as "nvml"
INFO[0000] Selecting /dev/nvidia0 as /dev/nvidia0
INFO[0000] Selecting /dev/dri/card1 as /dev/dri/card1
INFO[0000] Selecting /dev/dri/renderD128 as /dev/dri/renderD128
INFO[0000] Using driver version 550.67
INFO[0000] Selecting /dev/nvidia-modeset as /dev/nvidia-modeset
INFO[0000] Selecting /dev/nvidia-uvm-tools as /dev/nvidia-uvm-tools
INFO[0000] Selecting /dev/nvidia-uvm as /dev/nvidia-uvm
INFO[0000] Selecting /dev/nvidiactl as /dev/nvidiactl
INFO[0000] Selecting /usr/lib/libnvidia-egl-gbm.so.1.1.1 as /usr/lib/libnvidia-egl-gbm.so.1.1.1
INFO[0000] Selecting /usr/share/glvnd/egl_vendor.d/10_nvidia.json as /usr/share/glvnd/egl_vendor.d/10_nvidia.json
INFO[0000] Selecting /usr/share/vulkan/icd.d/nvidia_icd.json as /usr/share/vulkan/icd.d/nvidia_icd.json
WARN[0000] Could not locate vulkan/icd.d/nvidia_layers.json: pattern vulkan/icd.d/nvidia_layers.json not found
INFO[0000] Selecting /usr/share/vulkan/implicit_layer.d/nvidia_layers.json as /usr/share/vulkan/implicit_layer.d/nvidia_layers.json
INFO[0000] Selecting /usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json as /usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json
INFO[0000] Selecting /usr/share/egl/egl_external_platform.d/10_nvidia_wayland.json as /usr/share/egl/egl_external_platform.d/10_nvidia_wayland.json
INFO[0000] Selecting /usr/share/nvidia/nvoptix.bin as /usr/share/nvidia/nvoptix.bin
WARN[0000] Could not locate nvidia/xorg/nvidia_drv.so: pattern nvidia/xorg/nvidia_drv.so not found
INFO[0000] Selecting /usr/lib/nvidia/xorg/libglxserver_nvidia.so.550.67 as /usr/lib/nvidia/xorg/libglxserver_nvidia.so.550.67
WARN[0000] Could not locate X11/xorg.conf.d/10-nvidia.conf: pattern X11/xorg.conf.d/10-nvidia.conf not found
INFO[0000] Selecting /usr/lib/libEGL_nvidia.so.550.67 as /usr/lib/libEGL_nvidia.so.550.67
INFO[0000] Selecting /usr/lib/libGLESv1_CM_nvidia.so.550.67 as /usr/lib/libGLESv1_CM_nvidia.so.550.67
INFO[0000] Selecting /usr/lib/libGLESv2_nvidia.so.550.67 as /usr/lib/libGLESv2_nvidia.so.550.67
INFO[0000] Selecting /usr/lib/libGLX_nvidia.so.550.67 as /usr/lib/libGLX_nvidia.so.550.67
INFO[0000] Selecting /usr/lib/libcuda.so.550.67 as /usr/lib/libcuda.so.550.67
INFO[0000] Selecting /usr/lib/libcudadebugger.so.550.67 as /usr/lib/libcudadebugger.so.550.67
INFO[0000] Selecting /usr/lib/libnvcuvid.so.550.67 as /usr/lib/libnvcuvid.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-allocator.so.550.67 as /usr/lib/libnvidia-allocator.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-cfg.so.550.67 as /usr/lib/libnvidia-cfg.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-eglcore.so.550.67 as /usr/lib/libnvidia-eglcore.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-encode.so.550.67 as /usr/lib/libnvidia-encode.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-fbc.so.550.67 as /usr/lib/libnvidia-fbc.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-glcore.so.550.67 as /usr/lib/libnvidia-glcore.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-glsi.so.550.67 as /usr/lib/libnvidia-glsi.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-glvkspirv.so.550.67 as /usr/lib/libnvidia-glvkspirv.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-gpucomp.so.550.67 as /usr/lib/libnvidia-gpucomp.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-ml.so.550.67 as /usr/lib/libnvidia-ml.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-ngx.so.550.67 as /usr/lib/libnvidia-ngx.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-nvvm.so.550.67 as /usr/lib/libnvidia-nvvm.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-opticalflow.so.550.67 as /usr/lib/libnvidia-opticalflow.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-pkcs11-openssl3.so.550.67 as /usr/lib/libnvidia-pkcs11-openssl3.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-pkcs11.so.550.67 as /usr/lib/libnvidia-pkcs11.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-ptxjitcompiler.so.550.67 as /usr/lib/libnvidia-ptxjitcompiler.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-rtcore.so.550.67 as /usr/lib/libnvidia-rtcore.so.550.67
INFO[0000] Selecting /usr/lib/libnvidia-tls.so.550.67 as /usr/lib/libnvidia-tls.so.550.67
INFO[0000] Selecting /usr/lib/libnvoptix.so.550.67 as /usr/lib/libnvoptix.so.550.67
INFO[0000] Selecting /run/nvidia-persistenced/socket as /run/nvidia-persistenced/socket
WARN[0000] Could not locate /nvidia-fabricmanager/socket: pattern /nvidia-fabricmanager/socket not found
WARN[0000] Could not locate /tmp/nvidia-mps: pattern /tmp/nvidia-mps not found
INFO[0000] Selecting /lib/firmware/nvidia/550.67/gsp_ga10x.bin as /lib/firmware/nvidia/550.67/gsp_ga10x.bin
INFO[0000] Selecting /lib/firmware/nvidia/550.67/gsp_tu10x.bin as /lib/firmware/nvidia/550.67/gsp_tu10x.bin
INFO[0000] Selecting /usr/bin/nvidia-smi as /usr/bin/nvidia-smi
INFO[0000] Selecting /usr/bin/nvidia-debugdump as /usr/bin/nvidia-debugdump
INFO[0000] Selecting /usr/bin/nvidia-persistenced as /usr/bin/nvidia-persistenced
INFO[0000] Selecting /usr/bin/nvidia-cuda-mps-control as /usr/bin/nvidia-cuda-mps-control
INFO[0000] Selecting /usr/bin/nvidia-cuda-mps-server as /usr/bin/nvidia-cuda-mps-server
INFO[0000] Generated CDI spec with version 0.5.0
```
I don't see an indentation error in any of the two yaml files...

@klueska Podman always uses CDI, the `--gpus` flag works as an alias for `--device nvidia.com/gpu=`. See https://github.com/containers/podman/pull/21180

klueska: Interesting -- that's new (and unfortunate). The `--gpus` flag is inconsistent and never should have been introduced in the first place. We'd prefer it to die a slow death than be added anew to other projects.

deitch: > I don't see an indentation error in any of the two yaml files...

You are correct. That was strange, it was completely off the first time I loaded it. Must have been a GitHub presentation error. I just took the raw version this time.

deitch: In terms of the logs, there are only minor differences:

```
$ diff logs-old.txt logs-new.txt
0a1,2
> $ ./build/nvidia-ctk cdi generate
> WARN[0000] Failed to locate nvidia-cdi-hook: pattern nvidia-cdi-hook not found
3,4c5,6
< INFO[0000] Selecting /dev/dri/card1 as /dev/dri/card1
< INFO[0000] Selecting /dev/dri/renderD128 as /dev/dri/renderD128
---
> WARN[0000] Failed to evaluate symlink /dev/dri/by-path/pci--card; ignoring
> WARN[0000] Failed to evaluate symlink /dev/dri/by-path/pci--render; ignoring
```

The first error is understood. It is looking for `nvidia-cdi-hook` but didn't find it. If you have a moment, and don't mind compiling it and putting it in a standard path, and seeing if that matters, that would help.

For the second set - the issue with `card` and `render` - I will have to dig deeper to try and understand.

deitch: The differences in the yaml files at least partially are due to the card and render devices. Let's figure that out, and then see if the issue lingers.

deitch: The successful log line is [here](https://github.com/NVIDIA/nvidia-container-toolkit/blob/fde099d25b98fc10886b49a7a3730f49c284b13e/internal/discover/mounts.go#L97), while the failure line is [here](https://github.com/NVIDIA/nvidia-container-toolkit/blob/fde099d25b98fc10886b49a7a3730f49c284b13e/pkg/nvcdi/full-gpu-nvml.go#L199). Not clear the flow from one to the other.

I wouldn't mind walking through this in a debugger. I don't assume there is any way to walk it through together @lahwaacz ?

lahwaacz: OK, the `WARN[0000] Failed to locate nvidia-cdi-hook: pattern nvidia-cdi-hook not found` is gone when I add the build directory to `$PATH`.

For the `Failed to evaluate symlink` warnings, it seems there is some variable expansion missing: `/dev/dri/by-path/pci--card` should be `/dev/dri/by-path/pci-0000:21:00.0-card` and similar for `render`.

deitch: > OK, the WARN[0000] Failed to locate nvidia-cdi-hook: pattern nvidia-cdi-hook not found is gone when I add the build directory to $PATH.

One down, one to go (another town and one more show 🎶 ). Thanks, that helps.

> For the Failed to evaluate symlink warnings, it seems there is some variable expansion missing: /dev/dri/by-path/pci--card should be /dev/dri/by-path/pci-0000:21:00.0-card and similar for render

Ah, that makes sense. I will see if I can figure it out any further from here, and if not, we will look for alternate methods.

deitch: I don't think that part of the code is changed, but I will keep trying.

> I don't assume there is any way to walk it through together @lahwaacz ?

If this is a possibility, let me know?

deitch: @lahwaacz I added a commit with a whole bunch of debugging statements. Do you mind recompiling `nvidia-ctk`, and running generate again but with debug? `nvidia-ctk --debug cdi generate`? And posting logs?

lahwaacz: Here are the logs up to the two warnings:
```
DEBU[0000] Locating NVIDIA Container Toolkit CLI as nvidia-cdi-hook
DEBU[0000] Locating "nvidia-cdi-hook" in [/usr/local/sbin /usr/local/bin /usr/bin /opt/cuda/bin /opt/cuda/nsight_compute /opt/cuda/nsight_systems/bin /usr/bin/site_perl /usr/bin/vendor_perl /usr/bin/core_perl /home/lahwaacz/stuff/nvidia-container-toolkit/build /usr/sbin /sbin /bin]
DEBU[0000] Checking candidate '/home/lahwaacz/stuff/nvidia-container-toolkit/build/nvidia-cdi-hook'
DEBU[0000] Found 1 candidates; ignoring further candidates
DEBU[0000] Found nvidia-cdi-hook candidates: [/home/lahwaacz/stuff/nvidia-container-toolkit/build/nvidia-cdi-hook]
DEBU[0000] Using NVIDIA Container Toolkit CLI path nvidia-cdi-hook
DEBU[0000] Is WSL-based system? false: could not load DXCore library: libdxcore.so: cannot open shared object file: No such file or directory
DEBU[0000] Is NVML-based system? true: found NVML library
DEBU[0000] Is Tegra-based system? false: /sys/devices/soc0/family file not found
INFO[0000] Auto-detected mode as "nvml"
DEBU[0000] newFullGPUDiscoverer() devRoot / device &device.device{Device:nvml.nvmlDevice{Handle:(*nvml._Ctype_struct_nvmlDevice_st)(0x7a0371bfaf38)}, lib:(*device.devicelib)(0xc0000173c0), migProfiles:[]device.MigProfile(nil)}
DEBU[0000] newFullGPUDiscoverer() minor 0
DEBU[0000] newFullGPUDiscoverer() pciInfo nvml.PciInfo{BusIdLegacy:[16]int8{48, 48, 48, 48, 58, 50, 49, 58, 48, 48, 46, 48, 0, 0, 0, 0}, Domain:0x0, Bus:0x21, Device:0x0, PciDeviceId:0x1eb010de, PciSubSystemId:0x0, BusId:[32]int8{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}}
DEBU[0000] newFullGPUDiscoverer() pciBusID
DEBU[0000] Locating /dev/nvidia0
DEBU[0000] Locating "/dev/nvidia0" in [/ /dev]
DEBU[0000] Checking candidate '/dev/nvidia0'
DEBU[0000] Located /dev/nvidia0 as [/dev/nvidia0]
INFO[0000] Selecting /dev/nvidia0 as /dev/nvidia0
DEBU[0000] byPathHookDiscoverer.deviceNodeLinks() for nvcdi.byPathHookDiscoverer{logger:(*logrus.Logger)(0xc0000d4a00), devRoot:"/", nvidiaCDIHookPath:"/home/klinkovsky/stuff/nvidia-container-toolkit/build/nvidia-cdi-hook", pciBusID:"", deviceNodes:(*discover.charDevices)(0xc0001d6850)}
DEBU[0000] returning cached mounts
DEBU[0000] byPathHookDiscoverer.deviceNodeLinks() devices []discover.Device{discover.Device{HostPath:"/dev/nvidia0", Path:"/dev/nvidia0"}}
DEBU[0000] byPathHookDiscoverer.deviceNodeLinks() candidates [/dev/dri/by-path/pci--card /dev/dri/by-path/pci--render]
WARN[0000] Failed to evaluate symlink /dev/dri/by-path/pci--card; ignoring
WARN[0000] Failed to evaluate symlink /dev/dri/by-path/pci--render; ignoring
```

deitch: > DEBU[0000] newFullGPUDiscoverer() pciInfo nvml.PciInfo{BusIdLegacy:[16]int8{48, 48, 48, 48, 58, 50, 49, 58, 48, 48, 46, 48, 0, 0, 0, 0}, Domain:0x0, Bus:0x21, Device:0x0, PciDeviceId:0x1eb010de, PciSubSystemId:0x0, BusId:[32]int8{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}}
> DEBU[0000] newFullGPUDiscoverer() pciBusID

pciBudID is all all `0x00`. This part comes from go-nvlib, which wasn't changed. Does this have to do with how we compile it in? `make build` just does a `go build ./...`, so I am not sure. @klueska do you understand how this part works?

deitch: Oh interesting. So I didn't introduce the bug. 

If you've compiled it from main, do you mind generating the CDI yaml from main, so we can compare? It doesn't get us all of the way, but it helps. 

Hopefully @klueska or one of the other maintainers will know what we need to do to complete the devices.

lahwaacz: @deitch CDI from main: https://gist.github.com/lahwaacz/ca4868a70ca106cb4a90580b6735595e

deitch: Thanks @lahwaacz . Those are identical, except for the executable. That is a good sign.

It means that whatever introduced the issue with the `busId`, is in main, and was not introduced by this PR.

deitch: I did a quick check of `main` vs `v1.15.0`, there is quite a bit in there that I do not understand. I think we are stuck until someone who gets how this works explains how to fix it in main (and we can apply here).

deitch:
Following up on this. Can someone help us understand what is different in `main` vs 1.15.0, such that the config it missing devices? How can we fix this?

elezar:
> Following up on this. Can someone help us understand what is different in main vs 1.15.0, such that the config it missing devices? How can we fix this?

I was able to reproduce the issue regarding the bus information for the device being empty for `main`. It seems as if this was related to our changes to the `go-nvml` wrappers. After updating the `go-nvml` dependency to `v0.12.0-5` this issue no longer presents.

deitch:
Excellent, thank you @elezar . It looks like that already is in, so I should be able to rebase on main and it will work?

deitch:
@lahwaacz I rebased on main, with the fix that @elezar put in. Do you mind running your generation again? Off of v1.15.0, main and this branch, and then posting the 3 in a linked gist, so we can compare?

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -132,9 +132,9 @@ func (m command) build() *cli.Command {
 			Destination: &opts.librarySearchPaths,
 		},
 		&cli.StringFlag{
-			Name:        "nvidia-ctk-path",
-			Usage:       "Specify the path to use for the nvidia-ctk in the generated CDI specification. If this is left empty, the path will be searched.",
-			Destination: &opts.nvidiaCTKPath,
+			Name:        "nvidia-cdi-hook-path",
```
elezar, line 135: Note that this is a breaking change. Does it make sense to add an `Alias` here instead and then add logic somewhere(TM) for determining whether the `nvidia-cdi-hook` or `nvidia-ctk` is used to know how we will generate the CDI spec?

internal/discover/graphics.go
```diff
@@ -36,20 +36,20 @@ import (
 // TODO: The logic for creating DRM devices should be consolidated between this
 // and the logic for generating CDI specs for a single device. This is only used
 // when applying OCI spec modifications to an incoming spec in "legacy" mode.
-func NewDRMNodesDiscoverer(logger logger.Interface, devices image.VisibleDevices, devRoot string, nvidiaCTKPath string) (Discover, error) {
+func NewDRMNodesDiscoverer(logger logger.Interface, devices image.VisibleDevices, devRoot string, nvidiaCDIHookPath string) (Discover, error) {
```
elezar, line 39: Using this as an example, I think we want to introduce an abstraction around a CDI hook that we inject here instead. That can contain the logic for generating the command / hook for a set of arguments.

We already have this logic in `CreateNvidiaCDIHook` (`internal/discover/hooks.go`) but it could be better organized.

Something like:
```
type HookCreator interface {
    CreateHook(name string, args ...string) discover.Hook
}
```

Where we could implement `CreateHook` as something like:
```
func (c *creator) CreateHook(name string, args ...string) Hook {
	return Hook{
		Lifecycle: cdi.CreateContainerHook,
		Path: c.Path,
		Args: append(c.requiredArgs(name), args...),
	}
}

func (c *creator) requiredArgs(name string) []string {
	base := filepath.Base(c.Path)
	if base == "nvidia-ctk" {
		return []string{base, "hook", name}
	}
	return []string{base, name}
}
```

elezar: This would allow us to make a decision on how to construct this when we are initialising things for the `nvidia-container-runtime` or for CDI spec generation instead of passing paths down and making this decision later.

Note that this can be a seprate PR which captures the existing behaviour and then we can add the new type of hook creator as a follow-up.

elezar (changes requested):
@deitch I think I'm ok with the mechanics of this. The one question I have is how to support the transition from the use of `nvidia-ctk hook` to `nvidia-cdi-hook`.

Note that we would also need to include the new executable in the deb and rpm package definitions.

deitch:
> The one question I have is how to support the transition from the use of nvidia-ctk hook to nvidia-cdi-hook.

Is there what to do for this? Since they are packaged together, if you install and run a newer version of a package with ctk, it installs `nvidia-cdi-hook`, and `nvidia-ctk cdi generate` will generate CDIs with `nvidia-cdi-hook` rather than `nvidia-ctk hook`. If you update your package, but have a CDI generated by an older version of `nvidia-ctk cdi generate`, it still will work, because the newer version of `nvidia-ctk` continues to support `nvidia-ctk hook`.

> Note that we would also need to include the new executable in the deb and rpm package definitions

Is that in [here](https://github.com/NVIDIA/nvidia-container-toolkit/tree/main/packaging)?

cmd/nvidia-ctk/cdi/generate/generate.go
deitch: > breaking change

Do you mean that someone who might have some command process (human or scripted) that called this with `--nvidia-ctk-path` will now break, because that option doesn't exist?

I would _think_ (cautiously, and probably wrongly) that it isn't that heavily used an option, and if you install a new version of the toolkit, it is ok for this to break.

If we do not want to allow that, then, sure, I can add an `Alias`. Your call.

elezar: Yes, I know that this may not be the most used option, but I would prefer to either add an alias or add a new option -- possibly marking the old one as deprecated. Given the discussion around a hook constructor below, maybe a separate option is better -- caling out in the usage text that this is intended to replace the `nvidia-ctk` command.

internal/discover/graphics.go
deitch: > Note that this can be a seprate PR which captures the existing behaviour and then we can add the new type of hook creator as a follow-up

That sounds like a good idea. I don't know that we want to mix these two distinct PRs: one that splits the functionality out into a library surface and creates a new CLI, with one that restructures some of how the things are called inside that "library".

And I am not 100% sure I understand it anyways, so it probably is better to have a dedicated discussion to it.

cmd/nvidia-ctk/cdi/generate/generate.go
deitch: Sure. Will add it.

deitch: What kind of behaviour do we want for that? The deprecation is easy to add to the comment:

```go
		&cli.StringFlag{
			Name:        "nvidia-ctk-path",
			Usage:       "Specify the path to use for the nvidia-ctk in the generated CDI specification. If this is left empty, the path will be searched. DEPRECATED. Use flag --nvidia-cdi-hook-path instead.",
		},
```

but when someone uses it, should we:

* treat it like they called `--nvidia-cdi-hook-path` and set that? I don't much like it, as they are distinct binaries.
* do nothing with it? That doesn't seem to add value.
* save it to somewhere and use it? But we don't use that anywhere.

🤷‍♂️

elezar:
> Is that in [here](https://github.com/NVIDIA/nvidia-container-toolkit/tree/main/packaging)?

Yes. That's the location where the executable should be added (to the `nvidia-container-toolkit-base` packages).

deitch:
> Yes. That's the location where the executable should be added (to the nvidia-container-toolkit-base packages).

OK, I think I got those @elezar ; please take a look.

The only remaining thing is [this](https://github.com/NVIDIA/nvidia-container-toolkit/pull/474#discussion_r1589143702).

We keep `--nvidia-ctk-path <path>` around as an option, but what do we do with it?

cmd/nvidia-ctk/cdi/generate/generate.go
elezar: From the user's perspective, I think the two are equivalent and we can handle differences when generating the hook.

Assuming we make the following changes to your PR:
```
diff --git a/internal/config/config.go b/internal/config/config.go
index 008517c5..1e5a8fdf 100644
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -192,6 +192,9 @@ func ResolveNVIDIACTKPath(logger logger.Interface, nvidiaCTKPath string) string
 // If the path is specified as an absolute path, it is used directly
 // without checking for existence of an executable at that path.
 func ResolveNVIDIACDIHookPath(logger logger.Interface, nvidiaCDIHookPath string) string {
+	if filepath.Base(nvidiaCDIHookPath) == "nvidia-ctk" {
+		return ResolveNVIDIACTKPath(logger, nvidiaCDIHookPath)
+	}
 	return resolveWithDefault(
 		logger,
 		"NVIDIA Container Toolkit CLI",
diff --git a/internal/discover/hooks.go b/internal/discover/hooks.go
index 39c93135..0db731ea 100644
--- a/internal/discover/hooks.go
+++ b/internal/discover/hooks.go
@@ -59,9 +59,23 @@ func CreateCreateSymlinkHook(nvidiaCDIHookPath string, links []string) Discover
 
 // CreateNvidiaCDIHook creates a hook which invokes the NVIDIA Container CLI hook subcommand.
 func CreateNvidiaCDIHook(nvidiaCDIHookPath string, hookName string, additionalArgs ...string) Hook {
+	return cdiHook(nvidiaCDIHookPath).Create(hookName, additionalArgs...)
+}
+
+type cdiHook string
+
+func (c cdiHook) Create(name string, args ...string) Hook {
 	return Hook{
 		Lifecycle: cdi.CreateContainerHook,
-		Path:      nvidiaCDIHookPath,
-		Args:      append([]string{filepath.Base(nvidiaCDIHookPath), hookName}, additionalArgs...),
+		Path:      string(c),
+		Args:      append(c.requiredArgs(name), args...),
+	}
+}
+
+func (c cdiHook) requiredArgs(name string) []string {
+	base := filepath.Base(string(c))
+	if base == "nvidia-ctk" {
+		return []string{base, "hook", name}
 	}
+	return []string{base, name}
 }
```

We would include logic to handle both the `nvidia-cdi-hook` and `nvidia-ctk` case from the perspective of CDI spec generation. This means that we can add an aliases `--nvidia-cdi-hook-path` and `--nvidia-ctk-path` and have them treated consistently. Note that this ASSUMES that users have not renamed their `nvidia-ctk` executable for the purpose of injecting the hook.

We can address the `nvidia-container-runtime` where we may still use `nvidia-ctk` when generating dynamic specs in a follow-up.

What do you think of this proposal?

internal/discover/graphics.go
elezar: I had a quick look at this, and although I think it would add value to better encapsulate hook creation, there are some other things that would need to be cleaned up first.

Let's ignore this from the point of view of this PR.

Note that the logic to optionally inject `"hook"` when generating the CDI spec may still be useful -- although it doesn't need the more agressive refactoring proposed here.

cmd/nvidia-ctk/cdi/generate/generate.go
deitch: > Note that this ASSUMES that users have not renamed their nvidia-ctk executable for the purpose of injecting the hook

If a user does that, they only have themselves to blame. You always can break something if you try hard enough. If they replace `/usr/bin/containerd` with a symlink to `/usr/sbin/halt`, I don't think we can help them 😆

deitch: > We would include logic to handle both the nvidia-cdi-hook and nvidia-ctk case from the perspective of CDI spec generation. This means that we can add an aliases --nvidia-cdi-hook-path and --nvidia-ctk-path and have them treated consistently

So

* if the user uses `--nvidia-ctk-path`, then it would inject `nvidia-ctk hook` (2 words) everywhere in CDI
* if the user uses `--nvidia-cdi-hook-path`, then it would inject `nvidia-cdi-hook` (1 word) everywhere in CDI
* if the user uses both, it is an error
* if the user uses neither, it uses `nvidia-cdi-hook` as the default

If I understood you correctly, sure, that works.

deitch: Anything else we need for this?

cmd/nvidia-cdi-hook/README.md
```diff
@@ -0,0 +1,31 @@
+# NVIDIA CDI Hook
+
+The CLI `nvidia-cdi-hook` provides container device runtime hook capabilities when
+called by a container runtime, as specific in a
+[Container Device Interface](https://github.com/cncf-tags/container-device-interface/blob/main/SPEC.md)
+file.
+
+## Generating a CDI
+
+The CDI itself is created for an Nvidia-capable device using the
```
elezar, line 10:
```suggestion
The CDI itself is created for an NVIDIA-capable device using the
```

cmd/nvidia-cdi-hook/README.md
```diff
@@ -0,0 +1,31 @@
...
+[`nvidia-ctk cdi`](../nvidia-ctk/) command.
```
elezar, line 11:
```suggestion
[`nvidia-ctk cdi generate`](../nvidia-ctk/) command.
```

cmd/nvidia-cdi-hook/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+When `nvidia-ctk cdi generate` is run, the CDI specification is generated as a yaml file.
+The CDI file provides instructions for a container runtime to set up devices, files and
```
elezar, line 14:
```suggestion
The CDI specification provides instructions for a container runtime to set up devices, files and
```

cmd/nvidia-cdi-hook/README.md
```diff
@@ -0,0 +1,31 @@
+# NVIDIA CDI Hook
+
+The CLI `nvidia-cdi-hook` provides container device runtime hook capabilities when
+called by a container runtime, as specific in a
+[Container Device Interface](https://github.com/cncf-tags/container-device-interface/blob/main/SPEC.md)
```
elezar, line 5:
```suggestion
[Container Device Interface](https://tags.cncf.io/container-device-interface/blob/main/SPEC.md)
```

cmd/nvidia-cdi-hook/main.go
```diff
@@ -0,0 +1,99 @@
+/**
+# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
elezar, line 2:
```suggestion
# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```

cmd/nvidia-ctk/cdi/generate/generate.go
elezar: Does:
```
diff --git a/cmd/nvidia-ctk/cdi/generate/generate.go b/cmd/nvidia-ctk/cdi/generate/generate.go
index fba22301..9f9e994b 100644
--- a/cmd/nvidia-ctk/cdi/generate/generate.go
+++ b/cmd/nvidia-ctk/cdi/generate/generate.go
@@ -132,13 +132,11 @@ func (m command) build() *cli.Command {
 			Destination: &opts.librarySearchPaths,
 		},
 		&cli.StringFlag{
-			Name:        "nvidia-cdi-hook-path",
-			Usage:       "Specify the path to use for the nvidia-cdi-hook in the generated CDI specification. If this is left empty, the path will be searched.",
-			Destination: &opts.nvidiaCDIHookPath,
-		},
-		&cli.StringFlag{
-			Name:        "nvidia-ctk-path",
-			Usage:       "Specify the path to use for the nvidia-ctk in the generated CDI specification. If this is left empty, the path will be searched. DEPRECATED. Use flag --nvidia-cdi-hook-path instead.",
+			Name:    "nvidia-cdi-hook-path",
+			Aliases: []string{"nvidia-ctk-path"},
+			Usage: "Specify the path to use for the nvidia-cdi-hook in the generated CDI specification. " +
+				"If not specified, the PATH will be searched for `nvidia-cdi-hook`. " +
+				"NOTE: That if this is specified as `nvidia-ctk`, the PATH will be searched for `nvidia-ctk` instead.",
 			Destination: &opts.nvidiaCDIHookPath,
 		},
 		&cli.StringFlag{

```
make sense?

pkg/nvcdi/lib.go
```diff
@@ -49,6 +49,7 @@ type nvcdilib struct {
 	driverRoot         string
 	devRoot            string
 	nvidiaCTKPath      string
+	nvidiaCDIHookPath  string
```
elezar, line 52: Since we're effectively replacing `nvidiaCTKPath` with this member, I would recommend applying the following diff:
```
diff --git a/internal/modifier/cdi.go b/internal/modifier/cdi.go
index c53cb996..c5af4f88 100644
--- a/internal/modifier/cdi.go
+++ b/internal/modifier/cdi.go
@@ -185,7 +185,7 @@ func newAutomaticCDISpecModifier(logger logger.Interface, cfg *config.Config, de
 func generateAutomaticCDISpec(logger logger.Interface, cfg *config.Config, devices []string) (spec.Interface, error) {
 	cdilib, err := nvcdi.New(
 		nvcdi.WithLogger(logger),
-		nvcdi.WithNVIDIACTKPath(cfg.NVIDIACTKConfig.Path),
+		nvcdi.WithNVIDIACDIHookPath(cfg.NVIDIACTKConfig.Path),
 		nvcdi.WithDriverRoot(cfg.NVIDIAContainerCLIConfig.Root),
 		nvcdi.WithVendor("runtime.nvidia.com"),
 		nvcdi.WithClass("gpu"),
diff --git a/internal/modifier/csv.go b/internal/modifier/csv.go
index 54304428..0905d5da 100644
--- a/internal/modifier/csv.go
+++ b/internal/modifier/csv.go
@@ -62,7 +62,7 @@ func NewCSVModifier(logger logger.Interface, cfg *config.Config, image image.CUD
 	cdilib, err := nvcdi.New(
 		nvcdi.WithLogger(logger),
 		nvcdi.WithDriverRoot(cfg.NVIDIAContainerCLIConfig.Root),
-		nvcdi.WithNVIDIACTKPath(cfg.NVIDIACTKConfig.Path),
+		nvcdi.WithNVIDIACDIHookPath(cfg.NVIDIACTKConfig.Path),
 		nvcdi.WithMode(nvcdi.ModeCSV),
 		nvcdi.WithCSVFiles(csvFiles),
 	)
diff --git a/internal/platform-support/tegra/csv_test.go b/internal/platform-support/tegra/csv_test.go
index 2e4aca34..2e8e42fe 100644
--- a/internal/platform-support/tegra/csv_test.go
+++ b/internal/platform-support/tegra/csv_test.go
@@ -187,7 +187,6 @@ func TestDiscovererFromCSVFiles(t *testing.T) {
 
 			o := tegraOptions{
 				logger:              logger,
-				nvidiaCTKPath:       "/usr/bin/nvidia-ctk",
 				nvidiaCDIHookPath:   "/usr/bin/nvidia-cdi-hook",
 				csvFiles:            []string{"dummy"},
 				ignorePatterns:      tc.ignorePatterns,
diff --git a/internal/platform-support/tegra/tegra.go b/internal/platform-support/tegra/tegra.go
index 597d8b16..1031fc72 100644
--- a/internal/platform-support/tegra/tegra.go
+++ b/internal/platform-support/tegra/tegra.go
@@ -30,7 +30,6 @@ type tegraOptions struct {
 	csvFiles           []string
 	driverRoot         string
 	devRoot            string
-	nvidiaCTKPath      string
 	nvidiaCDIHookPath  string
 	ldconfigPath       string
 	librarySearchPaths []string
@@ -134,13 +133,6 @@ func WithCSVFiles(csvFiles []string) Option {
 	}
 }
 
-// WithNVIDIACTKPath sets the path to the nvidia-container-toolkit binary.
-func WithNVIDIACTKPath(nvidiaCTKPath string) Option {
-	return func(o *tegraOptions) {
-		o.nvidiaCTKPath = nvidiaCTKPath
-	}
-}
-
 // WithNVIDIACDIHookPath sets the path to the nvidia-cdi-hook binary.
 func WithNVIDIACDIHookPath(nvidiaCDIHookPath string) Option {
 	return func(o *tegraOptions) {
diff --git a/pkg/nvcdi/lib-csv.go b/pkg/nvcdi/lib-csv.go
index f6f652ba..649b801a 100644
--- a/pkg/nvcdi/lib-csv.go
+++ b/pkg/nvcdi/lib-csv.go
@@ -44,7 +44,6 @@ func (l *csvlib) GetAllDeviceSpecs() ([]specs.Device, error) {
 		tegra.WithLogger(l.logger),
 		tegra.WithDriverRoot(l.driverRoot),
 		tegra.WithDevRoot(l.devRoot),
-		tegra.WithNVIDIACTKPath(l.nvidiaCTKPath),
 		tegra.WithNVIDIACDIHookPath(l.nvidiaCDIHookPath),
 		tegra.WithLdconfigPath(l.ldconfigPath),
 		tegra.WithCSVFiles(l.csvFiles),
diff --git a/pkg/nvcdi/lib.go b/pkg/nvcdi/lib.go
index 31e08e82..c56c18e1 100644
--- a/pkg/nvcdi/lib.go
+++ b/pkg/nvcdi/lib.go
@@ -48,7 +48,6 @@ type nvcdilib struct {
 	deviceNamers       DeviceNamers
 	driverRoot         string
 	devRoot            string
-	nvidiaCTKPath      string
 	nvidiaCDIHookPath  string
 	ldconfigPath       string
 	configSearchPaths  []string
@@ -88,9 +87,6 @@ func New(opts ...Option) (Interface, error) {
 	if l.devRoot == "" {
 		l.devRoot = l.driverRoot
 	}
-	if l.nvidiaCTKPath == "" {
-		l.nvidiaCTKPath = "/usr/bin/nvidia-ctk"
-	}
 	if l.nvidiaCDIHookPath == "" {
 		l.nvidiaCDIHookPath = "/usr/bin/nvidia-cdi-hook"
 	}
diff --git a/pkg/nvcdi/options.go b/pkg/nvcdi/options.go
index a90450f6..5a490619 100644
--- a/pkg/nvcdi/options.go
+++ b/pkg/nvcdi/options.go
@@ -63,10 +63,10 @@ func WithLogger(logger logger.Interface) Option {
 }
 
 // WithNVIDIACTKPath sets the path to the NVIDIA Container Toolkit CLI path for the library
+//
+// Deprecated: Use WithNVIDIACDIHookPath instead.
 func WithNVIDIACTKPath(path string) Option {
-	return func(l *nvcdilib) {
-		l.nvidiaCTKPath = path
-	}
+	return WithNVIDIACDIHookPath(path)
 }
 
 // WithNVIDIACDIHookPath sets the path to the NVIDIA Container Toolkit CLI path for the library
```

Which switches to using `nvidiaCDIHookPath` everywhere -- except where we read `nvidia-ctk.path` config option` and pass that to the constructor. We also mark the `nvcdi.WithNVIDIACTKPath` option as deprecated (although we could also not do so and just set the member).

The internal usage in the `tegra` package has also been changed to directly refer to `nvidiaCDIHookPath` instead. I don't think we need to deprecat that option since we don't consume this in any of our othe components.

packaging/rpm/SPECS/nvidia-container-toolkit.spec
```diff
@@ -62,6 +64,7 @@ fi
 %files
 %license LICENSE
 %{_bindir}/nvidia-container-runtime-hook
+%{_bindir}/nvidia-cdi-hook
```
elezar, line 67: I think this should be droped from this section and added to the `%files base` section below where `%{_bindir}/nvidia-ctk` is also defined.

elezar (changes requested):
Thanks @deitch. It's looking much better already.

There are some minor changes required.

cmd/nvidia-cdi-hook/README.md
deitch: Done

cmd/nvidia-cdi-hook/README.md
deitch: Done

cmd/nvidia-cdi-hook/README.md
deitch: Done

cmd/nvidia-cdi-hook/main.go
deitch: Done. No time travel.

packaging/rpm/SPECS/nvidia-container-toolkit.spec
deitch: Good point. I should have done that to begin with. Done.

pkg/nvcdi/lib.go
deitch: > make sense?

Sure, a clean alias, and an explanation. Added.

deitch: Done.

deitch:
All comments and suggestions added; thanks for them @elezar . Back to you.

deitch:
Is anything else needed for this @elezar ?

deitch:
Was a golang unit test error. Fixed.

internal/config/config.go
```diff
@@ -186,6 +187,22 @@ func ResolveNVIDIACTKPath(logger logger.Interface, nvidiaCTKPath string) string
 	)
 }
 
+// ResolveNVIDIACDIHookPath resolves the path to the nvidia-cdi-hook binary.
```
elezar, line 190: Looking at this and `ResolveNVIDIACTKPath` above again, what do you think of rewriting this as:
```
// ResolveNVIDIACTKPath resolves the path to the nvidia-ctk binary.
// This executable is used in hooks and needs to be an absolute path.
// If the path is specified as an absolute path, it is used directly
// without checking for existence of an executable at that path.
//
// Deprecated: Use ResolveNVIDIACDIHookPath directly instead.
func ResolveNVIDIACTKPath(logger logger.Interface, nvidiaCTKPath string) string {
	return resolveWithDefault(
		logger,
		"NVIDIA Container Toolkit CLI",
		nvidiaCTKPath,
		nvidiaCTKDefaultFilePath,
	)
}

// ResolveNVIDIACDIHookPath resolves the path to the nvidia-cdi-hook binary.
// This executable is used in hooks and needs to be an absolute path.
// If the path is specified as an absolute path, it is used directly
// without checking for existence of an executable at that path.
func ResolveNVIDIACDIHookPath(logger logger.Interface, nvidiaCDIHookPath string) string {
	if filepath.Base(nvidiaCDIHookPath) == "nvidia-ctk" {
		return resolveWithDefault(
			logger,
			"NVIDIA Container Toolkit CLI",
			nvidiaCDIHookPath,
			nvidiaCTKDefaultFilePath,
		)
	}
	return resolveWithDefault(
		logger,
		"NVIDIA CDI Hook CLI",
		nvidiaCDIHookPath,
		nvidiaCDIHookDefaultFilePath,
	)
}
```

and then adding:
```
	//nolint:staticcheck  // TODO(elezar): We should swith the nvidia-container-runtime from using nvidia-ctk to using nvidia-cdi-hook.
	cfg.NVIDIACTKConfig.Path = config.ResolveNVIDIACTKPath(r.logger, cfg.NVIDIACTKConfig.Path)
```
where we call `ResolveNVIDIACTKPath`?

We can work on properly handling the config changes for the runtime in a follow-up.

elezar (changes requested):
In addition to the comment below, we also need:
```
diff --git a/internal/discover/ldconfig_test.go b/internal/discover/ldconfig_test.go
index 67149845..0b214c77 100644
--- a/internal/discover/ldconfig_test.go
+++ b/internal/discover/ldconfig_test.go
@@ -25,8 +25,8 @@ import (
 )
 
 const (
-	testNvidiaCDIPath = "/foo/bar/nvidia-cdi-path"
-	testLdconfigPath  = "/bar/baz/ldconfig"
+	testNvidiaCDIHookPath = "/foo/bar/nvidia-cdi-hook"
+	testLdconfigPath      = "/bar/baz/ldconfig"
 )
 
 func TestLDCacheUpdateHook(t *testing.T) {
@@ -92,12 +92,12 @@ func TestLDCacheUpdateHook(t *testing.T) {
 				},
 			}
 			expectedHook := Hook{
-				Path:      testNvidiaCDIPath,
+				Path:      testNvidiaCDIHookPath,
 				Args:      tc.expectedArgs,
 				Lifecycle: "createContainer",
 			}
 
-			d, err := NewLDCacheUpdateHook(logger, mountMock, testNvidiaCDIPath, tc.ldconfigPath)
+			d, err := NewLDCacheUpdateHook(logger, mountMock, testNvidiaCDIHookPath, tc.ldconfigPath)
 			require.NoError(t, err)
 
 			hooks, err := d.Hooks()
```

deitch:
> In addition to the comment below, we also need:

Yup already working on that one. Since it causes CI to fail, I get to see it. :-)

internal/config/config.go
deitch: I like it. Change made and pushed.

internal/config/config.go
```diff
@@ -186,6 +189,27 @@ func ResolveNVIDIACTKPath(logger logger.Interface, nvidiaCTKPath string) string
...
+// This executable is used in hooks and needs to be an absolute path.
+// If the path is specified as an absolute path, it is used directly
+// without checking for existence of an executable at that path.
+func ResolveNVIDIACDIHookPath(logger logger.Interface, nvidiaCDIHookPath string) string {
```
elezar, line 196: In responding to https://github.com/cncf-tags/container-device-interface/issues/203#issuecomment-2117628618 I had the thought that the concept of a CDI Hook executable could be standardized so that the `nvidia-cdi-hook` is a vendor-specific implementation. Does that mean that we should use `cdiHookPath` instead of using `nvidiaCDIHookPath` for our variables and config values?

internal/config/config.go
```diff
@@ -186,6 +189,27 @@ func ResolveNVIDIACTKPath(logger logger.Interface, nvidiaCTKPath string) string
...
+		return resolveWithDefault(
+			logger,
+			"NVIDIA Container Toolkit CLI",
+			nvidiaCDIHookPath,
+			nvidiaCTKDefaultFilePath,
+		)
+	}
+	return resolveWithDefault(
+		logger,
+		"NVIDIA Container Toolkit CLI",
```
elezar, line 207:
```suggestion
		"NVIDIA CDI Hook CLI",
```

internal/config/config.go
deitch: I don't think it matters all that much. Until such time as that actually happens, might as well get this in and iterate on it?

internal/config/config.go
deitch: Sure. Done. Also rebased on latest `main`.

elezar:
One final pass @deitch. I have created https://github.com/deitch/nvidia-container-toolkit/pull/1 with a proposal for a changelog addition and also adding a constructor for the hook commands to ensure that these are consistent across the two binaries.

I think with those in we can wrap this up.

deitch:
@elezar do you mind changing [deitch#1](https://github.com/deitch/nvidia-container-toolkit/pull/1) to be a PR against the branch I used for this PR [separate-hook-binary](https://github.com/deitch/nvidia-container-toolkit/tree/separate-hook-binary) rather than against `deitch:main`? That way I can see just the changed files there and merge it into my branch?

elezar:
@deitch updated.

elezar (approved):
Thanks for the work on this @deitch.

I think this is cleaner than having the hooks as a subcommand of the nvidia-ctk -- especially given the discussion around https://github.com/cncf-tags/container-device-interface/issues/203

deitch:
> Thanks for the work on this @deitch

My pleasure. I am glad it worked out.

PR #484, opened 2024-05-03
Use : as a config --set slice separator

### LINKED ISSUE #466: nvidia-ctk config set list still not work on 1.15.0
I notice that this bug has been fixed on 1.15.0 rc1, but it still don't work after I install nvidia-container-toolkit 1.15.0 referencing the official guide. And I wonder if I'm using in wrong way?
```
root@ecs-bf13:~# nvidia-ctk config -i --set nvidia-container-runtime.runtimes=docker-runc,runc,crun
ERRO[0000] invalid --set option runc: invalid config option: undefined field: "runc" 
root@ecs-bf13:~# nvidia-ctk --version
NVIDIA Container Toolkit CLI version 1.15.0
commit: ddeeca392c7bd8b33d0a66400b77af7a97e16cef
```

elezar:
This change switches to using `:` as the separator for slices when calling `nvidia-ctk config --set`. This works around an issue with the handling of string-slice arguments in the `urfave/cli` module.

For example, running:
```
nvidia-ctk config --set nvidia-container-runtime.runtimes=some:list:of:runtimes
```

yields:
```
runtimes = ["some", "list", "of", "runtimes"]
```

Fixes #466

### DIFF
--- a/cmd/nvidia-ctk/config/config.go
+++ b/cmd/nvidia-ctk/config/config.go
@@ -38,7 +38,8 @@ type command struct {
 // options stores the subcommand options
 type options struct {
 	flags.Options
-	sets cli.StringSlice
+	setListSeparator string
+	sets             cli.StringSlice
 }
 
 // NewCommand constructs an config command with the specified logger
@@ -57,6 +58,9 @@ func (m command) build() *cli.Command {
 	c := cli.Command{
 		Name:  "config",
 		Usage: "Interact with the NVIDIA Container Toolkit configuration",
+		Before: func(ctx *cli.Context) error {
+			return validateFlags(ctx, &opts)
+		},
 		Action: func(ctx *cli.Context) error {
 			return run(ctx, &opts)
 		},
@@ -71,9 +75,20 @@ func (m command) build() *cli.Command {
 			Destination: &opts.Config,
 		},
 		&cli.StringSliceFlag{
-			Name:        "set",
-			Usage:       "Set a config value using the pattern key=value. If value is empty, this is equivalent to specifying the same key in unset. This flag can be specified multiple times",
+			Name: "set",
+			Usage: "Set a config value using the pattern 'key[=value]'. " +
+				"Specifying only 'key' is equivalent to 'key=true' for boolean settings. " +
+				"This flag can be specified multiple times, but only the last value for a specific " +
+				"config option is applied. " +
+				"If the setting represents a list, the elements are colon-separated.",
 			Destination: &opts.sets,
+		},
+		&cli.StringFlag{
+			Name:        "set-list-separator",
+			Usage:       "Specify a separator for lists applied using the set command.",
+			Hidden:      true,
+			Value:       ":",
+			Destination: &opts.setListSeparator,
 		},
 		&cli.BoolFlag{
 			Name:        "in-place",
@@ -96,6 +111,13 @@ func (m command) build() *cli.Command {
 	return &c
 }
 
+func validateFlags(c *cli.Context, opts *options) error {
+	if opts.setListSeparator == "" {
+		return fmt.Errorf("set-list-separator must be set")
+	}
+	return nil
+}
+
 func run(c *cli.Context, opts *options) error {
 	cfgToml, err := config.New(
 		config.WithConfigFile(opts.Config),
@@ -105,7 +127,7 @@ func run(c *cli.Context, opts *options) error {
 	}
 
 	for _, set := range opts.sets.Value() {
-		key, value, err := setFlagToKeyValue(set)
+		key, value, err := setFlagToKeyValue(set, opts.setListSeparator)
 		if err != nil {
 			return fmt.Errorf("invalid --set option %v: %w", set, err)
 		}
@@ -139,7 +161,7 @@ var errInvalidFormat = errors.New("invalid format")
 // setFlagToKeyValue converts a --set flag to a key-value pair.
 // The set flag is of the form key[=value], with the value being optional if key refers to a
 // boolean config option.
-func setFlagToKeyValue(setFlag string) (string, interface{}, error) {
+func setFlagToKeyValue(setFlag string, setListSeparator string) (string, interface{}, error) {
 	setParts := strings.SplitN(setFlag, "=", 2)
 	key := setParts[0]
 
@@ -172,7 +194,7 @@ func setFlagToKeyValue(setFlag string) (string, interface{}, error) {
 	case reflect.String:
 		return key, value, nil
 	case reflect.Slice:
-		valueParts := strings.Split(value, ",")
+		valueParts := strings.Split(value, setListSeparator)
 		switch field.Elem().Kind() {
 		case reflect.String:
 			return key, valueParts, nil
--- a/cmd/nvidia-ctk/config/config_test.go
+++ b/cmd/nvidia-ctk/config/config_test.go
@@ -25,11 +25,12 @@ import (
 func TestSetFlagToKeyValue(t *testing.T) {
 	// TODO: We need to enable this test again since switching to reflect.
 	testCases := []struct {
-		description   string
-		setFlag       string
-		expectedKey   string
-		expectedValue interface{}
-		expectedError error
+		description      string
+		setFlag          string
+		setListSeparator string
+		expectedKey      string
+		expectedValue    interface{}
+		expectedError    error
 	}{
 		{
 			description:   "option not present returns an error",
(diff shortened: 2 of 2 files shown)


klueska:
I glanced through the linked issue but its still not clear to me why `,` doesn't work as the separator.

elezar:
@klueska they behaviour for `StringSliceFlags` was changed in urfave v2.5.1. This was to fix a bug for properly merging repeated flags. We were however depending on the former behaviour where we allow a flag to be repeated, but don't split their values. This meant that this functionality broke when we updated urfave as part of our release process.

There may be a more invasive change that I could make here, but I thought allowing the separator to be customized was a low-risk change.

klueska:
Do you mean `v2.25.1`?  Is this the change that broke things for us: https://github.com/urfave/cli/pull/1710?

cmd/nvidia-ctk/config/config.go
```diff
@@ -72,9 +76,16 @@ func (m command) build() *cli.Command {
 		},
 		&cli.StringSliceFlag{
 			Name:        "set",
-			Usage:       "Set a config value using the pattern key=value. If value is empty, this is equivalent to specifying the same key in unset. This flag can be specified multiple times",
+			Usage:       "Set a config value using the pattern key=value. If value is empty, this is equivalent to specifying the same key in unset. This flag can be specified multiple times. If the setting represents a slice, the elements are semi-colon-separated.",
```
klueska, line 79:
```suggestion
			Usage:       "Set a config value using the pattern key=value. If value is empty, this is equivalent to specifying the same key in unset. This flag can be specified multiple times. If the setting represents a list, the elements are semi-colon-separated.",
```

cmd/nvidia-ctk/config/config.go
```diff
@@ -72,9 +76,16 @@ func (m command) build() *cli.Command {
...
 			Destination: &opts.sets,
 		},
+		&cli.StringFlag{
+			Name:        "set-slice-separator",
+			Usage:       "Specify a separator for slices applied using the set command.",
```
klueska, line 84:
```suggestion
			Name:        "set-list-separator",
			Usage:       "Specify a separator for lists applied using the set command.",
```

cmd/nvidia-ctk/config/config.go
klueska: Was semicolon the default previously? I'm still trying to wrap my head around the previous behaviour vs. the new behaviour.

What I _think_ you are saying in another comment is that we previously we always interpreted every input as a single string, and if it was *supposed* to be slice, we did the separation ourselves. What's not clear (assuming I got this right) is whether we previously did the separation with semicolons or commas, and (if it was commas), why we con't continue to use commas now instead of semilcolons.

elezar:
> Do you mean `v2.25.1`? Is this the change that broke things for us: [urfave/cli#1710](https://github.com/urfave/cli/pull/1710)?

No, I mean `v2.5.1`. We were using `v2.3.0` when we originally released this functionality and updated directly to `v2.27.1` in https://github.com/NVIDIA/nvidia-container-toolkit/pull/309. The following PR https://github.com/urfave/cli/pull/1377 was released as part fo `v2.5.1` and after testing urfave updates after `v2.3.0` locally found this to be the culprit.

cmd/nvidia-ctk/config/config.go
elezar: Yes, we were previously treating a comma-separated argument as a single string (due to a bug in urfave) and we were splitting this by commas ourselves to construct a slice to insert into the config file. We can't continue to use commas now because this urfave already assumes commas as the separator for repeated args. We could consider overriding this instead, but we can only do this at an Application level and not at a Command level. It may be possible with the alpha v3.0.0 release, but I didn't want to rely on an unstable dependency.

Consider the following code:
```
package main

import (
	"fmt"
	"os"

	"github.com/urfave/cli/v2"
)

func main() {
	var slice cli.StringSlice

	c := cli.NewApp()
	c.Flags = []cli.Flag{
		&cli.StringSliceFlag{
			Name:        "slice-flag",
			Destination: &slice,
		},
	}

	c.Action = func(ctx *cli.Context) error {
		fmt.Printf("slice=%q\n", slice.Value())
		return nil
	}

	c.Run(os.Args)
}
```

When using `v2.5.0` we get:
```
$ go run ./main.go --slice-flag=foo,bar,baz
slice=["foo,bar,baz"]
$ go run ./main.go --slice-flag=foo,bar,baz --slice-flag=boom
slice=["foo,bar,baz" "boom"]
```

When updating to `v2.5.1` we get:
```
$ go run ./main.go --slice-flag=foo,bar,baz
slice=["foo" "bar" "baz"]
$ go run ./main.go --slice-flag=foo,bar,baz --slice-flag=boom
slice=["foo" "bar" "baz" "boom"]
```
clearly indicating that a bug was fixed for handling comma-separated repeated arguments when updating to at least `v2.5.1`.

We were however depending on the broken behaviour, meaning that when a user ran:
```
nvidia-ctk config --set nvidia-container-runtime.runtimes=crun,runc
```
we had a slice with the following contents:
```
[]string{"nvidia-container-runtime.runtimes=crun,runc"}
```
When we could process -- first splitting by `=` and then the second argument by `,` to generate the eventually slice `[]string{"crun", "runc"}`.

With the update to `v2.27.1` when a user runs:
```
nvidia-ctk config --set nvidia-container-runtime.runtimes=crun,runc
```
we have a slice with the following contents:
```
[]string{"nvidia-container-runtime.runtimes=crun" ,"runc"}
```
Which we cannot effectively parse.

With the switch to `;` as our own separator, we no have the inputs:
```
nvidia-ctk config --set nvidia-container-runtime.runtimes=crun;runc
```
which would be parsed to:
```
[]string{"nvidia-container-runtime.runtimes=crun;runc"}
```
meaning that we can do the same processing as before.

klueska: OK, so this is defining the separator we want to use to break apart elements of an *embedded* list of strings that we don't want to become elements of the overall `StringSlice`.

klueska: Meaning that `,` is still the default for the top-level separator, and we now defining `;` as the embedded separator.

klueska: I think this is fine then -- my only worry would be if / where else we rely on this "bug". I seem to remember relying on this functionality in the device-plugin as well, for example.

klueska: Seems reasonable to me

ArangoGutierrez (commented):
PR now looks ok to me, and addresses @klueska comments.
Only comment please update the PR description which still uses `;`

cmd/nvidia-ctk/config/config.go
```diff
@@ -96,6 +111,13 @@ func (m command) build() *cli.Command {
 	return &c
 }
 
+func validateFlags(c *cli.Context, opts *options) error {
+	if opts.setSliceSeparator == "" {
+		return fmt.Errorf("set-slice-separator must be set")
```
klueska, line 116:
```suggestion
		return fmt.Errorf("set-list-separator must be set")
```

klueska (commented):
Minor comments. LGTM otherwise

cmd/nvidia-ctk/config/config.go
```diff
@@ -71,10 +75,21 @@ func (m command) build() *cli.Command {
...
+				"config option is applied. " +
+				"If the setting represents a list, the elements are colon-separated.",
 			Destination: &opts.sets,
 		},
+		&cli.StringFlag{
+			Name:        "set-list-separator",
+			Usage:       "Specify a separator for lists applied using the set command.",
+			Hidden:      true,
+			Value:       ":",
+			Destination: &opts.setSliceSeparator,
```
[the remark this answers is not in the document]
elezar: Sure. Will update.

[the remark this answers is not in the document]
elezar: Done.

cmd/nvidia-ctk/config/config.go
elezar: Done.

ArangoGutierrez (approved):

cmd/nvidia-ctk/config/config.go
```diff
@@ -71,10 +75,21 @@ func (m command) build() *cli.Command {
...
+			Name: "set",
+			Usage: "Set a config value using the pattern 'key[=value]'. " +
+				"Specifying only 'key' is equivalent to 'key=true' for boolean settings. " +
+				"This flag can be specified multiple times, but only the last value for a specific " +
+				"config option is applied. " +
+				"If the setting represents a list, the elements are colon-separated.",
 			Destination: &opts.sets,
 		},
+		&cli.StringFlag{
+			Name:        "set-list-separator",
```
tariq1890, line 87: I think the `set` is redundant. When we specify `--list-separator` , it is pretty much implied that we intend to use the argument as the list separator.

elezar: The `set` is not reduntant in this case as it refers to the `set` flag. We are specifying the list separator for the set flag.

tariq1890: Ohh got it.

tariq1890 (approved):

PR #490, opened 2024-05-10
Inject additional libraries required for full display functionality

ehfd:
Requested reviewer: @elezar 

This PR tries the fix the remainder of issues which prevent proper X11 deployment in containers.

Largely, two things:

- Add `$LD_LIBRARY_PATH/gbm/nvidia-drm_gbm.so` and other related libraries including `libnvidia-egl-wayland.so.1` not already included by the container toolkit.
- Make the container toolkit search for more paths than `$LD_LIBRARY_PATH/nvidia/xorg/libglxserver_nvidia.so` and `$LD_LIBRARY_PATH/nvidia/xorg/nvidia_drv.so` for the critical X11 libraries. This is because NVIDIA drivers installed with `.run` installs the libraries into `/usr/lib/xorg/modules/extensions/libglxserver_nvidia.so.530.41.03` and `/usr/lib/xorg/modules/drivers/nvidia_drv.so`, as specified in https://download.nvidia.com/XFree86/Linux-x86_64/550.78/README/installedcomponents.html.

Please edit the PR as necessary.

Few things of note:

`libnvidia-vulkan-producer.so` is required before r550 but is removed with r550.


internal/discover/graphics.go
```diff
@@ -271,13 +274,21 @@ func newXorgDiscoverer(logger logger.Interface, driver *root.Driver, nvidiaCTKPa
...
 			lookup.WithRoot(driver.Root),
-			lookup.WithSearchPaths(libRoot, "/usr/lib/x86_64-linux-gnu"),
+			lookup.WithSearchPaths([]string{
+				libRoot + "/nvidia/xorg",
+				libRoot + "/xorg/modules/extensions",
+				"/usr/lib/xorg/modules/extensions",
+				"/usr/X11R6/modules/extensions",
+				libRoot + "/xorg/modules/drivers",
+				"/usr/lib/xorg/modules/drivers",
+				"/usr/X11R6/modules/drivers",
```
elezar, line 284: First question: Is the ordering of these important? Does it make sense to group the `libRoot` search paths?

Second: Are any of these paths controllable by envvars (such as `XDG_DATA_DIRS` for icd loaders)?

internal/discover/graphics.go
```diff
@@ -271,13 +274,21 @@ func newXorgDiscoverer(logger logger.Interface, driver *root.Driver, nvidiaCTKPa
 		lookup.NewFileLocator(
 			lookup.WithLogger(logger),
 			lookup.WithRoot(driver.Root),
-			lookup.WithSearchPaths(libRoot, "/usr/lib/x86_64-linux-gnu"),
+			lookup.WithSearchPaths([]string{
+				libRoot + "/nvidia/xorg",
```
elezar, line 278: If `libcuda.so.<RM_VERSION>` is at `/usr/lib/x86_64-linux-gnu`, for example, libRoot will be `/usr/lib/x86_64-linux-gnu` is this the intent? Or should this always be `/usr/lib/...` or `/usr/X11R/...`?

internal/discover/graphics.go
```diff
@@ -56,6 +56,9 @@ func NewGraphicsMountsDiscoverer(logger logger.Interface, driver *root.Driver, n
 		driver.Root,
 		[]string{
 			"libnvidia-egl-gbm.so.*",
+			"gbm/nvidia-drm_gbm.so",
+			"libnvidia-egl-wayland.so.*",
+			"libnvidia-vulkan-producer.so",
```
elezar, line 61: `libnvidia-vulkan-producer.so` is a symlink which should resolve to `libnvidia-vulkan-producer.so.RM_VERSION`. The latter should be detected by the "standard" driver libraries unless it's at a different location to `libcuda.so.RM_VERSION`. 

Is the issue that the `.RM_VERSION` library is not injected, or that the symlink is not created?

Looking at the driver manifest we have:
```
libnvidia-vulkan-producer.so.535.86.10 0755 OPENGL_LIB NATIVE MODULE:egl
libnvidia-vulkan-producer.so 0000 UTILITY_LIB_SYMLINK NATIVE libnvidia-vulkan-producer.so.535.86.10 MODULE:egl
```

elezar (commented):
Thanks @ehfd.

Some questions to start with.

elezar:
If you could provide some steps on how you would go about validating correct functionality, that would also be great -- we are looking to improve our automated test coverage and having a starting point would be welcome.

internal/discover/graphics.go
ehfd: 1. Ordering is not important. I moved the libroot paths together.

2. Envvars: ... I don't really think there's an envvar specifically for where the X.Org modules are.
The additional paths are because these libraries are not guaranteed to be installed in `libRoot` for some types of installations including `.run`, so they have to be hard-coded.

internal/discover/graphics.go
ehfd: Yes. This is the intent. I removed `"/usr/lib/x86_64-linux-gnu"` because it's not arch-neutral and redundant with libroot.

In Ubuntu's APT installation for NVIDIA drivers, the paths where these packages are installed resolve to `/usr/lib/x86_64-linux-gnu/nvidia/xorg` for both X11-related .so files.

internal/discover/graphics.go
ehfd: Neither are injected. Is it ok if I simply put `"libnvidia-vulkan-producer.so*",`?

ehfd: `Looking at the driver manifest we have:`

Where can I see this?

ehfd:
> If you could provide some steps on how you would go about validating correct functionality, that would also be great -- we are looking to improve our automated test coverage and having a starting point would be welcome.

This is a pretty hard question...

The first priority regardless is to consult the NVIDIA driver team on structuring libraries and these things. Like a channel which you can get informed of library structures and such related aspects of the driver.

The second is to use different types of driver installation methods; Ubuntu/Debian's PPA, Arch, RHEL/Fedora-based, and the .run file, and check that the number of libraries provisioned into the container are consistent.

internal/discover/graphics.go
ehfd: Check the below comment for an answer to this: https://github.com/NVIDIA/nvidia-container-toolkit/pull/490#issuecomment-2104836490

internal/discover/graphics.go
elezar: The manifest is available when extracting a driver `.run` file.

> Neither are injected. Is it ok if I simply put "libnvidia-vulkan-producer.so*",?

No, since we need to treat symlinks correctly. I assume that the issue is then that `libnvidia-vulkan-producer.so.<RM_VERSION>` doesn't exist at the same location as `libcuda.so.<RM_VERSION>` on your system. Can you confirm this?

ehfd: I believe so.

internal/discover/graphics.go
```diff
@@ -56,6 +56,9 @@ func NewGraphicsMountsDiscoverer(logger logger.Interface, driver *root.Driver, n
 		driver.Root,
 		[]string{
 			"libnvidia-egl-gbm.so.*",
+			"gbm/nvidia-drm_gbm.so",
+			"libnvidia-egl-wayland.so.*",
+			"libnvidia-vulkan-producer.so*",
```
elezar, line 61: We want:
```suggestion
			"libnvidia-vulkan-producer.so.*",
```
and need to create the `.so` symlink if required as a hook.

internal/discover/graphics.go
```diff
@@ -271,13 +274,33 @@ func newXorgDiscoverer(logger logger.Interface, driver *root.Driver, nvidiaCTKPa
 		lookup.NewFileLocator(
 			lookup.WithLogger(logger),
 			lookup.WithRoot(driver.Root),
-			lookup.WithSearchPaths(libRoot, "/usr/lib/x86_64-linux-gnu"),
+			lookup.WithSearchPaths([]string{
+				libRoot + "/nvidia/xorg",
+				libRoot + "/xorg/modules/extensions",
+				libRoot + "/xorg/modules/drivers",
+				"/usr/lib/xorg/modules/extensions",
+				"/usr/lib/xorg/modules/drivers",
```
elezar, line 282: This looks like we're searching the following 2 paths: `[]string{"extensions", "drivers"}` in `[]string{"modules", "modules/updates"}` across the following "roots": `[]string{libRoot + "/xorg", "/usr/lib", "/usr/lib64", "/usr/X11R6"}`.

Does it make sense to construct this dynamically?

elezar (changes requested):

internal/discover/graphics.go
ehfd: Yes, it would.

ehfd: Feel free to push to my branch.

ehfd: @elezar How's the progress?

ehfd:
Another update: `gbm/nvidia-drm_gbm.so` seems to be a symlink to `libnvidia-allocator.so.1`.

https://github.com/games-on-whales/wolf/issues/52#issuecomment-2143612304

@elezar

elezar:
Thanks for the update @ehfd. I will have a look at this again this week.

internal/discover/graphics.go
tux-rampage: Addressed this one

internal/discover/graphics.go
tux-rampage: Symlinks should be created via hook. A bind mount may fail.

I can rewrite this for the vulkan producer and `gbm/nvidia-drm_gbm.so`

internal/discover/graphics.go
```diff
@@ -74,17 +77,75 @@ func NewGraphicsMountsDiscoverer(logger logger.Interface, driver *root.Driver, n
...
+	for _, mount := range mounts {
+		filename := filepath.Base(mount.HostPath)
+
+		// nvidia-drm_gbm.so is a symlink to libnvidia-allocator.so
+		// Make sure this is actually available
+		if strings.HasPrefix(filename, "libnvidia-allocator.so.") {
+			linkDir := filepath.Dir(mount.Path)
+			linkPath := filepath.Join(linkDir, "gbm", "nvidia-drm_gbm.so")
+			target := filepath.Join("..", filename)
+			links = append(links, fmt.Sprintf("%s::%s", target, linkPath))
```
tux-rampage, line 129: @elezar I changed this to a symlink hook, but I'm not sure if this should be done this way.

When doing it via `NewMounts` as before, the locator will attempt to resolve the symlink on the host side. That means the path gets only injected when the library actually exists on the host and future changes to this path (link target changes, becoming a regular file, etc...) will also be covered.

So I'd advocate to put this back to `NewMounts` as originally suggested by @ehfd 

Your Thoughts?

internal/discover/graphics.go
tux-rampage: > Symlinks should be created via hook. A bind mount may fail.

After looking at the code, this is not true here. The driver locator will take care of symlinks and resolve them on the host side.

ehfd:
Thank you for the contributions on the Go code! @tux-rampage

ehfd:
@tux-rampage Note that something like `Signed-off-by: Name <email@users.noreply.github.com>` is required in your commits in order to be merged into this repository. Might need to rebase.

ehfd:
> In conclusion, this is a really complex issue, where this PR did only partially solve the whole issue.
> 
> For example, on top of this PR, it would be ideal if it's possible to push `nvidia_drv.so` and `libglxserver_nvidia.so.*` into `/usr/lib/x86_64-linux-gnu/nvidia/xorg/` or (in a different notation) `libRoot + /nvidia/xorg/` inside the container, regardless of where it was found, and the container toolkit to generate a `/usr/share/X11/xorg.conf.d/10-nvidia.conf` file with the following content regardless of whether `10-nvidia.conf` exists in the host:
> 
> ```bash
> Section "OutputClass"
>     Identifier "nvidia"
>     MatchDriver "nvidia-drm"
>     Driver "nvidia"
>     Option "AllowEmptyInitialConfiguration"
>     ModulePath "/usr/lib/x86_64-linux-gnu/nvidia/xorg"
> EndSection
> ```
> 
> The above is the default behavior for the `nvidia-driver-550` from Ubuntu APT.
> 
> ---
> 
> `nvidia-xconfig` (and also possibly `nvidia-config`, combined with the NVIDIA GTK libraries and `libnvidia-wayland-client.so` which are dependencies) should also be pushed into the container for a full X11 experience.
> 
> Then, the X11 aspect will be solved and we developers can call it a day.
> 
> Also, injecting 32-bit libraries is definitely desirable for usage with Wine/Proton/etc.

Some thoughts or contributions on this part of the issue can also help a lot. @tux-rampage

internal/discover/graphics.go
ehfd: This depends on @elezar 's policies on this matter.

elezar: I don't think that adding `gbm/nvidia-drm_gbm.so` to the `NewMounts` will have the desired effect. Since this resolves to `libnvidia-allocator.so` we will only include `libnvidia-allocator.so` in the list of mounts and NOT create the symlink.

Note that you comment on handling changes in this path is not 100% accurate. While this is true for `gbm/nvidia-drm_gbm.so`, changing this to a regular file would have the downside that we may lose the resolution of the target library in this case.

Your comment does raise an interesting question though. If `gbm/nvidia-drm_gbm.so` exists on the host, then we could follow a similar strategy to what we do for Tegra-based systems [here](https://github.com/NVIDIA/nvidia-container-toolkit/blob/dfc8e22e12fccde4afcde6ce21f2fc84a64d98ec/internal/platform-support/tegra/csv.go#L62-L68). 

Where we distinguish between regular libraries and symlinks. In both cases any symlinks are followed to ensure that we only mount regular files into a container, but in the case of symlinks we recreate the entire symlink chain in the container. If we were to factor out this logic to be reused here, we could treat the symlinks here explicitly as separate entities to discover.

I do think that thsi can be done as a follow-up to this PR though.

ehfd: I agree that merging first is better since this is taking such a long time.

@elezar So are you satisfied with his changes?

tux-rampage:
> Some thoughts or contributions on this part of the issue can also help a lot. @tux-rampage

Sure. I'll take a look when I have some time available, but that might take a day or two.

internal/discover/graphics.go
```diff
@@ -251,6 +312,26 @@ func optionalXorgDiscoverer(logger logger.Interface, driver *root.Driver, nvidia
 	return xorg
 }
 
+func buildXOrgSearchPaths(libRoot string) []string {
```
elezar, line 315: Now that I see this as the nested for loops, it seems more difficult to actually understand what's being searched.

Does it make sense to flatten this instead and just do something like:
```
	return []string{
		filepath.Join(libRoot, "nvidia/xorg"),
		filepath.Join(libRoot, "xorg", "modules", "drivers"),
		filepath.Join(libRoot, "xorg", "modules", "extensions"),
		filepath.Join(libRoot, "xorg", "modules/updates", "drivers"),
		filepath.Join(libRoot, "xorg", "modules/updates", "extensions"),
		filepath.Join("/usr/lib/xorg", "modules", "drivers"),
		filepath.Join("/usr/lib/xorg", "modules", "extensions"),
		filepath.Join("/usr/lib/xorg", "modules/updates", "drivers"),
		filepath.Join("/usr/lib/xorg", "modules/updates", "extensions"),
		...         
	}
```

internal/discover/graphics.go
elezar: @ehfd I have just tried to push to your branch, but I'm getting permission errors.

I have rebased your branch off `main` and created https://github.com/ehfd/nvidia-container-toolkit/pull/1/files with my proposed changes.

internal/discover/graphics.go
ehfd: I guess so.

internal/discover/graphics.go
ehfd: @elezar I added you to the repository for write access. Merge when you're ready.

elezar:
@tux-rampage @ehfd I have added some commits on top of this branch to reorganise things a bit. I don't really have a great way to verify the changes on my end -- most of my development is headless and compute focussed -- so please check the latest iteration so that we can move forward.

@ehfd there are a number of steps in https://github.com/NVIDIA/nvidia-container-toolkit/pull/490#issuecomment-2148884758 tha I think should be follow-ups for this. Please create an issue where these are tracked as tasks or equivalent so that we can also look at them.

ehfd:
> there are a number of steps in https://github.com/NVIDIA/nvidia-container-toolkit/pull/490#issuecomment-2148884758 tha I think should be follow-ups for this. Please create an issue where these are tracked as tasks or equivalent so that we can also look at them.

I will. I also prefer merging quickly.

ehfd:
I did my testing after compiling, and after initial tests, the libraries are pushed and most of the things (outside the ones that should be follow-ups) work. @elezar

elezar (approved):

PR #491, opened 2024-05-13
feat: Introduce shell metacharacter escaping for exec

servusdei2018:
In order to mitigate potential security vulnerabilities arising from shell injection attacks, this PR introduces a function to escape shell metacharacters which may be present in command-line arguments.

It's worth noting that two potential vulnerabilities still exist in https://github.com/NVIDIA/nvidia-container-toolkit/blob/973a6633b30a95e27aff4ef07772387aac961395/tools/container/nvidia-toolkit/run.go#L249-L252 and https://github.com/NVIDIA/nvidia-container-toolkit/blob/973a6633b30a95e27aff4ef07772387aac961395/tools/container/nvidia-toolkit/run.go#L275-L278 where the string `o.runtimeArgs` is directly interpolated into `cmdline`, leaving it susceptible to injection attacks. To address this, a more comprehensive solution would involve reimplementing `o.runtimeArgs` as `[]string`, allowing for proper sanitization using the introduced `oci.Escape()` function; however, this likely involves a breaking change.

### DIFF
--- a/cmd/nvidia-container-runtime/main_test.go
+++ b/cmd/nvidia-container-runtime/main_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/modifier"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/oci"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/test"
 )
 
@@ -87,8 +88,7 @@ func TestBadInput(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	//nolint:gosec // TODO: Can we harden this so that there is less risk of command injection
-	cmdCreate := exec.Command(nvidiaRuntime, "create", "--bundle")
+	cmdCreate := exec.Command(oci.Escape1(nvidiaRuntime), oci.Escape([]string{"create", "--bundle"})...) //nolint:gosec
 	t.Logf("executing: %s\n", strings.Join(cmdCreate.Args, " "))
 	err = cmdCreate.Run()
 	require.Error(t, err, "runtime should return an error")
@@ -105,8 +105,8 @@ func TestGoodInput(t *testing.T) {
 		t.Fatalf("error generating runtime spec: %v", err)
 	}
 
-	//nolint:gosec // TODO: Can we harden this so that there is less risk of command injection
-	cmdRun := exec.Command(nvidiaRuntime, "run", "--bundle", cfg.bundlePath(), "testcontainer")
+	//nolint:gosec
+	cmdRun := exec.Command(oci.Escape1(nvidiaRuntime), oci.Escape([]string{"run", "--bundle", cfg.bundlePath(), "testcontainer"})...)
 	t.Logf("executing: %s\n", strings.Join(cmdRun.Args, " "))
 	output, err := cmdRun.CombinedOutput()
 	require.NoErrorf(t, err, "runtime should not return an error", "output=%v", string(output))
@@ -116,8 +116,8 @@ func TestGoodInput(t *testing.T) {
 	require.NoError(t, err, "should be no errors when reading and parsing spec from config.json")
 	require.Empty(t, spec.Hooks, "there should be no hooks in config.json")
 
-	//nolint:gosec // TODO: Can we harden this so that there is less risk of command injection
-	cmdCreate := exec.Command(nvidiaRuntime, "create", "--bundle", cfg.bundlePath(), "testcontainer")
+	//nolint:gosec
+	cmdCreate := exec.Command(oci.Escape1(nvidiaRuntime), oci.Escape([]string{"create", "--bundle", cfg.bundlePath(), "testcontainer"})...)
 	t.Logf("executing: %s\n", strings.Join(cmdCreate.Args, " "))
 	err = cmdCreate.Run()
 	require.NoError(t, err, "runtime should not return an error")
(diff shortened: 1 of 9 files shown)


servusdei2018:
cc @elezar

elezar (commented):
My one question would be whether the logic that we're implementing here exists in some third-pary package somewhere. I'm sure that `runc` for example needs to also check the arguments that are passed to its hooks.

servusdei2018:
Hmm I believe I did look at `runc` and couldn't find the exact logic they used, I looked at Docker and they use [shellescape](https://github.com/alessio/shellescape) internally.

ArangoGutierrez:
Hey @servusdei2018, let's revise this valuable contribution. Recently, we have done a lot of hardening work, so it's worth revisiting your contribution, starting with a rebase

servusdei2018:
Rebased @ArangoGutierrez

internal/oci/runtime_syscall_exec_test.go
```diff
@@ -0,0 +1,55 @@
+/*
+# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
ArangoGutierrez:
```suggestion
# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```

internal/oci/runtime_syscall_exec.go
```diff
@@ -41,3 +53,22 @@ func (r syscallExec) Exec(args []string) error {
 func (r syscallExec) String() string {
 	return "exec"
 }
+
+// Escape1 escapes shell metacharacters in a single command-line argument.
+func Escape1(arg string) string {
+	if strings.ContainsAny(arg, shellMetachars) {
+		e := metacharRegex.ReplaceAllString(arg, `\$1`)
+		return fmt.Sprintf(`"%s"`, e)
```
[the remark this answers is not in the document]
ArangoGutierrez: simply 
```go
return metacharRegex.ReplaceAllString(arg, `\$1`)
```

internal/oci/runtime_syscall_exec.go
```diff
@@ -19,16 +19,28 @@ package oci
...
+	"strings"
 	"syscall"
 )
 
+// shellMetachars represents a set of shell metacharacters that are commonly
+// used for shell scripting and may lead to security vulnerabilities if not
+// properly handled.
+//
+// These metacharacters include: | & ; ( ) < > \t \n $ \ `
+const shellMetachars = "|&;()<> \t\n$\\`"
```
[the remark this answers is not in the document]
ArangoGutierrez: please look into this

internal/oci/runtime_syscall_exec.go
```diff
@@ -41,3 +53,22 @@ func (r syscallExec) Exec(args []string) error {
 func (r syscallExec) String() string {
 	return "exec"
 }
+
+// Escape1 escapes shell metacharacters in a single command-line argument.
+func Escape1(arg string) string {
```
ArangoGutierrez: theres' no need for this func to be public, also the name could be something more saying than `escape1`

ArangoGutierrez (changes requested):

internal/oci/runtime_syscall_exec.go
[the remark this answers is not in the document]
servusdei2018: Looks like adding the double- and single-quote don't break anything, so I went ahead and added them to `shellMetachars`.

internal/oci/runtime_syscall_exec.go
```diff
@@ -41,3 +53,21 @@ func (r syscallExec) Exec(args []string) error {
 func (r syscallExec) String() string {
 	return "exec"
 }
+
+// EscapeArg escapes shell metacharacters in a single command-line argument.
+func EscapeArg(arg string) string {
```
ArangoGutierrez:
```suggestion
func escapeArg(arg string) string {
```

ArangoGutierrez: I don't see why this needs to be public

ArangoGutierrez (changes requested):
Can you prepend to the commits that start with `nit`  a prefix `[no-relnote]` , thanks

internal/oci/runtime_syscall_exec.go
servusdei2018: It's being referenced in https://github.com/servusdei2018/nvidia-container-toolkit/blob/b491a15571355cd6ea677dcc2abe958228cdab1e/cmd/nvidia-container-runtime/main_test.go#L102 to escape any metachars in `nvidiaRuntime` (which is currently a constant, but technically a variable as far as golint sees)

cmd/nvidia-container-runtime/main_test.go
```diff
@@ -105,8 +105,8 @@ func TestGoodInput(t *testing.T) {
 		t.Fatalf("error generating runtime spec: %v", err)
 	}
 
-	//nolint:gosec // TODO: Can we harden this so that there is less risk of command injection
-	cmdRun := exec.Command(nvidiaRuntime, "run", "--bundle", cfg.bundlePath(), "testcontainer")
+	//nolint:gosec
+	cmdRun := exec.Command(oci.EscapeArg(nvidiaRuntime), oci.Escape([]string{"run", "--bundle", cfg.bundlePath(), "testcontainer"})...)
```
ArangoGutierrez:
```suggestion
	cmdRun := exec.Command(oci.Escape(nvidiaRuntime)[0], oci.Escape([]string{"run", "--bundle", cfg.bundlePath(), "testcontainer"})...)
```

This way we could have `EscapeArg` private

servusdei2018: That works too! Reverted

rajatchopra:
Closing this PR.
Reasoning: Shell escaping is needed when a shell interpreter is invoked. Most of the execution here happens directly (clone/fork).

Some escaping may still be required if we can verify that the executable involves the shell also.
Please re-open the PR if anyone feels otherwise.

PR #506, opened 2024-05-21
Set mininum spec version on save

elezar:
This change allows an in-memory spec to have its minimum version set on save.

This ensures that this version is updated for subsequent transforms such as root edits.

### DIFF
--- a/pkg/nvcdi/spec/builder.go
+++ b/pkg/nvcdi/spec/builder.go
@@ -39,6 +39,8 @@ type builder struct {
 	mergedDeviceOptions []transform.MergedDeviceOption
 	noSimplify          bool
 	permissions         os.FileMode
+
+	transformOnSave transform.Transformer
 }
 
 // newBuilder creates a new spec builder with the supplied options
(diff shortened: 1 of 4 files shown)


pkg/nvcdi/spec/builder.go
```diff
@@ -55,7 +56,7 @@ func newBuilder(opts ...Option) *builder {
 	}
 
 	if s.version == "" {
-		s.version = DetectMinimumVersion
+		s.version = string(DetectMinimumVersion)
```
ArangoGutierrez, line 59: Wouldn't it be better to a a string method to `minimum` type?

ArangoGutierrez (commented):
Quick question

pkg/nvcdi/spec/builder.go
elezar: I don't know whether it would be better. The implementation would be:
```
func (v minimum) String() string {
    return string(v)
}
```
which would cast anyway.

ArangoGutierrez: "better" was a wrong word, yeah same functionality. I was just wondering if it would make sense to have it that way, but it's ok

pkg/nvcdi/transform/merge.go
```diff
@@ -21,13 +21,25 @@ import "tags.cncf.io/container-device-interface/specs-go"
...
+		if t == nil {
+			continue
+		}
+		nonNilTransformers = append(nonNilTransformers, t)
+	}
+	return merged(nonNilTransformers)
 }
 
 // Transform applies all the transformers in the merged set.
+// Nil transformers are skipped
```
ArangoGutierrez, line 37:
```suggestion
// Transform applies all the transformers in the merged set.
// Nil transformers are skipped.
```

ArangoGutierrez (changes requested):

ArangoGutierrez (approved):

PR #544, opened 2024-06-14
Allow toolkit.pid path to be specified

elezar:
This change makes the following changes:
* Allows the toolkit.pid path to be specified
* Creates the toolkit.pid file at /run/nvidia/toolkit/toolkit.pid by default
* Handles failures to remove the /run/nvidia/toolkit folder

### DIFF
--- a/tools/container/nvidia-toolkit/run.go
+++ b/tools/container/nvidia-toolkit/run.go
@@ -15,10 +15,10 @@ import (
 )
 
 const (
-	runDir         = "/run/nvidia"
-	pidFile        = runDir + "/toolkit.pid"
-	toolkitCommand = "toolkit"
-	toolkitSubDir  = "toolkit"
+	toolkitPidFilename = "toolkit.pid"
+	defaultPidFile     = "/run/nvidia/toolkit/" + toolkitPidFilename
+	toolkitCommand     = "toolkit"
+	toolkitSubDir      = "toolkit"
 
 	defaultToolkitArgs = ""
 	defaultRuntime     = "docker"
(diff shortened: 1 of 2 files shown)


tools/container/nvidia-toolkit/run.go
```diff
@@ -104,18 +115,26 @@ func main() {
 	log.Infof("Completed %v", c.Name)
 }
 
+func validateFlags(_ *cli.Context, o *options) error {
+	if filepath.Base(o.pidFile) != "toolkit.pid" {
```
tariq1890, line 119: nit: Let's move `toolkit.pid` to a constant var

tools/container/nvidia-toolkit/run.go
```diff
@@ -182,9 +201,16 @@ func verifyFlags(o *options) error {
 	return nil
 }
 
-func initialize() error {
+func initialize(pidFile string) error {
```
tariq1890, line 204: Is it possible to run this as an `init()` function instead? In this case, it would be guaranteed to be run only once and we can access `o.pidFile` globally instead of passing it as an argument

cdesiniotis (approved):
This looks good to me, besides the typo pointed out by @tariq1890.

tools/container/nvidia-toolkit/run.go
elezar: We can't run it as an init function because `o.pidFile` is only set when the CLI application is run and the relevant flags are parsed. It is my understanding that the `init` function is invoked before this.

Assuming that we prescribe the path, then we could do it but we would have to also include logic to close the file and release the exclusive lock.

tools/container/nvidia-toolkit/run.go
elezar: Done.

PR #548, opened 2024-06-17
Inject additional libraries for full X11 functionality

elezar:
This change updates X11 library detection to ensure that this works for a wider range of driver installations.

This includes the changes from #490


internal/discover/graphics.go
```diff
@@ -71,20 +64,178 @@ func NewGraphicsMountsDiscoverer(logger logger.Interface, driver *root.Driver, n
...
+	if len(links) == 0 {
+		return nil, nil
+	}
+
+	hooks := CreateCreateSymlinkHook(d.nvidiaCDIHookPath, links)
+
+	return hooks.Hooks()
+}
+
+// isDriverLibrary checke whether the specified filename is a specific driver library.
```
ArangoGutierrez, line 174:
```suggestion
// isDriverLibrary checks whether the specified filename is a specific driver library.
```

ArangoGutierrez (dismissed):
small typo

internal/discover/graphics.go
elezar: Thanks. Updated.

ehfd:
I think @tux-rampage might not have access to the branch.

`Signed-off-by: tux-rampage <tuxrampage@gmail.com>`

This is the required string.

elezar:
> I think @tux-rampage might not have access to the branch.
> 
> `Signed-off-by: tux-rampage <tuxrampage@gmail.com>`
> 
> This is the required string.

Yes, I would suppose that they don't have access. @tux-rampage feel free to create a PR targetting this branch indicating that you sign-off your work. Or coinfirm that I can add the string myself.

ehfd:
@elezar On a separate note, what would be the best way to detect the kernel driver version other than `/proc/driver/nvidia/version`, which is not available with WSL?

elezar:
> @elezar On a separate note, what would be the best way to detect the kernel driver version from inside the container, other than /proc/driver/nvidia/version, which is not available with WSL?

You mean the "RM_VERSION"? Here one would probably have to rely on an NVML call.

ehfd:
@elezar I did a bit more exploring, and it seems like the RM_VERSION of WSL is similar, but almost always different to the https://download.nvidia.com/XFree86/Linux-x86_64/ driver download page. Having issues with provisioning the `nvidia_drv.so` and `libglxserver_nvidia.so` X server libraries inside WSL + Container toolkit.

I'll open a new issue about this.

ehfd:
Continuing in https://github.com/NVIDIA/nvidia-container-toolkit/issues/563

ehfd:
@tux-rampage Could you possibly approve this PR and indicate that `Signed-off-by: tux-rampage <tuxrampage@gmail.com>` is good?

tux-rampage:
> @tux-rampage Could you possibly approve this PR and indicate that `Signed-off-by: tux-rampage <tuxrampage@gmail.com>` is good? We're unable to merge this.

Sure, will do this later or tomorrow morning.

Issue #511, opened 2024-05-25
Excessive runtime logging could cause Kubernetes workload deployment failure


weistonedawei:
Observed Kubernetes workload deployment failure caused by excessive logging in /run/containerd/io.containerd.runtime.v2.task/k8s.io/<container id>/log.json file.  This leads to /run tmpfs mount to be at 100% utilization, which prevents further container creation on the affected node. 

When container spec uses `exec` livenessProbe, the following log entries will be logged:
```log.json excerpt
[...]

PR #560, opened 2024-06-24
Reduce logging verbosity in NVIDIA Container Runtime

elezar:
These changes reduce the verbosity of the logging of the NVIDIA Container Runtime -- especially for the case where no modifications are required.

See:
* https://github.com/containerd/containerd/issues/8972#issuecomment-2067595892
* https://github.com/NVIDIA/nvidia-container-toolkit/issues/511

## Testing
1. Build the executable:
  ``` 
  make cmd-nvidia-container-runtime
  ```
2. Create a dummy config file:
  ```
  mkdir -p .config/nvidia-container-runtime
  cat <<EOF > .config/nvidia-container-runtime/config.toml 
  [nvidia-container-runtime]
  log-level = "info"
  runtimes = ["undefined", "missing", "runtime"]
  EOF
  ```
3. Create a dummy runtime:
  ```
  cat <<EOF > runtime
  #!/bin/bash
  echo "runtime args \$@"
  EOF
  chmod +x runtime
  ```
4. Run the NVIDIA runtime with the create command:
  ```
  PATH=$(pwd):${PATH} XDG_CONFIG_HOME=$(pwd)/.config \
  ./nvidia-container-runtime --log=$(pwd)/runtime.log create foo
  ```
5. Check the generated log:
  ```
  cat runtime.log
  ```

We can change the following:
* the log-level in the config file. Note that adding `--debug` flag to the command line (e.g. `./nvidia-container-runtime --debug`) will configure debug logging.
* The runc command to check the behaviour for non-`create` commands.

### DIFF
--- a/internal/oci/runtime_low_level.go
+++ b/internal/oci/runtime_low_level.go
@@ -31,8 +31,6 @@ func NewLowLevelRuntime(logger logger.Interface, candidates []string) (Runtime, 
 	if err != nil {
 		return nil, fmt.Errorf("error locating runtime: %v", err)
 	}
-
-	logger.Infof("Using low-level runtime %v", runtimePath)
 	return NewRuntimeForPath(logger, runtimePath)
 }
 
@@ -45,13 +43,12 @@ func findRuntime(logger logger.Interface, candidates []string) (string, error) {
 
 	locator := lookup.NewExecutableLocator(logger, "/")
 	for _, candidate := range candidates {
-		logger.Debugf("Looking for runtime binary '%v'", candidate)
+		logger.Tracef("Looking for runtime binary '%v'", candidate)
 		targets, err := locator.Locate(candidate)
 		if err == nil && len(targets) > 0 {
-			logger.Debugf("Found runtime binary '%v'", targets)
+			logger.Tracef("Found runtime binary '%v'", targets)
 			return targets[0], nil
 		}
-		logger.Debugf("Runtime binary '%v' not found: %v (targets=%v)", candidate, err, targets)
 	}
 
 	return "", fmt.Errorf("no runtime binary found from candidate list: %v", candidates)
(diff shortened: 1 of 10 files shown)


internal/oci/runtime_low_level.go
```diff
@@ -31,8 +31,6 @@ func NewLowLevelRuntime(logger logger.Interface, candidates []string) (Runtime,
 	if err != nil {
 		return nil, fmt.Errorf("error locating runtime: %v", err)
 	}
-
-	logger.Infof("Using low-level runtime %v", runtimePath)
```
ArangoGutierrez, line 35: maybe just move it to TraceF?

internal/oci/runtime_low_level.go
```diff
@@ -45,13 +43,12 @@ func findRuntime(logger logger.Interface, candidates []string) (string, error) {
...
 	for _, candidate := range candidates {
-		logger.Debugf("Looking for runtime binary '%v'", candidate)
+		logger.Tracef("Looking for runtime binary '%v'", candidate)
 		targets, err := locator.Locate(candidate)
 		if err == nil && len(targets) > 0 {
-			logger.Debugf("Found runtime binary '%v'", targets)
+			logger.Tracef("Found runtime binary '%v'", targets)
 			return targets[0], nil
 		}
-		logger.Debugf("Runtime binary '%v' not found: %v (targets=%v)", candidate, err, targets)
```
ArangoGutierrez, line 54: why removing it, maybe just move it to TraceF

ArangoGutierrez (approved):
Changes look sane, I don't have a system to test the changes but I don't see anything breaking/

Only a question at `internal/oci/runtime_low_level.go` , not a blocker

internal/oci/runtime_low_level.go
elezar: The message is redundant. We already log the candidates and then targets if they are non-emtpy

internal/oci/runtime_low_level.go
elezar: I wanted to rather control when I log this at a higher level instead of doing this for all calls to this contructor.

PR #576, opened 2024-07-02
Extract options from default runtime if runc does not exist

elezar:
This change updates the logic to populate the options for the nvidia runtime configs added to containerd or crio from a default runtime if this is specified and a runc entry is not found.

This allows the default runtime values for settings such as SystemdCgroup to be applied correctly.

### DIFF
--- a/pkg/config/engine/containerd/config_v1.go
+++ b/pkg/config/engine/containerd/config_v1.go
@@ -39,12 +39,22 @@ func (c *ConfigV1) AddRuntime(name string, path string, setAsDefault bool, confi
 
 	config.Set("version", int64(1))
 
-	if runc, ok := config.GetPath([]string{"plugins", "cri", "containerd", "runtimes", "runc"}).(*toml.Tree); ok {
-		runc, _ = toml.Load(runc.String())
-		config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name}, runc)
+	// By default we extract the runtime options from the runc settings; if this does not exist we get the options from the default runtime specified in the config.
+	runtimeNamesForConfig := []string{"runc"}
+	if name, ok := config.GetPath([]string{"plugins", "cri", "containerd", "default_runtime_name"}).(string); ok && name != "" {
+		runtimeNamesForConfig = append(runtimeNamesForConfig, name)
+	}
+	for _, r := range runtimeNamesForConfig {
+		if options, ok := config.GetPath([]string{"plugins", "cri", "containerd", "runtimes", r}).(*toml.Tree); ok {
+			c.Logger.Debugf("using options from runtime %v: %v", r, options.String())
+			options, _ = toml.Load(options.String())
+			config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name}, options)
+			break
+		}
 	}
 
 	if config.GetPath([]string{"plugins", "cri", "containerd", "runtimes", name}) == nil {
+		c.Logger.Warningf("could not infer options from runtimes %v; using defaults", runtimeNamesForConfig)
 		config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "runtime_type"}, c.RuntimeType)
 		config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "runtime_root"}, "")
 		config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "runtime_engine"}, "")
--- a/pkg/config/engine/containerd/config_v2.go
+++ b/pkg/config/engine/containerd/config_v2.go
@@ -33,12 +33,22 @@ func (c *Config) AddRuntime(name string, path string, setAsDefault bool, configO
 
 	config.Set("version", int64(2))
 
-	if runc, ok := config.GetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", "runc"}).(*toml.Tree); ok {
-		runc, _ = toml.Load(runc.String())
-		config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", name}, runc)
+	// By default we extract the runtime options from the runc settings; if this does not exist we get the options from the default runtime specified in the config.
+	runtimeNamesForConfig := []string{"runc"}
+	if name, ok := config.GetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "default_runtime_name"}).(string); ok && name != "" {
+		runtimeNamesForConfig = append(runtimeNamesForConfig, name)
+	}
+	for _, r := range runtimeNamesForConfig {
+		if options, ok := config.GetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", r}).(*toml.Tree); ok {
+			c.Logger.Debugf("using options from runtime %v: %v", r, options.String())
+			options, _ = toml.Load(options.String())
+			config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", name}, options)
+			break
+		}
 	}
 
 	if config.GetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", name}) == nil {
+		c.Logger.Warningf("could not infer options from runtimes %v; using defaults", runtimeNamesForConfig)
 		config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", name, "runtime_type"}, c.RuntimeType)
 		config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", name, "runtime_root"}, "")
 		config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", name, "runtime_engine"}, "")
--- a/pkg/config/engine/crio/crio.go
+++ b/pkg/config/engine/crio/crio.go
@@ -21,11 +21,15 @@ import (
 
 	"github.com/pelletier/go-toml"
 
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
 	"github.com/NVIDIA/nvidia-container-toolkit/pkg/config/engine"
 )
 
 // Config represents the cri-o config
-type Config toml.Tree
+type Config struct {
+	*toml.Tree
+	Logger logger.Interface
+}
 
 var _ engine.Interface = (*Config)(nil)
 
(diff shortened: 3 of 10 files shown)


pkg/config/engine/containerd/cgroup_driver.go
```diff
@@ -0,0 +1,53 @@
...
+
+var (
+	runningSystemd bool
+	detectSystemd  sync.Once
+)
+
+// getCgroupDriver returns the cgroup driver to use.
+// This is adapted from the containerd implementation.
+// https://github.com/containerd/containerd/blob/1fb1882c7de7239711c343c65fa266a2f1eba9db/internal/cri/server/runtime_config_linux.go#L63-L68
+func getCgroupDriver(root string) cgroupDriver {
```
tariq1890, line 40: Can we consider limiting the PR to just extracting the options from the defaultRuntime? 

I believe this would suffice when it comes to fixing issues like those we've seen in AKS.  Since the `SystemdCgroup = true` needs to be set explicitly, I think it's safe to assume that either i) the `runc` runtime block would have it set OR ii) the `default_runtime` runtime block , given any containerd config toml

pkg/config/engine/containerd/config_v1.go
```diff
@@ -39,9 +39,17 @@ func (c *ConfigV1) AddRuntime(name string, path string, setAsDefault bool, confi
...
 
-	if runc, ok := config.GetPath([]string{"plugins", "cri", "containerd", "runtimes", "runc"}).(*toml.Tree); ok {
-		runc, _ = toml.Load(runc.String())
-		config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name}, runc)
+	// By default we extract the runtime options from the runc settings; if this does not exist we get the options from the default runtime specified in the config.
+	runtimeNamesForConfig := []string{"runc"}
+	if name, ok := config.GetPath([]string{"plugins", "cri", "containerd", "default_runtime_name"}).(string); ok && name != "" {
+		runtimeNamesForConfig = append(runtimeNamesForConfig, name)
+	}
+	for _, runtimeNameForOptions := range runtimeNamesForConfig {
```
tariq1890, line 47: As per Go conventions, It's preferable to use single letters for iterator variables. 
```suggestion
	for _, r := range runtimeNamesForConfig {
```

pkg/config/engine/containerd/cgroup_driver.go
```diff
@@ -0,0 +1,53 @@
...
+		if err != nil {
+			return
+		}
+		runningSystemd = fi.IsDir()
+	})
+
+	if !runningSystemd {
+		return cgroupDriverCgroupFS
+	}
+	return cgroupDriverSystemd
```
klueska, line 52: Just because sytemd is running doesn't mean that the systemd cgroup driver is in use by containerd. In older versions of containerd `cgroupfs` was the default and would be used even if systemd was in use by the machine in general. Are we OK, saying that we only support newer versions of containerd where the default cgroup driver is implied by the presence of systemd?

pkg/config/engine/containerd/config_v1.go
```diff
@@ -82,6 +90,14 @@ func (c *ConfigV1) AddRuntime(name string, path string, setAsDefault bool, confi
 		}
 	}
 
+	// If the systemd cgroup driver is required, and no other setting is applied, we set SystemdCgroup.
+	if c.CgroupDriver == cgroupDriverSystemd {
+		valueSet := config.GetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "options", "SystemdCgroup"})
+		if valueSet == nil {
+			config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "options", "SystemdCgroup"}, true)
+		}
+	}
```
klueska, line 99: Can we instead specifically look to see if `SystemdCgroup` is set, and if not, then set it appropriately (i.e. either true or false) depending on the value of `c.CgroupDriver`.

pkg/config/engine/containerd/config_v2.go
```diff
@@ -60,6 +68,14 @@ func (c *Config) AddRuntime(name string, path string, setAsDefault bool, configO
 		config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "default_runtime_name"}, name)
 	}
 
+	// If the systemd cgroup driver is required, and no other setting is applied, we set SystemdCgroup.
+	if c.CgroupDriver == cgroupDriverSystemd {
+		valueSet := config.GetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", name, "options", "SystemdCgroup"})
+		if valueSet == nil {
+			config.SetPath([]string{"plugins", "io.containerd.grpc.v1.cri", "containerd", "runtimes", name, "options", "SystemdCgroup"}, true)
+		}
+	}
```
klueska, line 77: Same comment as above

pkg/config/engine/containerd/config_v1.go
```diff
@@ -39,9 +39,17 @@ func (c *ConfigV1) AddRuntime(name string, path string, setAsDefault bool, confi
 
 	config.Set("version", int64(1))
 
-	if runc, ok := config.GetPath([]string{"plugins", "cri", "containerd", "runtimes", "runc"}).(*toml.Tree); ok {
-		runc, _ = toml.Load(runc.String())
-		config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name}, runc)
+	// By default we extract the runtime options from the runc settings; if this does not exist we get the options from the default runtime specified in the config.
```
klueska, line 42: Just out of curiosity, why `runc` and then `default_runtime`?  Does it make sense to do it in the other order?

pkg/config/engine/containerd/config_v1.go
klueska: So like this:
```
valueSet := config.GetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "options", "SystemdCgroup"})
if valueSet == nil && c.CgroupDriver == cgroupDriverSystemd {
	config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "options", "SystemdCgroup"}, true)
}
if valueSet == nil && c.CgroupDriver == cgroupDriverCgroupFS {
	config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "options", "SystemdCgroup"}, false)
}

File: pkg/config/engine/containerd/cgroup_driver.go (hunk 1 above)
Comment (klueska) [reply pkg/config/engine/containerd/cgroup_driver.go:40]: The problem is that we don't have access to the default config.toml. If we did, we wouldn't need this logic at all and would just pull it from the config file.

File: pkg/config/engine/containerd/cgroup_driver.go [hunk 7]
```diff
@@ -0,0 +1,53 @@
...
+	detectSystemd  sync.Once
+)
+
+// getCgroupDriver returns the cgroup driver to use.
+// This is adapted from the containerd implementation.
+// https://github.com/containerd/containerd/blob/1fb1882c7de7239711c343c65fa266a2f1eba9db/internal/cri/server/runtime_config_linux.go#L63-L68
+func getCgroupDriver(root string) cgroupDriver {
+	detectSystemd.Do(func() {
+		fi, err := os.Lstat(filepath.Join(root, "/run/systemd/system"))
+		if err != nil {
```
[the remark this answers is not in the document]
klueska: We should ignore the error if it is "file does not exist" but log any others.

pkg/config/engine/containerd/config_v1.go
elezar: The main motivation is to not change the existing behaviour by default.

pkg/config/engine/containerd/cgroup_driver.go
elezar: Given that we only consume this setting if no config entry exists from which to infer it, does it make sense to change this function to `hasSystemd` or something similar? We would then only use it to set `SystemdCgroup` if not already specified.

In terms of only supporting newer containerd versions, I'm not sure what the right approach there should be. Since 1.6 is an LTS release, we should at least support this. Maybe this speaks to @tariq1890's point that we should not implement the automatic detection here and consider the implications more closely.

pkg/config/engine/containerd/config_v1.go
elezar: Sure.

pkg/config/engine/containerd/cgroup_driver.go
klueska: Yeah, that sounds fine to me.

Let's call it `hasSystemd`, and then do what I suggest below about *explicitly* setting `SytemdCgroup` to `true` or `false` based on the results of it.

It still implies the existence of a "newish" containerd, but I think I'm OK with this.

Where the new semantics are -- "use whatever setting is in the config file, and if there is no setting, decide how to set it based on the existence of systemd on the machine".

This *may* break for older containerds, but there is a simple workaround (i.e. add an explicit setting in the config file) and over time we will converge on this as the "right fix".

pkg/config/engine/containerd/config_v1.go
klueska: OK, that's fine then.

pkg/config/engine/containerd/cgroup_driver.go
elezar: I have moved the systemd-specifics to #578 to keep this PR more focussed.

pkg/config/engine/containerd/config_v1.go
elezar: I have moved the systemd-specifics to #578 to keep this PR more focussed.

pkg/config/engine/containerd/config_v2.go
elezar: I have moved the systemd-specifics to #578 to keep this PR more focussed.

tariq1890:
Can we add unit tests to cover this new change?

pkg/config/engine/containerd/cgroup_driver.go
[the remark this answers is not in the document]
elezar: Logging was added to #578

pkg/config/engine/crio/crio.go
```diff
@@ -47,6 +47,19 @@ func (c *Config) AddRuntime(name string, path string, setAsDefault bool, _ ...ma
 
 	config := (toml.Tree)(*c)
 
+	// By default we extract the runtime options from the runc settings; if this does not exist we get the options from the default runtime specified in the config.
+	runtimeNamesForConfig := []string{"runc"}
+	if name, ok := config.GetPath([]string{"crio", "runtime", "default_runtime"}).(string); ok && name != "" {
+		runtimeNamesForConfig = append(runtimeNamesForConfig, name)
+	}
+	for _, runtimeNameForOptions := range runtimeNamesForConfig {
```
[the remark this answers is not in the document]
elezar: Done.

elezar:
> Can we add unit tests to cover this new change?

Good call. Added some basic tests for the resolution logic.

tariq1890 (approved):

PR #578, opened 2024-07-03
Detect the presence of systemd and use this to set SystemdCgroup

elezar:
This change checks whether systemd is running and uses this to set
the containerd runtime options SystemdCgroup if not overridden in
the config explicitly.

This builds on #576

### DIFF
--- a/pkg/config/engine/containerd/systemd.go
+++ b/pkg/config/engine/containerd/systemd.go
@@ -1,0 +1,43 @@
+/**
+# Copyright 2024 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


pkg/config/engine/containerd/systemd.go
```diff
@@ -0,0 +1,43 @@
...
+
+var (
+	runningSystemd bool
+	detectSystemd  sync.Once
+)
+
+// hasSystemd returns whether systemd is running on this system.
+// This is adapted from the containerd implementation.
+// https://github.com/containerd/containerd/blob/1fb1882c7de7239711c343c65fa266a2f1eba9db/internal/cri/server/runtime_config_linux.go#L63-L68
+func (b *builder) hasSystemd() bool {
```
tariq1890, line 33: Is it possible that a node has a running systemd , but cgroup driver configured for the kubelet is `cgroupfs` ?

elezar: I would say that it's possible but unlikely with newer containerd / kubelet versions.

As of k8s 1.22 the default value was `systemd` (see https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/#configuring-the-kubelet-cgroup-driver).

It is also recommended to use the `systemd` driver with `cgroupv2` when using containerd at least: https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd-systemd

elezar:
I am closing this in favour of extracting the correct config from the container runtimes as in #641

PR #601, opened 2024-07-10
[no-relnote] Fix release tooling

elezar:


hack/prepare-artifacts.sh
```diff
@@ -47,7 +47,44 @@ PACKAGE_ROOT=release-${VERSION}-${REPO}
...
+  elif is_command openssl; then
+    hash=$(openssl -dst openssl dgst -sha256 "$TARGET") || return 1
+    echo "$hash" | cut -d ' ' -f a
+  else
+    log_err "hash_sha256 unable to find command to compute sha-256 hash"
+    return 1
+  fi
+)
+
+
```
ArangoGutierrez, line 79: nit trailing new lines

hack/create-release.sh
```diff
@@ -31,7 +31,7 @@ REPOSITORY=NVIDIA/nvidia-container-toolkit
 echo "Creating draft release"
 ./hack/generate-changelog.sh --version ${VERSION} | \
     grep -v "### Version v" | \
```
ArangoGutierrez, line 33: generate-changelog doesn't have a string `### Version v`,  this match won't happen

elezar: You're right. It's an inverted match anyway, so we'll skip something that's not there.

ArangoGutierrez (approved):
tested all the bash bits, can't test the `gh` bits since it doesn't provide a dry run option. But PR looks good

PR #602, opened 2024-07-12
TOFIX: packages in image

elezar:


cmd/nvidia-ctk-installer/main.go
```diff
@@ -123,11 +125,17 @@ func (a app) build() *cli.App {
 			EnvVars:     []string{"TOOLKIT_INSTALL_DIR", "ROOT"},
 		},
 		&cli.StringFlag{
-			Name:        "source-root",
-			Value:       "/",
-			Usage:       "The folder where the required toolkit artifacts can be found",
+			Name:        "toolkit-source-root",
```
ArangoGutierrez, line 128: Do we need to update documentation on this flag name change?

cmd/nvidia-ctk-installer/toolkit/installer/package-type.go
```diff
@@ -0,0 +1,19 @@
+/*
```
ArangoGutierrez, line 1: What's the purpose of this file?

cmd/nvidia-ctk-installer/main.go
```diff
@@ -277,3 +294,35 @@ func (a *app) shutdown(pidFile string) {
...
+		return "/artifacts/deb", nil
+	case "rpm":
+		return "/artifacts/rpm", nil
+	default:
+		return "", fmt.Errorf("invalid package type: %v", resolvedPackageType)
+	}
+}
+
+func (a *app) resolvePackageType(hostRoot string, packageType string) (rPackageTypes string, rerr error) {
+	if packageType != "" && packageType != "auto" {
```
ArangoGutierrez, line 314:
```suggestion
	if packageType != ""  || packageType != "auto" {
```

cmd/nvidia-ctk-installer/main.go
```diff
@@ -277,3 +294,35 @@ func (a *app) shutdown(pidFile string) {
...
+		return packageType, nil
+	}
+
+	locator := lookup.NewExecutableLocator(a.logger, hostRoot)
+	if candidates, err := locator.Locate("/usr/bin/rpm"); err == nil && len(candidates) > 0 {
+		return "rpm", nil
```
ArangoGutierrez, line 320: Should we have `rpm` and `deb` as pkg consts so is safer to use the var instead of a string.

ArangoGutierrez (commented):
Just some comments, will do a second review once PR becomes ready for review

cmd/nvidia-ctk-installer/main.go
elezar: This is a new flag and is not publicly accessible. It is internal only.

cmd/nvidia-ctk-installer/main.go
elezar: I want to return the value that `packageType` is set to anything that is not `""` or `"auto"`. When using `||` `packageType == "auto"` will return auto.

cmd/nvidia-ctk-installer/toolkit/installer/package-type.go
elezar: It should be removed.

deployments/container/Dockerfile.ubi8
```diff
@@ -47,33 +47,92 @@ ARG VERSION="N/A"
...
+# - The extracted deb packages
+# - The extracted rpm packages
+# - The nvidia-ctk-installer binary
+FROM nvcr.io/nvidia/cuda:12.8.1-base-ubi8 AS artifacts
+
+COPY --from=rpmpackages /artifacts/rpm /artifacts/rpm
+COPY --from=debpackages /artifacts/deb /artifacts/deb
+COPY --from=build /artifacts/bin /artifacts/build
+
+FROM nvcr.io/nvidia/cuda:12.8.1-base-ubi8
```
elezar, line 125: @tariq1890 this could be distroless if required.

tariq1890: Ack

deployments/container/Dockerfile
```diff
@@ -0,0 +1,148 @@
+# Copyright (c) 2019-2021, NVIDIA CORPORATION.  All rights reserved.
```
tariq1890: update the year here

deployments/container/Dockerfile
```diff
@@ -0,0 +1,148 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG GOLANG_VERSION=x.x.x
+ARG VERSION="N/A"
+
+FROM nvcr.io/nvidia/cuda:12.9.0-base-ubi8 AS build
+
+RUN yum install -y \
```
tariq1890: nit: `yum` has long since been an alias for `dnf`. Let's start using dnf as it is more accurate
```suggestion
RUN dnf install -y \
```

deployments/container/Dockerfile
```diff
@@ -0,0 +1,148 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG GOLANG_VERSION=x.x.x
+ARG VERSION="N/A"
+
+FROM nvcr.io/nvidia/cuda:12.9.0-base-ubi8 AS build
```
tariq1890, line 19: Can we start using `ubi9` instead?

deployments/container/Dockerfile
```diff
@@ -0,0 +1,148 @@
...
+
+# Create a manifest.txt file with the absolute paths of all deb and rpm packages in the container
+RUN echo "#IMAGE_EPOCH=$(date '+%s')" > /artifacts/manifest.txt && \
+    env | sed 's/^/#/g' >> /artifacts/manifest.txt && \
+    find /artifacts/packages -iname '*.deb' -o -iname '*.rpm' >> /artifacts/manifest.txt
+
+RUN mkdir /licenses && mv /NGC-DL-CONTAINER-LICENSE /licenses/NGC-DL-CONTAINER-LICENSE
+
+# The debpackages stage is used to extract the contents of deb packages.
+FROM nvcr.io/nvidia/cuda:12.9.0-base-ubuntu20.04 AS debpackages
```
tariq1890, line 76: Ubuntu 20.04 has been EOL'ed as of May 31, 2025
```suggestion
FROM nvcr.io/nvidia/cuda:12.9.0-base-ubuntu22.04 AS debpackages
```

deployments/container/Dockerfile
```diff
@@ -0,0 +1,148 @@
...
+
+ARG TARGETARCH
+ARG PACKAGE_DIST_DEB=ubuntu18.04
```
tariq1890, line 79: Should we continue using `ubuntu18.04` given ubuntu 20.04 has reached end-of-life?

deployments/container/Dockerfile
elezar: We only track the year where the file was created. I've updated to 2019 for consistency.

deployments/container/Dockerfile
elezar: We can, but I would prefer to use that as a follow-up. Note that we don't need to use the CUDA images for our intermediate images and we could consider something else entirely.

deployments/container/Dockerfile
elezar: Sure. This was moved from the original ubi8 image. Will update it here though.

deployments/container/Dockerfile
elezar: We are still building our packages using `ubuntu18.04` and `centos7` to ensure the broadest compatibility (glibc version).

elezar: Even if we change, this is out of scope for this PR.

deployments/container/Dockerfile
elezar: That image isn't available.

PR #629, opened 2024-08-05
Add changes for usage of nvsandboxutils

sananya12:
This request adds all the changes related to the usage of the new library _libnvidia-sandboxutils_ (Sandboxutils) to obtain the file system information in CDI spec generation.


internal/nvsandboxutils/gpu-resources.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+//	type GpuFileInfo_v1 struct {
+//		Next        *GpuFileInfo_v1
+//		FileType    uint32
+//		FileSubType uint32
+//		Module      uint32
+//		Flags       uint32
+//		FilePath    *int8
+//	}
```
elezar, line 36: Is this commented block required?

internal/nvsandboxutils/gpu-resources.go
```diff
@@ -0,0 +1,74 @@
...
+type GpuFileInfo struct {
+	Path    string
+	Type    FileType
+	SubType FileSystemSubType
+	Module  FileModule
+	Flags   FileFlag
+}
+
+func (l *library) GetGpuResource(uuid string) ([]GpuFileInfo, Ret) {
```
elezar, line 45: One thing that we have considered doing in `go-nvml` as well (see https://github.com/NVIDIA/go-nvml/pull/125) is returning an `error` instead of a `Ret` value. This cleans up most standard interactions with the library in the calling code.

I'm happy to take this on as a follow-up though if we don't want to do it as part of the initial commits.

pkg/nvcdi/lib.go
```diff
@@ -213,7 +228,7 @@ func (l *nvcdilib) resolveMode() (rmode string) {
 }
 
 // getCudaVersion returns the CUDA version of the current system.
-func (l *nvcdilib) getCudaVersion() (string, error) {
+func (l *nvcdilib) getCudaVersionNvml() (string, error) {
```
elezar, line 231: nit: for this diff, let's keep the `getCudaVersion()` implementation here and then have the two implementations below this (for golang there is no need to have a function declared before its use).

internal/platform-support/dgpu/dgpu.go
```diff
@@ -25,22 +27,79 @@ import (
 )
 
 // NewForDevice creates a discoverer for the specified Device.
+// nvsandboxutils is used for discovery if specified, otherwise NVML is used.
 func NewForDevice(d device.Device, opts ...Option) (discover.Discover, error) {
 	o := new(opts...)
+	o.isMigDevice = false
```
elezar, line 33: nit :`isMigDevice` should already be initialized to `false` and this is unneeded.

elezar (commented):
Thanks @sananya12.

I have some minor comments. I don't think any of them are blockers for getting this merged though.

internal/nvsandboxutils/gpu-resources.go
sananya12: No, I think this can be removed now. Thanks!

pkg/nvcdi/lib.go
sananya12: Didn't know that, thanks. Moved the two implementations below.

internal/platform-support/dgpu/dgpu.go
sananya12: Removed this, thanks.

internal/nvsandboxutils/gpu-resources.go
sananya12: I see, it does look cleaner. I can create a follow-up patch for those changes once the initial commits for the library are in.

elezar (approved):

PR #630, opened 2024-08-05
Generate CDI specification including additional GIDs

elezar:
This change adds logic to include the `AdditionalGIDs` required for a device node in a generated CDI specification.

When a device node requires specific group membership for access, a non-root user running in a container (e.g. using the `-u` Docker command line argument) may not have access to the device.

in the v0.7.0 CDI specification, support was added for specifying additional GIDs in the CDI specification and these changes extract the required information from the device nodes on the host if available.

### DIFF
--- a/internal/edits/device.go
+++ b/internal/edits/device.go
@@ -17,6 +17,9 @@
 package edits
 
 import (
+	"os"
+
+	"golang.org/x/sys/unix"
 	"tags.cncf.io/container-device-interface/pkg/cdi"
 	"tags.cncf.io/container-device-interface/specs-go"
 
@@ -32,9 +35,15 @@ func (d device) toEdits() (*cdi.ContainerEdits, error) {
 		return nil, err
 	}
 
+	var additionalGIDs []uint32
+	if requiredGID, _ := d.getRequiredGID(); requiredGID != 0 {
+		additionalGIDs = append(additionalGIDs, requiredGID)
+	}
+
 	e := cdi.ContainerEdits{
 		ContainerEdits: &specs.ContainerEdits{
-			DeviceNodes: []*specs.DeviceNode{deviceNode},
+			DeviceNodes:    []*specs.DeviceNode{deviceNode},
+			AdditionalGIDs: additionalGIDs,
 		},
 	}
 	return &e, nil
@@ -52,10 +61,37 @@ func (d device) toSpec() (*specs.DeviceNode, error) {
 	if hostPath == d.Path {
 		hostPath = ""
 	}
+
 	s := specs.DeviceNode{
 		HostPath: hostPath,
 		Path:     d.Path,
 	}
 
 	return &s, nil
+}
+
+// getRequiredGID returns the group id of the device if the device is not world read/writable
+func (d device) getRequiredGID() (uint32, error) {
+	path := d.HostPath
+	if path == "" {
+		path = d.Path
+	}
+	if path == "" {
+		return 0, nil
+	}
+
+	var stat unix.Stat_t
+	if err := unix.Lstat(path, &stat); err != nil {
+		return 0, err
+	}
+	// This is only supported for char devices
+	if stat.Mode&unix.S_IFMT != unix.S_IFCHR {
+		return 0, nil
+	}
+
+	permissions := os.FileMode(stat.Mode).Perm()
+	if permissions&06 == 0 {
+		return stat.Gid, nil
+	}
+	return 0, nil
 }
--- a/internal/edits/resource.go
+++ b/internal/edits/resource.go
@@ -1,0 +1,30 @@
+/**
+# Copyright 2023 NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package edits
+
+import (
+	"tags.cncf.io/container-device-interface/pkg/cdi"
+	"tags.cncf.io/container-device-interface/specs-go"
+)
+
+// NewResource creates a CDI resource (Device) with the specified name.
+func NewResource(name string, edits *cdi.ContainerEdits) specs.Device {
+	return specs.Device{
+		Name:           name,
+		ContainerEdits: *edits.ContainerEdits,
+	}
+}
--- a/pkg/nvcdi/full-gpu-nvml.go
+++ b/pkg/nvcdi/full-gpu-nvml.go
@@ -30,7 +30,7 @@ import (
 
 // GetGPUDeviceSpecs returns the CDI device specs for the full GPU represented by 'device'.
 func (l *nvmllib) GetGPUDeviceSpecs(i int, d device.Device) ([]specs.Device, error) {
-	edits, err := l.GetGPUDeviceEdits(d)
+	e, err := l.GetGPUDeviceEdits(d)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get edits for device: %v", err)
 	}
@@ -41,10 +41,7 @@ func (l *nvmllib) GetGPUDeviceSpecs(i int, d device.Device) ([]specs.Device, err
 		return nil, fmt.Errorf("failed to get device name: %v", err)
 	}
 	for _, name := range names {
-		spec := specs.Device{
-			Name:           name,
-			ContainerEdits: *edits.ContainerEdits,
-		}
+		spec := edits.NewResource(name, e)
 		deviceSpecs = append(deviceSpecs, spec)
 	}
 
--- a/pkg/nvcdi/gds.go
+++ b/pkg/nvcdi/gds.go
@@ -38,15 +38,12 @@ func (l *gdslib) GetAllDeviceSpecs() ([]specs.Device, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to create GPUDirect Storage discoverer: %v", err)
 	}
-	edits, err := edits.FromDiscoverer(discoverer)
+	e, err := edits.FromDiscoverer(discoverer)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create container edits for GPUDirect Storage: %v", err)
 	}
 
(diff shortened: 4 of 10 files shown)


elezar:
Note: This updates the default spec version to v0.7.0 since the DRM devices generally require additional GIDs. As such, this should probably be an opt-in (or at least an opt-out) feature.

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -169,6 +174,11 @@ func (m command) build() *cli.Command {
 			Usage:       "Specify a pattern the CSV mount specifications.",
 			Destination: &opts.csv.ignorePatterns,
 		},
+		&cli.BoolFlag{
+			Name:        "--allow-additional-gids",
+			Usage:       "Allow the use of the additionalGIDs field for generated CDI specifications. Note this will generate a v0.7.0 CDI specification.",
```
klueska, line 179: Which versions of containerd / cri-o (and other runtimes for that matter) support this?

internal/config/features.go
```diff
@@ -19,18 +19,20 @@ package config
...
 type features struct {
-	GDS      *feature `toml:"gds,omitempty"`
-	MOFED    *feature `toml:"mofed,omitempty"`
-	NVSWITCH *feature `toml:"nvswitch,omitempty"`
-	GDRCopy  *feature `toml:"gdrcopy,omitempty"`
+	GDS              *feature `toml:"gds,omitempty"`
+	MOFED            *feature `toml:"mofed,omitempty"`
+	NVSWITCH         *feature `toml:"nvswitch,omitempty"`
+	GDRCopy          *feature `toml:"gdrcopy,omitempty"`
+	NoAdditionalGIDs *feature `toml:"no-additional-gids,omitempty"`
```
klueska, line 35: Why is this reversed (i.e. `no` rather than `allow`) from the flag passed in?

cmd/nvidia-ctk/cdi/generate/generate.go
elezar: I would have to check. Note that this is not enabled by default when generating the CDI spec and only for internal representations.

internal/config/features.go
elezar: Because this has different default behaviour. For the NVIDIA Contianer Runtime I want to generated an internal representation including the `additionalGIDs` container edits for device nodes. This allows users to opt out if there is an issue.

I suppose we could make the argument that if we backport this (which we may want to due to the Jetpack feature request) we should keep the existing behaviour and not make these modifications either.

cmd/nvidia-ctk/cdi/generate/generate.go
elezar: * containerd (unreleased) - https://github.com/containerd/containerd/commit/7a2f49f70f1d2cacfededbbf65452d481476bc10
* docker (>= 26.1.0) - https://github.com/moby/moby/commit/745e2356ab01363a02cfec947c624a9f6220fe7f
* podman (>= 5.1.0) - https://github.com/containers/podman/commit/a40cf3195acb6ac5fea5ab4617afb99006a3bed7
* crio (>= 1.30.0) - https://github.com/cri-o/cri-o/commit/fd9aa76250fe05625d8c968b922cd1a0ae88eb1b

internal/edits/lib.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	mounts, err := d.Mounts()
+	if err != nil {
+		return nil, fmt.Errorf("failed to discover mounts: %v", err)
+	}
+
+	hooks, err := d.Hooks()
+	if err != nil {
+		return nil, fmt.Errorf("failed to discover hooks: %v", err)
```
tariq1890, line 56:
```suggestion
		return nil, fmt.Errorf("failed to discover hooks: %w", err)
```

internal/edits/lib.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	c := NewContainerEdits()
+	for _, d := range devices {
+		edits, err := device(d).toEdits(o.allowAdditionalGIDs)
+		if err != nil {
+			return nil, fmt.Errorf("failed to created container edits for device: %v", err)
```
tariq1890, line 63:
```suggestion
			return nil, fmt.Errorf("failed to created container edits for device: %w", err)
```

internal/edits/lib.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return c, nil
+}
+
+// SpecModifierFromDiscoverer creates a SpecModifier that defines the required OCI spec edits (as CDI ContainerEdits) from the specified
+// discoverer.
+func (o *options) SpecModifierFromDiscoverer(d discover.Discover) (oci.SpecModifier, error) {
+	c, err := o.EditsFromDiscoverer(d)
+	if err != nil {
+		return nil, fmt.Errorf("error constructing container edits: %v", err)
```
tariq1890, line 84:
```suggestion
		return nil, fmt.Errorf("error constructing container edits: %w", err)
```

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -221,7 +231,7 @@ func (m command) validateFlags(c *cli.Context, opts *options) error {
 	return nil
 }
 
-func (m command) run(c *cli.Context, opts *options) error {
+func (m command) run(_ *cli.Context, opts *options) error {
```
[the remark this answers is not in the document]
elezar: I suppose we COULD use it, but we don't really make use of non-`cli.Contexts` in this particular codebase.

Since this change is actually unrelated to anything else we're doing here, I'll rather roll it back. I'm happy to revisit in future.

internal/edits/device.go
```diff
@@ -26,15 +29,23 @@ import (
...
-func (d device) toEdits() (*cdi.ContainerEdits, error) {
+func (d device) toEdits(allowAdditionalGIDs bool) (*cdi.ContainerEdits, error) {
 	deviceNode, err := d.toSpec()
 	if err != nil {
 		return nil, err
 	}
 
+	var additionalGIDs []uint32
+	if allowAdditionalGIDs {
+		if requiredGID, _ := d.getRequiredGID(); requiredGID != 0 {
```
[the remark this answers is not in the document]
elezar: Updated to not return an error.

internal/config/features.go
elezar: Ok, after having thought this through again, I have disabled this by default and switched the feature flag to `--allow-additional-gids`.

elezar: Resolving this outdated conversation. Note that for the `1.19.0` release we will have this ON by default. This can be disabled using a feature flag or disabled in CDI spec generation by specifying a feature flag.

Happy to hear discussions on this though.

Will update the description with more information on the change.

cmd/nvidia-ctk/cdi/generate/generate.go
elezar: Added this to the PR description.

internal/edits/edits.go
```diff
@@ -33,14 +34,34 @@ const (
 
 type Factory interface {
 	New() *cdi.ContainerEdits
+	FromDiscoverer(discover.Discover) (*cdi.ContainerEdits, error)
 }
 
 type empty string
 
+type factory struct {
```
cdesiniotis, line 42: nit: let's add a compile-time check to ensure `factory` implements the `Factory` interface
```
var _ Factory = (*factory)(nil)
```

internal/edits/device.go
```diff
@@ -104,3 +107,25 @@ func ptrIfNonZero(id uint32) *uint32 {
...
+	if dn.FileMode == nil {
+		return nil
+	}
+	if dn.FileMode.Type() != os.ModeCharDevice {
+		return nil
+	}
+
+	if permissionsForOther := dn.FileMode.Perm(); permissionsForOther&06 != 0 {
+		return []uint32{*dn.GID}
+	}
```
cdesiniotis: Not sure I follow this. The device is world readable or writeable if `(dn.FileMode.Perm() & 06) != 0`. Don't we want to add the group id for the opposite condition when the device is NOT world readable AND world writeable?

In my mind, this should be re-written as:
```
if rwPermissionsForOther := (dn.FileMode.Perm() & 06); rwPermissionsForOther != 06 {
  return []uint32{*dn.GID}
}
```
but let me know if I am misunderstanding.

api/config/v1/features.go
```diff
@@ -47,6 +47,9 @@ type features struct {
 	// possibly bypassing other checks by an orchestration system such as
 	// kubernetes.
 	IgnoreImexChannelRequests *feature `toml:"ignore-imex-channel-requests,omitempty"`
+	// NoAdditionalGIDsForDeviceNodes disables the injection of additional GIDs
+	// for a device node when the node is not readable by the user.
```
cdesiniotis: Not critical, but should we mention both read and write permissions?
`s/not readable by the user/not readable and writeable by the user/`

internal/modifier/cdi.go
```diff
@@ -177,6 +177,11 @@ func newAutomaticCDISpecModifier(logger logger.Interface, cfg *config.Config, de
 
 	cdiModeIdentifiers := cdiModeIdentfiersFromDevices(devices...)
 
+	nvcdiFeatureFlags := cfg.NVIDIAContainerRuntimeConfig.Modes.JitCDI.NVCDIFeatureFlags
+	if cfg.Features.NoAdditionalGIDsForDeviceNodes.IsEnabled() {
+		nvcdiFeatureFlags = append(nvcdiFeatureFlags, nvcdi.FeatureDisableMultipleCSVDevices)
```
cdesiniotis: Could you explain why this feature flag is also being set?

pkg/nvcdi/api.go
```diff
@@ -92,4 +92,8 @@ const (
 	// FeatureDisableMultipleCSVDevices disables the handling of multiple devices
 	// in CSV mode.
 	FeatureDisableMultipleCSVDevices = FeatureFlag("disable-multiple-csv-devices")
+
+	// NoAdditionalGIDsForDeviceNodes disables the injection of additional GIDs
+	// for a device node when the node is not readable by the user.
```
cdesiniotis: Same comment as before. Should we mention both read and write permissions?
`s/not readable by the user/not readable and writeable by the user/`

internal/edits/device.go
elezar: Yes. You're correct. The logic is not complete. Thanks for your suggestion. Let me also add some unit tests.

*Update*: After adding the unit test, I decided to also introduce two helpers (`isWorldReadable, `isWorldWritable` for readability).

internal/modifier/cdi.go
elezar: It shouldn't be. This was an auto-complete error on my side. Fixed.

PR #641, opened 2024-08-08
fetch current container runtime config through the command line

tariq1890:
In this PR, we change how the baseline container runtime configuration is fetched when using containerd/crio.

We now know that neither `/etc/containerd/config.toml` nor `/etc/crio/crio.conf` files provide a complete picture of the current configuration. To address this gap, we run the `containerd config dump` or `crio status config` commands to get the complete runtime configuration that reflects the current state.

This PR has been tested in the following scenarios

i) `nvidia-ctk runtime configure --runtime=containerd ` - Standalone 
ii) nvidia-container-toolkit daemonset - GPU Operator Stack

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,6 +18,10 @@ linters:
 linters-settings:
   goimports:
     local-prefixes: github.com/NVIDIA/nvidia-container-toolkit
+  gosec:
+    excludes:
+    # TODO: Consider hardening security of command line invocations
+    - G204
 
 issues:
   exclude:
--- a/pkg/config/engine/containerd/option.go
+++ b/pkg/config/engine/containerd/option.go
@@ -18,7 +18,7 @@ package containerd
 
 import (
 	"fmt"
-	"os"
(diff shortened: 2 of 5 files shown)


elezar:
Thanks for starting on this @tariq1890. I was thinking a bit on how to implement this and created https://github.com/NVIDIA/nvidia-container-toolkit/pull/643 (more from a refactoring perspective).

The idea is that we introduce an interface that returns a `*toml.Tree` when called. We then have different implementations of this for:
1. a `.toml` file.
2. a command such as `containerd config dump`
3. a direct query to the cri socket (not yet implemented).

This has the benefit that we can separate the construction of this `Source` / `Reference` from its use in the various runtimes.

What are your thoughts?

tariq1890:
Thanks @elezar . I have opened a new PR with your refactoring changes: https://github.com/NVIDIA/nvidia-container-toolkit/pull/686

elezar:
Does #686 replace this? If so, can we close this PR?

PR #643, opened 2024-08-09
WIP: Add TomlSource as abstract concept

elezar:


pkg/config/source.go
```diff
@@ -0,0 +1,115 @@
...
+	cmd.Stdout = &outb
+	cmd.Stderr = &errb
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("failed to run command %v %v: %w", c.command, c.args, err)
+	}
+
+	return LoadBytes(outb.Bytes())
+}
+
+type tomlFromCRI struct {
```
tariq1890, line 105: Given that `crictl info` doesn't provide the full config for CRI-O, this is probably not a useful toml source candidate

tariq1890: We should probably just stick to the file-based and CLI-based sources for now IMO

pkg/config/raw.go
```diff
@@ -14,19 +14,19 @@
 # limitations under the License.
 **/
 
-package engine
+package config
```
tariq1890, line 17: Why has this been moved from `engine` to `config`

tariq1890 (approved):

pkg/config/engine/api.go
```diff
@@ -19,7 +19,7 @@ package engine
 // Interface defines the API for a runtime config updater.
 type Interface interface {
 	DefaultRuntime() string
-	AddRuntime(string, string, bool, ...map[string]interface{}) error
+	AddRuntime(string, string, bool) error
```
tariq1890, line 22: Question: why remove this?

pkg/config/engine/crio/option.go
```diff
@@ -17,17 +17,14 @@
...
-
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+	"github.com/NVIDIA/nvidia-container-toolkit/pkg/config/toml"
 )
 
 type builder struct {
-	logger logger.Interface
-	path   string
+	logger    logger.Interface
+	reference toml.Loader
```
tariq1890, line 26: I feel like the name `reference` may be a bit too vague. How about `configRef`?

pkg/config/toml/source-file.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Load loads the contents of the specified TOML file as a map.
+func (f tomlFile) Load() (*Tree, error) {
+	info, err := os.Stat(string(f))
+	if os.IsExist(err) && info.IsDir() {
+		return nil, fmt.Errorf("config file %s is a directory", string(f))
+	}
+
+	if os.IsNotExist(err) {
+		return Empty.Load()
```
tariq1890, line 36: I think we should fail fast here and error out indicating that the file wasn't found

tools/container/crio/crio.go
```diff
@@ -222,6 +223,7 @@ func setupConfig(o *options) error {
 
 	cfg, err := crio.New(
 		crio.WithPath(o.Config),
+		crio.WithReference(toml.FromFile(o.Config)),
```
tariq1890, line 226: I think `WithConfigReference` may be more suggestive of the struct field's role and purpose

pkg/config/engine/api.go
elezar: This was added to allow users to override the `SystemdCgroup` option but we decided not to actually follow this as a solution. Including this in the API overcomplicates the implementation and I would rather remove it as unused instead of keeping the additional complexity.

pkg/config/engine/crio/option.go
elezar: Sure.

pkg/config/raw.go
elezar: Because this represents a `Raw` config. Importing `enging.Raw` does not make the intent as clear as `config.Raw` when imported.

pkg/config/toml/source-file.go
elezar: That is not how the current code works. We load an empty config if the file doesn't exist.

Can you think of a reason to fail instead? How would a user indicate that the config is optional?

pkg/config/engine/crio/option.go
elezar: What about `tomlSource`?

tools/container/crio/crio.go
elezar: I think `WithConfigSource` is clearer. Have updated.

pkg/config/engine/api.go
tariq1890: Understood

pkg/config/toml/source-file.go
tariq1890: A path to a file that does not exist should be considered as invalid input IMO. So we should error out accordingly

PR #659, opened 2024-08-19
Use empty string for default runtime-config-override

### LINKED ISSUE #631: NVIDIA Runtime Configuration Warning in Docker
I am experiencing a warning related to the NVIDIA runtime configuration while setting up Docker. Despite configuring the NVIDIA runtime correctly, I am seeing the following warning message:

sudo nvidia-ctk runtime configure --runtime=docker

WARN[0000] Ignoring runtime-config-override flag for docker
INFO[0000] Loading config from /etc/docker/daemon.json
INFO[0000] Wrote updated config to /etc/docker/daemon.json
INFO[0000] It is recommended that docker daemon be restarted.

Could you please help me understand why the runtime-config-override flag is being ignored and how to ensure that the NVIDIA runtime is properly set up as the default or specified runtime in Docker?

Thank you!

Fatemeh

elezar:
Cherry picks #658

Fixes #631

### DIFF
--- a/internal/discover/mounts_test.go
+++ b/internal/discover/mounts_test.go
@@ -100,7 +100,7 @@ func TestMounts(t *testing.T) {
 				lookup: &lookup.LocatorMock{
 					LocateFunc: func(s string) ([]string, error) {
 						if s == "error" {
-							return nil, fmt.Errorf(s)
+							return nil, fmt.Errorf("error")
 						}
 						return []string{s}, nil
 					},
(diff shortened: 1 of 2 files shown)


klueska (approved):

cdesiniotis (approved):

internal/discover/mounts_test.go
```diff
@@ -100,7 +100,7 @@ func TestMounts(t *testing.T) {
 				lookup: &lookup.LocatorMock{
 					LocateFunc: func(s string) ([]string, error) {
 						if s == "error" {
-							return nil, fmt.Errorf(s)
+							return nil, fmt.Errorf("error")
```
tariq1890, line 103: nit: (Optional) If we are using constant strings, we can use the stdlib `errors.New` instead. Also we need a better error message than just `error` as it doesn't say much. I understand that it's a test case, but expressive error messages also help us better debug and trace unit test failures
```suggestion
							return nil, fmt.Errorf("error")
```

elezar: Those are all valid points. Since this is a cherry-pick, I'm not going to address these here and would prefer to keep the functionality the same.

PR #660, opened 2024-08-20
Exclude libnvidia-allocator from graphics mounts

elezar:
This change ensures that the `libnvidia-allocator.so.RM_VERSION` mount is not injected BOTH by the `nvidia-container-runtime` through modifying the OCI spec AND libnvidia-container.

First injecting the mount through the OCI spec and then also following the mount logic implemented in `libnvidia-container` causes mounts to be leaked to the host when bind-propagation is enabled in a container.

To verify these changes set up the following folder:
```
$ mkdir -p /tmp/empty
```

Run the existing `nvidia` runtime:
```
$ nvidia-container-runtime --version
NVIDIA Container Runtime version 1.16.1
commit: a470818ba7d9166be282cd0039dd2fc9b0a34d73
spec: 1.2.0

runc version 1.1.12
commit: v1.1.12-0-g51d5e94
spec: 1.0.2-dev
go: go1.21.9
libseccomp: 2.5.1
```
```
$ docker run --runtime=nvidia --rm -ti -e NVIDIA_VISIBLE_DEVICES=all -e NVIDIA_DRIVER_CAPABILITIES=all \
     --mount type=bind,source=/tmp/empty,target=/empty,bind-propagation=shared \
     ubuntu mount | grep allocator
/dev/md0 on /usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15 type ext4 (ro,nosuid,nodev,relatime,nodelalloc,errors=remount-ro)
/dev/md0 on /usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15 type ext4 (ro,nosuid,nodev,relatime,nodelalloc,errors=remount-ro)
```
Note that the mounts are leaked to the host:
```
$ mount | grep allocator
/dev/md0 on /usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15 type ext4 (rw,relatime,nodelalloc,errors=remount-ro)
```

Clean up the mount:
```
$ sudo umount /usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15
```

With these changes:
```
$ ./nvidia-container-runtime --version
NVIDIA Container Runtime version 1.16.1
commit: beb921fafebaf416cf03a3dcfc5d250a03d325b2
spec: 1.2.0

runc version 1.1.12
commit: v1.1.12-0-g51d5e94
spec: 1.0.2-dev
go: go1.21.9
libseccomp: 2.5.1
```

```
$ docker run --runtime=elezar --rm -ti -e NVIDIA_VISIBLE_DEVICES=all -e NVIDIA_DRIVER_CAPABILITIES=all \
     --mount type=bind,source=/tmp/empty,target=/empty,bind-propagation=shared \
     ubuntu mount | grep allocator
/dev/md0 on /usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15 type ext4 (ro,nosuid,nodev,relatime,nodelalloc,errors=remount-ro)
```

Note that no mounts are present on the host:
```
$ mount | grep allocator
```

### DIFF
--- a/internal/discover/graphics.go
+++ b/internal/discover/graphics.go
@@ -146,6 +146,27 @@ func newGraphicsLibrariesDiscoverer(logger logger.Interface, driver *root.Driver
 	}
 }
 
+// Mounts discovers the required libraries and filters out libnvidia-allocator.so.
+// The library libnvidia-allocator.so is already handled by either the *.RM_VERSION
+// injection or by libnvidia-container. We therefore filter it out here as a
+// workaround for the case where libnvidia-container will re-mount this in the
+// container, which causes issues with shared mount propagation.
+func (d graphicsDriverLibraries) Mounts() ([]Mount, error) {
+	mounts, err := d.Discover.Mounts()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get library mounts: %v", err)
+	}
+
+	var filtered []Mount
+	for _, mount := range mounts {
+		if d.isDriverLibrary(filepath.Base(mount.Path), "libnvidia-allocator.so") {
+			continue
+		}
+		filtered = append(filtered, mount)
+	}
+	return filtered, nil
+}
+
 // Create necessary library symlinks for graphics drivers
 func (d graphicsDriverLibraries) Hooks() ([]Hook, error) {
 	mounts, err := d.Discover.Mounts()
--- a/internal/discover/graphics_test.go
+++ b/internal/discover/graphics_test.go
@@ -62,11 +62,7 @@ func TestGraphicsLibrariesDiscoverer(t *testing.T) {
 					return mounts, nil
 				},
 			},
-			expectedMounts: []Mount{
-				{
-					Path: "/usr/lib64/libnvidia-allocator.so.123.45.67",
-				},
-			},
+			expectedMounts: nil,
 			expectedHooks: []Hook{
 				{
 					Lifecycle: "createContainer",
@@ -121,9 +117,6 @@ func TestGraphicsLibrariesDiscoverer(t *testing.T) {
 				},
 			},
 			expectedMounts: []Mount{
-				{
-					Path: "/usr/lib64/libnvidia-allocator.so.123.45.67",
-				},
 				{
 					Path: "/usr/lib64/libnvidia-vulkan-producer.so.123.45.67",
 				},


cdesiniotis (approved):

klueska:
I believe you that this fixes things, but I still don't quite understand the root cause. Why does mounting a directory from `/tmp/empty` on the host to `/empty` inside the container (with or without bidirectional propagation) result in a leaked mount at `/usr/lib/x86_64-linux-gnu/`?

klueska:
Also, why is `libnvidia-allocator.so` special? What about all of the other libraries that are getting injected as bind-mounts?

elezar:
> 
Also, why is libnvidia-allocator.so special? What about all of the other libraries that are getting injected as bind-mounts?

The difference between `libnvidia-allocator.so` is that this is the only file that is mounted by both `libnvidia-container` and by injecting it into the OCI specification. In libnvidia-container we call `xmount` twice to mount and then remount the lib (see https://github.com/NVIDIA/libnvidia-container/blob/4c2494f16573b585788a42e9c7bee76ecd48c73d/src/nvc_mount.c#L135-L138). This seems to interact poorly with the fact that the file has already been mounted using the OCI spec.

By adjusting the `NVIDIA_DRIVER_CAPABILITIES` envvar, one can trigger any combination of the code paths in the above case and we only leak mounts if both modes are triggered.

elezar:
> Why does mounting a directory from /tmp/empty on the host to /empty inside the container (with or without bidirectional propagation) result in a leaked mount at /usr/lib/x86_64-linux-gnu/?

it only leaks the mount *if bidirectional propagation is enabled*. If a "regular" bind mount is selected, the mounts are not leaked.

klueska:
> it only leaks the mount _if bidirectional propagation is enabled_. If a "regular" bind mount is selected, the mounts are not leaked.
> 
> This interaction seems to be a known issue in docker. For example, the following issue [moby/moby#48305](https://github.com/moby/moby/issues/48305) shows the same behaviour when including multiple overlapping mounts on the docker command line.

I still don't understand completely. The mounts aren't overlapping in this case though, right? The command line mounts from `/tmp/empty` to `/empty` and `libnvidia-container` / `runc` (i.e. OCI) mount things from `/usr/lib/x86_64-linux-gnu/` to  `/usr/lib/x86_64-linux-gnu/`.

Where is the overlap?

elezar:
> Where is the overlap?

The overlap is between the two mechanims for mounting `/usr/lib/x86_64-linux-gnu/libnvidia-allocator.<RM_VERSION>`. This is(was) injected as a mount into the OCI runtime specification and then mounted by calls to `xmount` in `libnvidia-container`.

klueska:
> > Where is the overlap?
> 
> The overlap is between the two mechanims for mounting `/usr/lib/x86_64-linux-gnu/libnvidia-allocator.<RM_VERSION>`. This is(was) injected as a mount into the OCI runtime specification and then mounted by calls to `xmount` in `libnvidia-container`.

Then how / why does `--mount type=bind,source=/tmp/empty,target=/empty,bind-propagation=shared` come into play? Isn't the shared propogation limited to this single mount point?

elezar:
Upon further investigation I should point out that when including any mount with `bind-propagation=shared` docker and containerd also mount the rootfs with shared mount propagation. In this case, we create a mount at `/usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15` through OCI spec injection and then also mount at this same location using `libnvidia-container`. This is the mount that leaks and cand be confirmed by running:
```
mount --bind /usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15 /usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15
```
in the container, with each of these mounts being visible on the host.

Another thing to note is that if the `-v` flag is used to mount a file, the `rprivate` option is added to the OCI spec by default, this means that if we run something like:
```
docke run --rm -ti --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all -e NVIDIA_DRIVER_CAPABILITIES=all -v /usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15:/usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15 ubuntu
```
and then run:
```
mount --bind /usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15 /usr/lib/x86_64-linux-gnu/libnvidia-allocator.so.550.54.15
```
in the container this mount is not visible on the host. This could mean that a longer-term fix could be to add the `rprivate` option to the options in the internal CDI representation.

klueska:
After chatting offline, let's go ahead and merge this as is for now. I’d like to explore the `rprivate` solution mentioned above as a follow-up though.

klueska (approved):

PR #690, opened 2024-09-11
Only allow host-relative LDConfig paths

elezar:
This change only allows host-relative LDConfig paths (i.e. the `nvidia-container-cli.ldconfig` setting starts with an `@`)

An `allow-ldconfig-from-container` feature flag is added to allow for this
the default behaviour to be changed.

This feature can be toggled by running:
```
sudo nvidia-ctk config --in-place --set features.allow-ldconfig-from-container=true
```
or by including `allow-config-from-container` in the `NVIDIA_CONTAINER_TOOLKIT_OPT_IN_FEATURES` envvar.

### DIFF
--- a/internal/config/cli.go
+++ b/internal/config/cli.go
@@ -17,6 +17,7 @@
 package config
 
 import (
+	"fmt"
 	"os"
 	"strings"
 )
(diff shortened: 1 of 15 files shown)


internal/config/cli.go
```diff
@@ -47,7 +47,7 @@ func (c *ContainerCLIConfig) NormalizeLDConfigPath() string {
 // NormalizeLDConfigPath returns the resolved path of the configured LDConfig binary.
 // This is only done for host LDConfigs and is required to handle systems where
 // /sbin/ldconfig is a wrapper around /sbin/ldconfig.real.
-func NormalizeLDConfigPath(path string) string {
+func NormalizeLDConfigPath(path string) (rpath string) {
```
cdesiniotis, line 50: Why is this change required?

internal/config/config_test.go
```diff
@@ -293,6 +340,7 @@ func TestGetConfig(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.description, func(t *testing.T) {
+			defer setGetLdConfigPathForTest()
```
cdesiniotis, line 343: Question: why do we `defer` the execution of this function call? Don't we need to set the ldconfig path for the test case we are about to execute?

elezar: You're right. This should be:
```suggestion
			defer setGetLdConfigPathForTest()()
```

cdesiniotis (approved):

klueska (dismissed):
1. Let's continue to allow host-relative path to ldconfig even without the override
2. Let's then change the override flag to be called `allow-ldconfig-from-container`

elezar:
> Question -- is it possible to include a comment in the default NVIDIA Container Toolkit config file that gets installed at /etc/nvidia-container-runtime/config.toml which documents the new feature flag we are adding?

The features are currently `omitempty`, meaning that if none of them are set we would not show them. I can look at adding a comment similar to the docstring to the `ldconfig` option though.

cdesiniotis (approved):

PR #694, opened 2024-09-16
Skip injection of nvidia-persistenced socket by default

elezar:
This changes skips the injection of the nvidia-persistenced socket by default.

An include-persistenced-socket feature flag is added to allow the
injection of this socket to be explicitly requested.

See https://github.com/NVIDIA/libnvidia-container/pull/283

### DIFF
--- a/cmd/nvidia-container-runtime-hook/main.go
+++ b/cmd/nvidia-container-runtime-hook/main.go
@@ -89,6 +89,12 @@ func doPrestart() {
 	rootfs := getRootfsPath(container)
 
 	args := []string{getCLIPath(cli)}
+
+	// Only include the nvidia-persistenced socket if it is explicitly enabled.
+	if !hook.Features.IncludePersistencedSocket.IsEnabled() {
+		args = append(args, "--no-persistenced")
+	}
+
 	if cli.Root != "" {
 		args = append(args, fmt.Sprintf("--root=%s", cli.Root))
 	}
--- a/internal/discover/ipc.go
+++ b/internal/discover/ipc.go
@@ -24,7 +24,13 @@ import (
 type ipcMounts mounts
 
 // NewIPCDiscoverer creats a discoverer for NVIDIA IPC sockets.
-func NewIPCDiscoverer(logger logger.Interface, driverRoot string) (Discover, error) {
+func NewIPCDiscoverer(logger logger.Interface, driverRoot string, includePersistencedSocket bool) (Discover, error) {
+	var requiredSockets []string
+	if includePersistencedSocket {
+		requiredSockets = append(requiredSockets, "/nvidia-persistenced/socket")
+	}
+	requiredSockets = append(requiredSockets, "/nvidia-fabricmanager/socket")
+
 	sockets := newMounts(
 		logger,
 		lookup.NewFileLocator(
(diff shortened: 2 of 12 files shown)


cdesiniotis (approved):

cmd/nvidia-container-runtime-hook/main.go
```diff
@@ -89,6 +89,16 @@ func doPrestart() {
 	rootfs := getRootfsPath(container)
 
 	args := []string{getCLIPath(cli)}
+
+	// Opt out of the fabricmanager and persistenced sockets if these are not
+	// explicitly enabled.
+	if !hook.Features.AllowFabricmanager.IsEnabled() {
```
shivamerla, line 95: So, containers depending on this socket (e.g. NCCL library) will error out after this update by default? This will impact multi-node training Jobs etc.

klueska (dismissed):
1. Let's do just persistenced for now (not fabricmanager)
2. Let's change the feature-gate name to `include-persistenced` not `allow-persistenced`

internal/modifier/gated.go
```diff
@@ -46,31 +46,31 @@ func NewFeatureGatedModifier(logger logger.Interface, cfg *config.Config, image
 	driverRoot := cfg.NVIDIAContainerCLIConfig.Root
 	devRoot := cfg.NVIDIAContainerCLIConfig.Root
 
-	if cfg.Features.IsEnabled(config.FeatureGDS, image) {
+	if cfg.Features.IsEnabledInEnvironment(config.FeatureGDS, image) {
```
cdesiniotis, line 49: Question -- Is `config.FeatureGDS` the environment variable,`NVIDIA_GDS` ? Or is it the feature name `gds`?

cdesiniotis: On a quick pass, this appears to be `gds` when I would expect us to be passing `NVIDIA_GDS`.

cdesiniotis: Okay I am wrong. We are passing the feature name, and `IsEnabledInEnvironment` will translate this to the envvar name.

elezar: the feature name is `gds`, but this calls:
```
	case FeatureGDS:
		return fs.GDS.isEnabledWithEnvvarOverride("NVIDIA_GDS", in...)
```
Where the envvar is used.

internal/discover/ipc.go
```diff
@@ -34,10 +39,7 @@ func NewIPCDiscoverer(logger logger.Interface, driverRoot string) (Discover, err
 			lookup.WithCount(1),
 		),
 		driverRoot,
-		[]string{
-			"/nvidia-persistenced/socket",
-			"/nvidia-fabricmanager/socket",
```
cdesiniotis, line 39: Was removal of the fabricmanager socket intended?

cdesiniotis (approved):

cdesiniotis (changes requested):
Let's not remove the fabric-manager socket as part of this PR.

internal/discover/ipc.go
elezar: No. Updated.

cdesiniotis (approved):

cmd/nvidia-container-runtime-hook/main.go
elezar: Changing the scope of this PR to only `nvidia-persistenced`.

PR #696, opened 2024-09-17
Check for valid paths in create-symlinks hook

elezar:
This change updates the create-symlinks hook to check whether link paths resolve in the container's filesystem. In addition the executable is updated to return an error if a link could not be created.

### DIFF
--- a/cmd/nvidia-cdi-hook/create-symlinks/create-symlinks.go
+++ b/cmd/nvidia-cdi-hook/create-symlinks/create-symlinks.go
@@ -149,13 +149,12 @@ func (m command) run(c *cli.Context, cfg *config) error {
 	for _, l := range links {
 		parts := strings.Split(l, "::")
 		if len(parts) != 2 {
-			m.logger.Warningf("Invalid link specification %v", l)
-			continue
+			return fmt.Errorf("invalid symlink specification %v", l)
 		}
 
 		err := m.createLink(created, cfg.hostRoot, containerRoot, parts[0], parts[1])
 		if err != nil {
-			m.logger.Warningf("Failed to create link %v: %v", parts, err)
+			return fmt.Errorf("failed to create link %v: %w", parts, err)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/nvidia-cdi-hook/create-symlinks/create-symlinks.go
```diff
@@ -191,6 +201,33 @@ func (m command) createLink(created map[string]bool, hostRoot string, containerR
...
+// If the path does not exist, the parents of the path are resolved instead.
+func tryResolveLink(path string) (string, error) {
+	if path == "" || path == "/" {
+		return path, nil
+	}
+
+	resolved, err := filepath.EvalSymlinks(path)
+	if os.IsNotExist(err) {
+		resolvedParent, err := tryResolveLink(filepath.Dir(path))
+		return filepath.Join(resolvedParent, filepath.Base(path)), err
```
klueska, line 225: let's put an if/else here and return "" if there is an error

cmd/nvidia-cdi-hook/create-symlinks/create-symlinks.go
```diff
@@ -191,6 +201,33 @@ func (m command) createLink(created map[string]bool, hostRoot string, containerR
...
+		return err
+	}
+	if !strings.HasPrefix(resolved, root) {
+		return fmt.Errorf("path %v resolves to %v which is outside the root %v", path, resolved, root)
+	}
+	return nil
+}
+
+// tryResolveLink resolves the specified path following symlinks.
+// If the path does not exist, the parents of the path are resolved instead.
```
klueska, line 216:
```suggestion
// If the path does not exist, recursively resolve the parent of the path until one resolves successfully.
```
```suggestion
// If the path does not exist, the parents of the path are resolved instead.
```

klueska (dismissed):
a few nits, otherwise looks good

cdesiniotis (approved):

PR #700, opened 2024-09-18
Replace shell wrapper with a Go wrapper

jfroy:
This PR replaces the shell wrapper used by the container toolkit installed with a Go program. This allows the GPU operator to install the toolkit on distributions that do not have a shell, such as [Talos Linux](https://www.talos.dev/).

The Go wrapper reads its configuration from a dedicated ELF section, which although brings some complexity, allows for single-file standalone wrappers with room to expand the configuration options.

For environment variables, special prefixes can be used to prepend or append to a colon-delimited list variable, such a `PATH`. '<' is used to prepend and '>' is used to append. For example, `<PATH=foo` will prepend `foo` to `PATH`.


jfroy:
An alternative to the side files would be to embed the arguments and environments in custom ELF sections. It felt more opaque than side files, but let me know if you'd prefer that approach and I'll implement it.

elezar (commented):
@jfroy one thought I would have here is whether it makes sense to rather pull this functionality into the main runtime and feature gate it or obfuscate it behind a `mode` flag. We already produce mode-specific binaries for `cdi` and `legacy` for use on the operator and this could be an extension of that.

Not sure what @klueska's thoughts are on this though.

jfroy:
The nvidia module detection logic seems like a natural fit to be moved in the runtime program. It's less clear that the argv and envv stuff would fit there -- the purpose of that indirection is to allow the gpu-operator to customize the installation of the toolkit to the cluster, and to its parameters (e.g. host-provided driver vs driver container, etc).

--

As a third alternative implementation, the wrapper could look for a dot-config file recursively up the directory tree, aka .gitconfig. This would be a good fit as the wrappers mostly all have the same argv and envv computed by the toolkit installer.

elezar:
> As a third alternative implementation, the wrapper could look for a dot-config file recursively up the directory tree, à la .gitconfig. This would be a good fit as the wrappers mostly all have the same argv and envv computed by the toolkit installer.

It may not have to be up the tree. We could create a single `.env` file in the "toolkit" root and read it in the current folder?

By the way, I have created https://github.com/NVIDIA/nvidia-container-toolkit/pull/763 which should allow us to simplify this implementation further since we would not have to deal with command line arguments in addition to envvironment variables.

tools/container/wrapper/wrapper.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func isRuntimeWrapper(program string) bool {
+	return filepath.Base(program) == "nvidia-container-runtime" ||
+		filepath.Base(program) == "nvidia-container-runtime.cdi" ||
+		filepath.Base(program) == "nvidia-container-runtime.legacy"
+}
+
+func isNvidiaModuleLoaded() bool {
+	_, err := os.Stat("/proc/driver/nvidia/version")
```
elezar, line 59: What is the motivation for using this over:
```
cat /proc/modules | grep -e \"^nvidia \" >/dev/null 2>&1
```
here?

elezar: Note that I have also created https://github.com/NVIDIA/nvidia-container-toolkit/pull/764 which *should* pull this logic into the runtime -- guarded by a feature flag which is set by the containerized "installer".

tools/container/wrapper/wrapper.go
```diff
@@ -0,0 +1,115 @@
...
+		if err != nil {
+			log.Fatalf("failed to find runc: %v", err)
+		}
+		argv := []string{"runc"}
+		argv = append(argv, os.Args[1:]...)
+		execve(program, argv, os.Environ())
+	}
+	argv := makeArgv(program)
+	envv := makeEnvv(program)
+	execve(program+".real", argv, envv)
```
elezar, line 49: Question: Is there any reason to not call `syscal.Exec()` here directly?

jfroy:
> > As a third alternative implementation, the wrapper could look for a dot-config file recursively up the directory tree, à la .gitconfig. This would be a good fit as the wrappers mostly all have the same argv and envv computed by the toolkit installer.
> 
> It may not have to be up the tree. We could create a single `.env` file in the "toolkit" root and read it in the current folder?

Assuming all the tools can use a single set of environment variables, that would work,

> By the way, I have created #763 which should allow us to simplify this implementation further since we would not have to deal with command line arguments in addition to envvironment variables.

If I remember correctly, this means the patch could drop the ability to set arguments and only handle environment variables.

tools/container/wrapper/wrapper.go
jfroy: > What is the motivation for using this over:
> 
> ```
> cat /proc/modules | grep -e \"^nvidia \" >/dev/null 2>&1
> ```
> 
> here?

Avoid a regular expression.

jfroy: > Note that I have also created #764 which _should_ pull this logic into the runtime -- guarded by a feature flag which is set by the containerized "installer".

Great, we can drop that code.

tools/container/wrapper/wrapper.go
```diff
@@ -0,0 +1,115 @@
...
+		} else if strings.HasPrefix(kv[0], ">") {
+			kv[0] = kv[0][1:]
+			kv[1] = os.Getenv(kv[0]) + ":" + kv[1]
+		}
+		env = append(env, kv[0]+"="+kv[1])
+	}
+	if err := scanner.Err(); err != nil {
+		log.Fatalf("failed to read argv file: %v", err)
+	}
+	return append(env, os.Environ()...)
```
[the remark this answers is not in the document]
jfroy: In golang's[^1] and glibc's[^2] implementations, the first entry wins. I assume this is true for other languages (they probably all followed glibc). POSIX[^3] is mute on the required behavior.

[^1]: https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/syscall/env_unix.go;l=34
[^2]: https://sourceware.org/git/?p=glibc.git;a=blob;f=stdlib/getenv.c;hb=HEAD
[^3]: https://pubs.opengroup.org/onlinepubs/9799919799/functions/getenv.html

tools/container/wrapper/wrapper.go
jfroy: I just put it in a function to have a single point of error handling (not that fatal logging is much handling). I can inline this.

tools/container/wrapper/wrapper.go
jfroy: Ah thank you, I didn't appreciate that aspect. Then great, that can be removed.

jfroy:
@elezar I'd like to get this over the finish line. How do you you feel about your other PR?

jfroy:
I completely reworked the PR to rebase on the latest code and switch to a custom ELF section to store the wrapper configuration.

ArangoGutierrez:
/ok to test [f8fd37b](https://github.com/NVIDIA/nvidia-container-toolkit/pull/700/commits/f8fd37b82df6ebd7484dcb8defe7cfffa52f2c2a)

jfroy:
Looks like e2d passed. @elezar @cdesiniotis any further questions or feedback on this? It would be nice to not rely on a shell for host operating systems that are trying to be more "distroless" and locked down.

jfroy:
Rebased on latest main.

ssalagam:
@cdesiniotis , please review this. Thanks

rajatchopra:
#764 simplifies the kernel module check part. Perhaps that should go in first and then this can be rebased accordingly.

jfroy:
I can rebase this again!

cdesiniotis:
Closing in favor of https://github.com/NVIDIA/nvidia-container-toolkit/pull/1861 which provides an enablement path (centered around NRI) for getting the toolkit container to work on distributions, like Talos, which do not have a shell.

PR #715, opened 2024-09-26
Align driver symlinks with libnvidia-container

elezar:
This change aligns the creation of symlinks under CDI with
the implementation in libnvidia-container. If the driver libraries
are present, the following symlinks are created:
    
* {{ .LibRoot }}/libcuda.so -> libcuda.so.1
* {{ .LibRoot }}/libnvidia-opticalflow.so -> libnvidia-opticalflow.so.1
* {{ .LibRoot }}/libGLX_indirect.so.0 -> libGLX_nvidia.so.{{ .Version }}

### DIFF
--- a/pkg/nvcdi/driver-nvml.go
+++ b/pkg/nvcdi/driver-nvml.go
@@ -97,11 +97,15 @@ func NewDriverLibraryDiscoverer(logger logger.Interface, driver *root.Driver, nv
 		libraryPaths,
 	)
 
-	hooks, _ := discover.NewLDCacheUpdateHook(logger, libraries, nvidiaCDIHookPath, ldconfigPath)
+	updateLDCache, _ := discover.NewLDCacheUpdateHook(logger, libraries, nvidiaCDIHookPath, ldconfigPath)
 
 	d := discover.Merge(
-		libraries,
-		hooks,
+		discover.WithDriverDotSoSymlinks(
+			libraries,
+			version,
+			nvidiaCDIHookPath,
+		),
+		updateLDCache,
 	)
 
 	return d, nil
--- a/internal/discover/none.go
+++ b/internal/discover/none.go
@@ -24,15 +24,15 @@ var _ Discover = (*None)(nil)
 
 // Devices returns an empty list of devices
 func (e None) Devices() ([]Device, error) {
-	return []Device{}, nil
+	return nil, nil
 }
 
 // Mounts returns an empty list of mounts
 func (e None) Mounts() ([]Mount, error) {
-	return []Mount{}, nil
+	return nil, nil
 }
 
 // Hooks returns an empty list of hooks
 func (e None) Hooks() ([]Hook, error) {
-	return []Hook{}, nil
+	return nil, nil
 }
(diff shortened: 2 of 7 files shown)


pkg/nvcdi/driver-nvml.go
```diff
@@ -100,7 +106,11 @@ func NewDriverLibraryDiscoverer(logger logger.Interface, driver *root.Driver, nv
 	hooks, _ := discover.NewLDCacheUpdateHook(logger, libraries, nvidiaCDIHookPath, ldconfigPath)
 
 	d := discover.Merge(
-		libraries,
```
cdesiniotis, line 103: Did you intend to remove `libraries` from here? My understanding is that this PR is adding another hook (to create the `libcuda.so` symlink), not removing any of the discoverers in this list.

pkg/nvcdi/driver-nvml.go
```diff
@@ -220,3 +230,37 @@ func getVersionLibs(logger logger.Interface, driver *root.Driver, version string
...
+	mounts, err := d.Discover.Mounts()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get library mounts: %v", err)
+	}
+
+	var links []string
+	for _, mount := range mounts {
+		dir, filename := filepath.Split(mount.Path)
+		// TODO: We should include the other libraries as is done here:
+		// https://github.com/NVIDIA/nvidia-container-toolkit/blob/79c59aeb7f59dd612793ac80a8d7022c554634bb/internal/platform-support/tegra/symlinks.go#L84-L97
```
cdesiniotis, line 244: The code you linked for tegra creates `ibcuda.so -> libcuda.so.1` while the code below seems to be creating `ibcuda.so -> libcuda.so.RM_VERSION`. Why the difference?

elezar: You're right. I was trying to simplify things by linking to the `.RM_VERSION`, but since we have the following the driver manifest:
```
libcuda.so.550.120 0755 CUDA_LIB NATIVE / MODULE:gpgpu
libcuda.so.1 0000 CUDA_SYMLINK NATIVE / libcuda.so.550.120 MODULE:gpgpu
libcuda.so 0000 CUDA_SYMLINK NATIVE / libcuda.so.1 MODULE:gpgpu
```
we should link to `libcuda.so.1`.

klueska: Yes, the smylink should be to `libcuda.so.1` not the actual `.RM_VERSION` file.

klueska:
There are other libraries that we create these one-off symlinks for in libnvidia-container. Should we also add these as part of this PR?
https://github.com/NVIDIA/libnvidia-container/blob/921e2f3197385173cf8670342e96e98afe9b3dd3/src/nvc_mount.c#L548

elezar:
> There are other libraries that we create these one-off symlinks for in libnvidia-container. Should we also add these as part of this PR? https://github.com/NVIDIA/libnvidia-container/blob/921e2f3197385173cf8670342e96e98afe9b3dd3/src/nvc_mount.c#L548

Yes. This was the intent of the comment above the link. I think doing it now makes sense and will update.

pkg/nvcdi/driver-nvml.go
elezar: Updated in latest.

klueska (approved):

elezar:
Backport PR created as #724

PR #716, opened 2024-09-26
Also search for driver libraries in vdpau

elezar:
This change adds the vdpau subfolder to the paths searched for driver libraries. This allows the libvdpau_nvidia.so.RM_VERSION library to also be discovered.


cdesiniotis (approved):

elezar:
We should consider backporting these changes to `release-1.16`.

PR #717, opened 2024-09-26
Skip explicit creation of libnvidia-allocator.so.1 symlink

elezar:
Since we expect .so.1 symlinks to be created by ldconfig, we don't explicitly request these. This change removes the creation of a libnvidia-allocator.so.1 -> libnvidia-allocator.so.RM_VERSION symlink through the create-symlinks hook.


cdesiniotis:
Question -- does the ordering of hooks in the CDI spec file determine their execution order? E.g. if the `update-ldcache` hooks comes before the `create-symlinks` hook in the CDI spec, does the `update-ldcache` hook always run first?

elezar:
> Question -- does the ordering of hooks in the CDI spec file determine their execution order? E.g. if the `update-ldcache` hooks comes before the `create-symlinks` hook in the CDI spec, does the `update-ldcache` hook always run first?

Yes, I believe that the [OCI spec](https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-hooks) guarantees the order of execution. 

> Hooks allow users to specify programs to run before or after various lifecycle events. Hooks MUST be called in the listed order.

cdesiniotis (approved):

elezar:
These changes should be backported to `release-1.16`.

PR #686, opened 2024-09-09
Fetch current container runtime config

tariq1890:
Summary of changes made in this PR:

1. In CRI-O and Containerd config engines, retrieve the container runtime config commands via the following commands:
    i. `crio status config`
    ii. `containerd config dump`

2. When configuring CRI-O runtime, we prioritise the runtime designated as the `default_runtime` in the config when setting up the `nvidia` runtime handler as opposed to favouring the `runc` runtime handler at all times. This is needed as vanilla cri-o packages have `crun` as the default low-level-runtime. Live-swapping the low-level runtime causes the running containers in a cluster to break

3. Add the full path of the low-level runtime to the `nvidia-container-runtime.runtimes` config value, so that `nvidia-container-runtime` binds to a low-level runtime binary that cannot be resolved in the `PATH`.

PR #723, opened 2024-10-01
fetch current container runtime config through the command line

tariq1890:
Summary of changes made in this PR:

In CRI-O and Containerd config engines, retrieve the container runtime config commands via the following commands:
i. crio status config
ii. containerd config dump

When configuring CRI-O runtime, we prioritise the runtime designated as the default_runtime in the config when setting up the nvidia runtime handler as opposed to favouring the runc runtime handler at all times. This is needed as vanilla cri-o packages have crun as the default low-level-runtime. Live-swapping the low-level runtime causes the running containers in a cluster to break

Add the full path of the low-level runtime to the nvidia-container-runtime.runtimes config value, so that nvidia-container-runtime binds to a low-level runtime binary that cannot be resolved in the PATH.

This is the latest iteration of the #686 PR that is most up-to-date and will supersede the same


elezar:
Please close this PR and continue working on #686

PR #732, opened 2024-10-10
Add -z,lazy to LDFLAGS

elezar:
This fixes undefined symbol errors on platforms where -z,lazy may not be the default.

This was inadvertently removed in #333 and replaced with `-Wl,--export-dynamic -Wl,--unresolved-symbols=ignore-in-object-files` which seems to be equivalent on some systems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ endif
 cmds: $(CMD_TARGETS)
 
 ifneq ($(shell uname),Darwin)
-EXTLDFLAGS = -Wl,--export-dynamic -Wl,--unresolved-symbols=ignore-in-object-files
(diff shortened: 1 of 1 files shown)


klueska:
Is it an either/or with `--unresolved-symbols=ignore-in-object-files` and `-z,lazy` or is it OK on all platforms to have both?

elezar:
> Is it an either/or with `--unresolved-symbols=ignore-in-object-files` and `-z,lazy` or is it OK on all platforms to have both?

I can't make a comment on all platforms. From my testing, it works on Ubuntu, but let me check the documentation.

klueska (approved):

PR #734, opened 2024-10-10
minor cleanup and improvements

tariq1890:
i) Fix some typo errors
ii) Remove redundant import aliases
iii) Remove `deadline`, which is an invalid field in `golangci-lint.yml`, and replace it with `timeout`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
 run:
-  deadline: 10m
+  timeout: 10m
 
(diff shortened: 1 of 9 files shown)


elezar (commented):
Thanks @tariq1890.

Let's just update the commit message to start with the `[no-relnote]` tag in this case.

elezar (approved):

PR #736, opened 2024-10-11
Add -z,lazy to LDFLAGS

elezar:
This fixes undefined symbol errors on platforms where -z,lazy may not be the default.

Backports the changes from #732


elezar:
Closing since we have released v1.17.0.

PR #737, opened 2024-10-11
Fix race condition in mounts cache

### LINKED ISSUE #731: The mounts discoverer cache usage is racey
The mount cache is read here without locking:
https://github.com/NVIDIA/nvidia-container-toolkit/blob/4604e3b6c86aba52c7fe54241070d493938a28e6/internal/discover/mounts.go#L59-L67

But it is set here while locked: https://github.com/NVIDIA/nvidia-container-toolkit/blob/4604e3b6c86aba52c7fe54241070d493938a28e6/internal/discover/mounts.go#L116

Best case here is you end up with 2 (or technically possibly more) things running the mounts code needlessly.
Worse cases could be undefined behavior due to the [go memory model](https://go.dev/ref/mem).

elezar:
This change switches to using the WithCache decorator for mounts instead of keeping track of a cache locally.

This addresses a race condition when using the mounts structure.

Fixes #731

### DIFF
--- a/internal/discover/mounts.go
+++ b/internal/discover/mounts.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
-	"sync"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/lookup"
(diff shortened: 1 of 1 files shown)


jgehrcke (approved):
Thank you. I trust you know what you are doing.

PR #740, opened 2024-10-14
Allow IMEX channel requests by volume mount

elezar:
This allows IMEX channels to be requested as volume mounts in addition to the `NVIDIA_IMEX_CHANNELS` environment variable.

See https://github.com/NVIDIA/k8s-device-plugin/pull/985 where this is used in the NVIDIA Device Plugin when the `deviceListStrategy` is set to `volume-mounts`.

### DIFF
--- a/cmd/nvidia-container-runtime-hook/container_config.go
+++ b/cmd/nvidia-container-runtime-hook/container_config.go
@@ -6,7 +6,6 @@ import (
 	"log"
 	"os"
 	"path"
-	"path/filepath"
 
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/mod/semver"
@@ -15,23 +14,7 @@ import (
 )
 
 const (
-	envCUDAVersion          = "CUDA_VERSION"
-	envNVRequirePrefix      = "NVIDIA_REQUIRE_"
-	envNVRequireCUDA        = envNVRequirePrefix + "CUDA"
-	envNVDisableRequire     = "NVIDIA_DISABLE_REQUIRE"
-	envNVVisibleDevices     = "NVIDIA_VISIBLE_DEVICES"
-	envNVMigConfigDevices   = "NVIDIA_MIG_CONFIG_DEVICES"
-	envNVMigMonitorDevices  = "NVIDIA_MIG_MONITOR_DEVICES"
-	envNVImexChannels       = "NVIDIA_IMEX_CHANNELS"
-	envNVDriverCapabilities = "NVIDIA_DRIVER_CAPABILITIES"
-)
-
-const (
 	capSysAdmin = "CAP_SYS_ADMIN"
-)
-
-const (
-	deviceListAsVolumeMountsRoot = "/var/run/nvidia-container-devices"
 )
 
 type nvidiaConfig struct {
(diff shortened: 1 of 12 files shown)


cmd/nvidia-container-runtime-hook/container_config.go
```diff
@@ -31,15 +29,11 @@ const (
...
-	deviceListAsVolumeMountsRoot = "/var/run/nvidia-container-devices"
-)
-
 type nvidiaConfig struct {
-	Devices            string
+	Devices            []string
 	MigConfigDevices   string
 	MigMonitorDevices  string
-	ImexChannels       string
+	ImexChannels       []string
```
tariq1890, line 36:
```suggestion
	IMEXChannels       []string
```

elezar: I think `Imex` matches how we treat `Mig` in this context.

cmd/nvidia-container-runtime-hook/container_config.go
```diff
@@ -16,11 +14,11 @@ import (
...
-	envNVDisableRequire     = "NVIDIA_DISABLE_REQUIRE"
-	envNVVisibleDevices     = "NVIDIA_VISIBLE_DEVICES"
+	envCUDAVersion      = "CUDA_VERSION"
+	envNVRequirePrefix  = "NVIDIA_REQUIRE_"
+	envNVRequireCUDA    = envNVRequirePrefix + "CUDA"
+	envNVDisableRequire = "NVIDIA_DISABLE_REQUIRE"
+
 	envNVMigConfigDevices   = "NVIDIA_MIG_CONFIG_DEVICES"
 	envNVMigMonitorDevices  = "NVIDIA_MIG_MONITOR_DEVICES"
 	envNVImexChannels       = "NVIDIA_IMEX_CHANNELS"
```
klueska, line 24: If you are moving NVIDIA_VISIBLE_DEVICES should you also move NVIDIA_IMEX_CHANNELS?

That said -- I'M not a huge fan of splitting these up. Does it make sense to create a `consts` package (or soemthing less specific) that both the image package and this one can import to get access to these, so we can keep them all in the same place?

internal/config/image/cuda_image.go
```diff
@@ -218,14 +225,25 @@ func (i CUDA) OnlyFullyQualifiedCDIDevices() bool {
 	return hasCDIdevice
 }
 
-const (
-	deviceListAsVolumeMountsRoot = "/var/run/nvidia-container-devices"
-)
+// VisibleDevicesFromMounts returns the set of visible devices requested as mounts.
+func (i CUDA) VisibleDevicesFromMounts() []string {
```
klueska, line 229: Should there be a corresponding ` VisibleDevicesFromEnvVar()` for symmetry.

klueska (commented):
Minor comments, otherwise looks good

cmd/nvidia-container-runtime-hook/container_config.go
elezar: Since this code is importing the `image` package anyway, do you have an objection to putting them there?

elezar: I have reworked these constants. They should now all be in the `image` package.

internal/config/image/cuda_image.go
elezar: I renamed this. Note that `DevicesFromEnvvars` is a bit more general as it's also used for swarm support.

cmd/nvidia-container-runtime-hook/main.go
```diff
@@ -120,17 +120,17 @@ func doPrestart() {
...
 	if len(nvidia.MigConfigDevices) > 0 {
 		args = append(args, fmt.Sprintf("--mig-config=%s", nvidia.MigConfigDevices))
 	}
 	if len(nvidia.MigMonitorDevices) > 0 {
 		args = append(args, fmt.Sprintf("--mig-monitor=%s", nvidia.MigMonitorDevices))
 	}
-	if len(nvidia.ImexChannels) > 0 {
-		args = append(args, fmt.Sprintf("--imex-channel=%s", nvidia.ImexChannels))
+	if imexString := strings.Join(nvidia.ImexChannels, ","); len(imexString) > 0 {
+		args = append(args, fmt.Sprintf("--imex-channel=%s", imexString))
```
klueska, line 133: Minor comment -- not worth fixing in this PR -- but it would be nice to keep all of these consistent -- MigConfigDevices and MigMonitorDevices are still just strings rather than a list it looks like...

klueska (approved):

cmd/nvidia-container-runtime-hook/main.go
elezar: Yes. I will update in a follow-up.

PR #741, opened 2024-10-14
Add disableIMEXChannelCreation feature flag

elezar:
This change adds a `disableIMEXChannelCreation` feature flag that allows the `--no-create-imex-channels` option added  in https://github.com/NVIDIA/libnvidia-container/pull/286 to be toggled.

This allows IMEX channels to be exposed in environment where they are expected to be created up front.

See also https://github.com/NVIDIA/k8s-device-plugin/pull/985

### DIFF
--- a/internal/config/features.go
+++ b/internal/config/features.go
@@ -18,6 +18,9 @@ package config
 
 // features specifies a set of named features.
 type features struct {
+	// DisableImexChannelCreation ensures that the implicit creation of
+	// requested IMEX channels is skipped when invoking the nvidia-container-cli.
+	DisableImexChannelCreation *feature `toml:"disable-imex-channel-creation,omitempty"`
 }
 
 //nolint:unused
--- a/cmd/nvidia-container-runtime-hook/main.go
+++ b/cmd/nvidia-container-runtime-hook/main.go
@@ -95,6 +95,9 @@ func doPrestart() {
 	if cli.LoadKmods {
 		args = append(args, "--load-kmods")
 	}
+	if hook.Features.DisableImexChannelCreation.IsEnabled() {
+		args = append(args, "--no-create-imex-channels")
+	}
 	if cli.NoPivot {
 		args = append(args, "--no-pivot")
 	}
--- a/tools/container/toolkit/toolkit.go
+++ b/tools/container/toolkit/toolkit.go
@@ -79,6 +79,8 @@ type Options struct {
 	acceptNVIDIAVisibleDevicesAsVolumeMounts   bool
 
 	ignoreErrors bool
+
+	optInFeatures cli.StringSlice
 }
 
 func Flags(opts *Options) []cli.Flag {
@@ -201,6 +203,12 @@ func Flags(opts *Options) []cli.Flag {
 			Value:       cli.NewStringSlice("control"),
 			Destination: &opts.createDeviceNodes,
 			EnvVars:     []string{"CREATE_DEVICE_NODES"},
+		},
+		&cli.StringSliceFlag{
+			Name:        "opt-in-features",
+			Hidden:      true,
+			Destination: &opts.optInFeatures,
+			EnvVars:     []string{"NVIDIA_CONTAINER_TOOLKIT_OPT_IN_FEATURES"},
 		},
 	}
 
@@ -457,6 +465,10 @@ func installToolkitConfig(c *cli.Context, toolkitConfigPath string, nvidiaContai
 		configValues["nvidia-container-runtime.runtimes"] = toolkitRuntimeList
 	}
 
+	for _, optInFeature := range opts.optInFeatures.Value() {
+		configValues["features."+optInFeature] = true
+	}
+
 	for key, value := range configValues {
 		cfg.Set(key, value)
 	}
@@ -471,6 +483,7 @@ func installToolkitConfig(c *cli.Context, toolkitConfigPath string, nvidiaContai
 		"nvidia-container-runtime.runtimes":                      opts.ContainerRuntimeRuntimes,
 		"nvidia-container-cli.debug":                             opts.ContainerCLIDebug,
 	}
+
 	for key, value := range optionalConfigValues {
 		if !c.IsSet(key) {
 			log.Infof("Skipping unset option: %v", key)


internal/config/features.go
```diff
@@ -31,53 +33,57 @@ type features struct {
 	MOFED    *feature `toml:"mofed,omitempty"`
 	NVSWITCH *feature `toml:"nvswitch,omitempty"`
 	GDRCopy  *feature `toml:"gdrcopy,omitempty"`
+	// DisableImexChannelCreation ensures that the implicit creation of
+	// requested IMEX channels is skipped when invoking the nvidia-container-cli.
+	DisableImexChannelCreation *feature `toml:"disable-imex-channel-creation,omitempty"`
```
tariq1890, line 38: For consistency, we should name this as `IMEXChannel` imo

elezar: Do you mean `DisableIMEXChannelCreation`? (as in change the case to `IMEX`?

If you mean `IMEXChannel`, I don't agree. The intent of a feature called `IMEXChannel` as this feature sets a mode in libnvidia-container to NOT create IMEX channel device nodes automatically.

internal/config/features.go
```diff
@@ -23,6 +23,8 @@ const (
 	FeatureMOFED    = featureName("mofed")
 	FeatureNVSWITCH = featureName("nvswitch")
 	FeatureGDRCopy  = featureName("gdrcopy")
+
+	FeatureDisableIMEXChannelCreation = featureName("disable-imex-channel-creation")
```
klueska, line 27: Why the newline?

internal/config/features.go
```diff
@@ -31,53 +33,57 @@ type features struct {
...
+		return fs.MOFED.isEnabledWithEnvvarOverride("NVIDIA_MOFED", in...)
 	case FeatureNVSWITCH:
-		return fs.NVSWITCH.isEnabled(envvar, in...)
+		return fs.NVSWITCH.isEnabledWithEnvvarOverride("NVIDIA_NVSWITCH", in...)
 	case FeatureGDRCopy:
-		return fs.GDRCopy.isEnabled(envvar, in...)
+		return fs.GDRCopy.isEnabledWithEnvvarOverride("NVIDIA_GDRCOPY", in...)
+	// Features without envvar overrides
+	case FeatureDisableIMEXChannelCreation:
+		return fs.DisableIMEXChannelCreation.IsEnabled()
```
klueska, line 59: Is this just for backwards compatibility? If we could do things over again would we *not* create one envvar per feature?

internal/config/features.go
```diff
@@ -31,53 +33,57 @@ type features struct {
 	MOFED    *feature `toml:"mofed,omitempty"`
 	NVSWITCH *feature `toml:"nvswitch,omitempty"`
 	GDRCopy  *feature `toml:"gdrcopy,omitempty"`
+	// DisableIMEXChannelCreation ensures that the implicit creation of
+	// requested IMEX channels is skipped when invoking the nvidia-container-cli.
+	DisableIMEXChannelCreation *feature `toml:"disable-imex-channel-creation,omitempty"`
 }
 
 type feature bool
```
klueska, line 41: Does it make simplify things if this was extended to:
```
type feature struct {
    enabled bool
    envvarOverride string
}
```

Then you can define  `isEnabled()` generically as:
```
func (f *feature) isEnabled(ins ...getenver) bool {
	if f == nil {
		return false
	}
	if f.envvar != "" {
		for _, in := range ins {
			if in.Getenv(envvar) == "enabled" {
				return true
			}
		}
	}
	return f.enabled
}
```

klueska: Nevermind -- I see that the feature type is embedded directly in the config, so this wouldn't work.

elezar: Yes, the intent is to have something like the following:
```
[features]
  disable-imex-channel-creation = true
  another-opt-in-feature = true
  another-feature-explicitly-disabled = true
```

internal/config/features.go
elezar: It's not just for backward compatibility. The features with envvar overrides are ones that we allow users to select on a per-container basis. This feature is global an MUST be set for a given installation.

internal/config/features.go
elezar: We could look into having a type per feature and then associating the envvar with the type instead, but I don't really see the benefit of that now.

internal/config/features.go
elezar: I sometimes introduce newlines to prevent additional whitespace changes due to aligning the `=` in the assignment. Can roll it back if required.

internal/config/features.go
elezar: @klueska I was thinking the `DisableImexChannelCreation` actually reads better than `DisableIMEXChannelCreation`. Do you have thoughts on this?

internal/config/features.go
```diff
@@ -31,53 +33,57 @@ type features struct {
...
 	case FeatureMOFED:
-		return fs.MOFED.isEnabled(envvar, in...)
+		return fs.MOFED.isEnabledWithEnvvarOverride("NVIDIA_MOFED", in...)
 	case FeatureNVSWITCH:
-		return fs.NVSWITCH.isEnabled(envvar, in...)
+		return fs.NVSWITCH.isEnabledWithEnvvarOverride("NVIDIA_NVSWITCH", in...)
 	case FeatureGDRCopy:
-		return fs.GDRCopy.isEnabled(envvar, in...)
+		return fs.GDRCopy.isEnabledWithEnvvarOverride("NVIDIA_GDRCOPY", in...)
+	// Features without envvar overrides
```
elezar, line 57:
```suggestion
	// Global features.
	// These cannot be set on a per-container bases and as such don't have envvar overrides
```

internal/config/features.go
```diff
@@ -31,53 +33,57 @@ type features struct {
...
-		FeatureGDS:      "NVIDIA_GDS",
-		FeatureMOFED:    "NVIDIA_MOFED",
-		FeatureNVSWITCH: "NVIDIA_NVSWITCH",
-		FeatureGDRCopy:  "NVIDIA_GDRCOPY",
-	}
-
-	envvar := featureEnvvars[n]
+func (fs features) IsEnabledInEnvironment(n featureName, in ...getenver) bool {
 	switch n {
+	// Features with envvar overrides
```
elezar, line 48:
```suggestion
	// Features with per-container overrides using environment variables.
```

internal/config/features.go
elezar: Thanks for the review @klueska. I have reworked the feature handling to introduce a type alias for each feature.

This adds a little more boiler plate, but removes the switch statement that we had. The other option that I was considering was to implement as custom unmarshaller for the `features` struct and extract the envvar from struct tags using reflection. The in-memory representation could then be something that you propose. I don't think that this is worth blocking on now though.

internal/config/features.go
klueska: Yes, I agree with `DisableImexChannelCreation`. It's the same rationale we had for writing `Mig` instead of `MIG`.

internal/config/features.go
klueska: OK, I actually like it when its changed to line up the `=`. It makes it more readable in the code (even if the diff isn't as obvious). But I don't feel strongly.

internal/config/features.go
klueska: Got it. It does make it feel like this is something different though, and maybe belongs elsewhere than bundled with these other options.

internal/config/features.go
elezar: I have updated this now. Thanks for the feedback.

klueska (approved):

PR #745, opened 2024-10-17
Fix symlink resolution error message

elezar:
This change fixes an incorrectly constructed error message in symlink resolution.

### DIFF
--- a/cmd/nvidia-cdi-hook/create-symlinks/create-symlinks.go
+++ b/cmd/nvidia-cdi-hook/create-symlinks/create-symlinks.go
@@ -211,21 +211,14 @@ func changeRoot(current string, new string, path string) (string, error) {
 // Locate returns the link target of the specified filename or an empty slice if the
 // specified filename is not a symlink.
 func (m command) Locate(filename string) ([]string, error) {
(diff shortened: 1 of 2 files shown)


cdesiniotis (approved):

cmd/nvidia-cdi-hook/create-symlinks/create-symlinks.go
```diff
@@ -211,21 +211,14 @@ func changeRoot(current string, new string, path string) (string, error) {
 // Locate returns the link target of the specified filename or an empty slice if the
 // specified filename is not a symlink.
 func (m command) Locate(filename string) ([]string, error) {
-	info, err := os.Lstat(filename)
+	target, err := symlinks.Resolve(filename)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get file info: %v", info)
+		return nil, err
```
tariq1890, line 216: Why not return a wrapped error that mentions the symlink name?

elezar: I was trying to reduce the verbosity here. This is a low-level function and the caller can decide on whether to log / output more context.

PR #748, opened 2024-10-18
Add aliases for runtime-specific envvars

elezar:
This change ensures that the toolkit works with older versions of the GPU Operator where runtime-specific envvars are used to set options such as the config file location.

The support for these envvars was removed in #686

### DIFF
--- a/tools/container/runtime/runtime.go
+++ b/tools/container/runtime/runtime.go
@@ -50,14 +50,14 @@ func Flags(opts *Options) []cli.Flag {
 			Usage:       "Path to the runtime config file",
 			Value:       runtimeSpecificDefault,
 			Destination: &opts.Config,
-			EnvVars:     []string{"RUNTIME_CONFIG"},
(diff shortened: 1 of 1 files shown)


elezar:
@tariq1890 in addition to operator changes we also need to update the documentation referring to these envvars in the GPU Operator docs.

tools/container/runtime/runtime.go
```diff
@@ -50,14 +50,14 @@ func Flags(opts *Options) []cli.Flag {
 			Usage:       "Path to the runtime config file",
 			Value:       runtimeSpecificDefault,
 			Destination: &opts.Config,
-			EnvVars:     []string{"RUNTIME_CONFIG"},
+			EnvVars:     []string{"RUNTIME_CONFIG", "CONTAINERD_CONFIG", "DOCKER_CONFIG"},
```
tariq1890, line 53: This is missing the CRIO_ alias

tariq1890:
For posterity's sake, this was the PR which removed support for the runtime specific env vars

https://github.com/NVIDIA/nvidia-container-toolkit/pull/722

Issue #735, opened 2024-10-11
Allow LDCache to be used to discover libcuda.so in CDI spec generation


elezar:
On systems where driver libraries are not installed at "standard" library paths the `nvidia-ctk cdi generate --mode=nvml` command fails with output similar to:
```
# nvidia-ctk cdi generate --output=/etc/cdi/nvidia.yaml
WARN[0000] Failed to locate nvidia-cdi-hook: pattern nvidia-cdi-hook not found
INFO[0000] Using /var/lib/nvidia/lib64/libnvidia-ml.so.535.183.06
INFO[0000] Auto-detected mode as 'nvml'                
INFO[0000] Selecting /dev/nvidia0 as /dev/nvidia0      
INFO[0000] Selecting /dev/dri/card0 as /dev/dri/card0  
INFO[0000] Selecting /dev/dri/renderD128 as /dev/dri/renderD128
WARN[0000] failed to locate libcuda.so: pattern libcuda.so.*.* not found
64-bit library libcuda.so.*.*: not found; using *.*
INFO[0000] Using driver version 535.183.06              
ERRO[0000] failed to generate CDI spec: failed to create edits common for entities: failed to create discoverer for common entities: failed to create discoverer for driver files: failed to create discoverer for driver libraries: failed to get libraries for driver version: failed to locate libcuda.so.535.183.06: pattern libcuda.so.535.183.06 not found
64-bit library libcuda.so.535.183.06: not found
```

This can be rectified by running:
```
nvidia-ctk cdi generate --mode=nvml --library-search-path=/var/lib/nvidia/lib64
```

In the past the LDCache could be used on such systems allowing `libcuda.so` to be discovered successfully.

This was reported in v1.16.2

PR #751, opened 2024-10-19
add glibc/lib to library search paths

Hexoplon:
## Issue
When running Talos Linux, the Nvidia driver libraries are located under /usr/local/glibc/lib, and the binaries under /usr/local/bin. Using the default search paths, and a `driver-root` value of `/usr/local` the validator is able to locate the binaries, but not the libraries.

## Proposed solution
Add "glibc/lib" to the library search path. This has been tested to work on a Talos cluster running 1.8.1, production drivers (550.90.07) and H100 GPUs

This PR is part of a set of PRs to better support Talos Linux for the GPU Operator.

### DIFF
--- a/internal/lookup/library.go
+++ b/internal/lookup/library.go
@@ -56,6 +56,7 @@ func NewLibraryLocator(opts ...Option) Locator {
 			"/lib/aarch64-linux-gnu",
 			"/lib/x86_64-linux-gnu/nvidia/current",
 			"/lib/aarch64-linux-gnu/nvidia/current",
+			"/glibc/lib",
 		}...),
 	)
 	// We construct a symlink locator for expected library locations.


elezar:
@Hexoplon for CDI spec generation we already support specifying `--library-search-path` for alternative library search paths. Would exposing this through a config option address your use case?

It has recently been pointed out that certain ldcache-based lookups are not working as expected in the latest version. Could it be that this issue is also affecting you? See #735

Hexoplon:
@elezar Just noticed that Talos has made changes to the Nvidia extensions to support the GPU operator without requiring modifications, see https://github.com/siderolabs/extensions/commit/c7eb3771943eba55ec5fe61f7a31e6fce61d73d5. This should make this PR irrelevant.

PR #755, opened 2024-10-21
Fix bug where libcuda.so is not found in ldcache

elezar:
This change fixes a bug where libcuda.so cannot be located even if it is present in the ldcache. This is relevant on systems where the library is not present in one of the "standard" paths.


internal/lookup/root/root.go
```diff
@@ -79,6 +81,34 @@ func (r *Driver) Libraries() lookup.Locator {
...
+		return "", err
+	}
+
+	pattern := "*.*"
+	var matched []string
+	for _, candidate := range candidates {
+		// It is not guaranteed that candidate is a resolved symlink.
+		resolved, err := symlinks.Resolve(candidate)
+		if err != nil {
+			continue
```
tariq1890, line 97: Can we consider logging the error messages?

elezar: Done.

tariq1890 (approved):

internal/ldcache/ldcache.go
```diff
@@ -105,14 +101,7 @@ func New(logger logger.Interface, root string) (LDCache, error) {
...
 	f, err := os.Open(path)
-	if os.IsNotExist(err) {
-		logger.Warningf("Could not find ld.so.cache at %v; creating empty cache", path)
-		e := &empty{
-			logger: logger,
-			path:   path,
-		}
-		return e, nil
-	} else if err != nil {
+	if err != nil {
```
klueska, line 104: So this change implies that we error out unconditionally if there is not an ld.s.cache file now?

internal/lookup/library_test.go
```diff
@@ -24,81 +24,77 @@ import (
...
+// 	_, err := os.Create(versionLib)
+// 	require.NoError(t, err)
+// 	require.NoError(t, os.Symlink(versionLib, sonameLink))
+// 	require.NoError(t, os.Symlink(sonameLink, soLink))
+
+// 	lut := NewLdcacheLocator(
+// 		WithLogger(logger),
+// 		WithRoot(testDir),
+// 	)
+
```
klueska, line 49: What is all of this commented out test code?

elezar: I need to remove it.

internal/ldcache/ldcache.go
elezar: Yes, but this is handled in the caller now:
```
		b.logger.Warningf("Failed to load ldcache: %v", err)
		if b.isOptional {
			return &null{}
		}
		return &notFound{}
```

internal/lookup/library-ldcache.go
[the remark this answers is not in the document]
elezar: I don't think it does with the current implementation.

internal/ldcache/ldcache.go
elezar: The original implementation was mixing some of the concepts from finding libraries with just reading the ldcache. This has now been simplified so that this package only LOADS the cache and returns its contents as a list. The non-existence etc. is now handled higher up the call stack.

klueska: Ack

klueska (approved):
I didn't do a thorough code review of every line, but in general the new approach looks good and is in line with what we discussed.

PR #762, opened 2024-10-30
Fix bug when using just-in-time CDI spec generation

elezar:
This change fixes a bug when using just-in-time CDI spec generation for the NVIDIA Container Runtime for specific devices (i.e. not 'all').

Instead of unconditionally using the default nvsandboxutils library -- leading
to errors due to undefined symbols -- we check whether the library can be
properly initialised before continuing.

The functionality was added in https://github.com/NVIDIA/nvidia-container-toolkit/pull/629

Before this change:
```
$ docker run --rm -ti --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=0 ubuntu nvidia-smi -L
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: /home/elezar/src/container-toolkit/nvidia-container-runtime did not terminate successfully: exit status 127: /home/elezar/src/container-toolkit/nvidia-container-runtime: symbol lookup error: /home/elezar/src/container-toolkit/nvidia-container-runtime: undefined symbol: nvSandboxUtilsGetGpuResource
: unknown.
```

With this change:
```
$ docker run --rm -ti --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=0 ubuntu nvidia-smi -L
GPU 0: Tesla V100-SXM2-16GB-N (UUID: GPU-edfee158-11c1-52b8-0517-92f30e7fac88)
```

### DIFF
--- a/pkg/nvcdi/lib-nvml.go
+++ b/pkg/nvcdi/lib-nvml.go
@@ -103,7 +103,7 @@ func (l *nvmllib) GetDeviceSpecsByID(ids ...string) ([]specs.Device, error) {
 	return l.GetDeviceSpecsBy(identifiers...)
 }
 
-// GetDeviceSpecsBy is not supported for the gdslib specs.
+// GetDeviceSpecsBy returns the device specs for devices with the specified identifiers.
 func (l *nvmllib) GetDeviceSpecsBy(identifiers ...device.Identifier) ([]specs.Device, error) {
 	for _, id := range identifiers {
 		if id == "all" {
@@ -118,9 +118,22 @@ func (l *nvmllib) GetDeviceSpecsBy(identifiers ...device.Identifier) ([]specs.De
 	}
 	defer func() {
 		if r := l.nvmllib.Shutdown(); r != nvml.SUCCESS {
-			l.logger.Warningf("failed to shutdown NVML: %w", r)
+			l.logger.Warningf("failed to shutdown NVML: %v", r)
 		}
 	}()
+
+	if l.nvsandboxutilslib != nil {
+		if r := l.nvsandboxutilslib.Init(l.driverRoot); r != nvsandboxutils.SUCCESS {
+			l.logger.Warningf("Failed to init nvsandboxutils: %v; ignoring", r)
+			l.nvsandboxutilslib = nil
+		}
+		defer func() {
+			if l.nvsandboxutilslib == nil {
+				return
+			}
+			_ = l.nvsandboxutilslib.Shutdown()
+		}()
+	}
 
 	nvmlDevices, err := l.getNVMLDevicesByID(identifiers...)
 	if err != nil {


pkg/nvcdi/lib-nvml.go
```diff
@@ -118,10 +118,23 @@ func (l *nvmllib) GetDeviceSpecsBy(identifiers ...device.Identifier) ([]specs.De
 	}
 	defer func() {
 		if r := l.nvmllib.Shutdown(); r != nvml.SUCCESS {
-			l.logger.Warningf("failed to shutdown NVML: %w", r)
+			l.logger.Warningf("failed to shutdown NVML: %v", r)
 		}
 	}()
 
+	if l.nvsandboxutilslib != nil {
```
elezar, line 125: For the time being, I have duplicated the code for `GetAllDeviceSpecs` -- which is exercised more frequently -- and will clean the up more in a post-release follow-up.

ArangoGutierrez: Would it be worth to create a func now that we have the same code in 2 funcs?

pkg/nvcdi/lib-nvml.go
```diff
@@ -118,10 +118,23 @@ func (l *nvmllib) GetDeviceSpecsBy(identifiers ...device.Identifier) ([]specs.De
...
+		if r := l.nvsandboxutilslib.Init(l.driverRoot); r != nvsandboxutils.SUCCESS {
+			l.logger.Warningf("Failed to init nvsandboxutils: %v; ignoring", r)
```
ArangoGutierrez, line 127: Warning log on line 121 is lowercase and here is upper cased, is there a reason for the diff?

ArangoGutierrez (commented):
My questions are non blockers, looks good

ArangoGutierrez:
Tested 

```bash
$ docker run --rm -ti --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=0 ubuntu nvidia-smi -L
GPU 0: NVIDIA GH200 96GB HBM3 (UUID: GPU-0abb8bdb-b6c4-73fc-2b1d-a52db1c353e2)
```

patch restores the previous (1.16.2) behavior

pkg/nvcdi/lib-nvml.go
elezar: No specific reason. I will update these for consistency once we refactor.

pkg/nvcdi/lib-nvml.go
elezar: I don't know whether a functon would work since we need to defer the calls to `Shutdown()`. It is something we could look into, but I don't want to make that change this close to the release.

ArangoGutierrez: Agree, something to workshop for the next release. The patch does as described.

ArangoGutierrez (approved):

PR #763, opened 2024-10-30
Add support for specifying the config file path in an environment variable

elezar:
This change allows the config file used by the `nvidia-container-runtime` and `nvidia-container-runtime-hook` to be explicitly set using an `NVIDIA_CTK_CONFIG_FILE_PATH` environment variable. This removes the need to inject a `-config` option when installing the `nvidia-contianer-runtime-hook` wrapper from the toolkit container and allows the two components to treat this consistently.

This would allow the changes of #700 to be integrated more easily.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -32,9 +32,9 @@ import (
 
 const (
 	FilePathOverrideEnvVar = "NVIDIA_CTK_CONFIG_FILE_PATH"
+	RelativeFilePath       = "nvidia-container-runtime/config.toml"
 
-	configOverride = "XDG_CONFIG_HOME"
-	configFilePath = "nvidia-container-runtime/config.toml"
+	configRootOverride = "XDG_CONFIG_HOME"
 
 	nvidiaCTKExecutable          = "nvidia-ctk"
 	nvidiaCTKDefaultFilePath     = "/usr/bin/nvidia-ctk"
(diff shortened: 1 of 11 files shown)


internal/config/config.go
```diff
@@ -74,11 +76,15 @@ type Config struct {
 
 // GetConfigFilePath returns the path to the config file for the configured system
 func GetConfigFilePath() string {
-	if XDGConfigDir := os.Getenv(configOverride); len(XDGConfigDir) != 0 {
-		return filepath.Join(XDGConfigDir, configFilePath)
+	if configFilePathOverride := os.Getenv(FilePathOverrideEnvVar); configFilePathOverride != "" {
+		return configFilePathOverride
+	}
+	configRoot := "/etc"
```
ArangoGutierrez, line 82: nit: maybe a newline before `configRoot :=...`

ArangoGutierrez:
@elezar I think the PR title should be updated to better reflect the intent of the PR. 
`Allow config override by envvar` kind of tells me that I can override the config itself (like things inside the config), but the PR is just for the config PATH.

elezar:
@ArangoGutierrez is the change you require the PR title, or are you blocking on the nit?

ArangoGutierrez (approved):
LGTM

internal/config/config.go
ArangoGutierrez: non blocker

PR #764, opened 2024-10-30
Add feature to require kernel module

elezar:
This change adds a `require-nvidia-kernel-modules` feature flag that allows the logic for checking whether the `nvidia` kernel modules are loaded to be incorportated into the runtime itself.

This has the following advantages:
1. It allows support for platforms that don't have a shell (see #700)
2. It allows low-level runtimes other than `runc` to be used.


rajatchopra:
Closing out this PR.

Reasoning: While this feature is still useful to ensure the modified spec is called with the runtime only when environment is conducive, this path is less prominent with the advent of CDI.

Please re-open if anyone feels otherwise. The base code has been refactored quite a bit since the original PR also, so the work likely needs to be redone.

PR #765, opened 2024-10-30
Simplify standalone installer

elezar:
This PR starts refactoring the toolkit installation process from the toolkit contianer.

The changes are non-functional and limited to:
* Removing unused code
* Using a logger.Interface for logging instead of the top-level logger

The goal is to gradually move towards allowing #602 to be used for a single container image.


elezar:
I've pulled out the addition of tests to #768. I would rather merge those first.

cmd/nvidia-ctk-installer/container/toolkit/options.go
```diff
@@ -0,0 +1,40 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package toolkit
+
+import "github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+
+// An Option provides a mechanism to configure an Installer.
```
ArangoGutierrez, line 21:
```suggestion
// Option provides a mechanism to configure an Installer.
```

cmd/nvidia-ctk-installer/main.go
```diff
@@ -51,12 +51,46 @@ func (o options) toolkitRoot() string {
...
+	logger.Infof("Starting %v", c.Name)
+	if err := c.Run(remainingArgs); err != nil {
+		logger.Errorf("error running nvidia-toolkit: %v", err)
+		os.Exit(1)
+	}
+
+	logger.Infof("Completed %v", c.Name)
+}
+
+// An app represents the nvidia-ctk-installer.
```
ArangoGutierrez, line 74:
```suggestion
// app represents the nvidia-ctk-installer.
```

ArangoGutierrez (commented):
just 2 nits, not blockers. 
It would be good to add to the README a small section on how to use the installer from a container using `docker`/`podman` run

cmd/nvidia-ctk-installer/container/toolkit/options.go
elezar: Why this change? For nouns, preceding with `A` or `An` is accepted.

ArangoGutierrez: Yeah, that why I said is not a blocker, but we have other vars that are documented without the `A` or `An` so is more to be aligned

elezar: Local to this file there are no other *types* that require an `A` or an `An` to read more clearly. I agree that we have others in the project where this is not the case, but updating those is outside of the scope of this PR.

ArangoGutierrez (approved):

PR #768, opened 2024-11-01
[no-relnote] Add toolkit install unit test

elezar:
This change adds basic toolkit installation unit tests. This required that the source for files be specified when installing to allow for a testdata folder to be used.

This replaces the currently unused shell-based tests in /test/container.


tools/container/toolkit/toolkit.go
```diff
@@ -703,6 +703,7 @@ func findLibrary(root string, libName string) (string, error) {
 		}
 
 		return libraryCandidate, nil
+		// return strings.TrimPrefix(libraryCandidate, root), nil
```
tariq1890, line 706: Can we remove this?

tools/container/toolkit/toolkit.go
```diff
@@ -286,7 +286,7 @@ func TryDelete(cli *cli.Context, toolkitRoot string) error {
 
 // Install installs the components of the NVIDIA container toolkit.
 // Any existing installation is removed.
-func Install(cli *cli.Context, opts *Options, toolkitRoot string) error {
+func Install(cli *cli.Context, opts *Options, sourceRoot string, toolkitRoot string) error {
```
tariq1890, line 289: Can we add a small note stating the purpose of sourceRoot here?

Nice-to-have: 
```suggestion
func Install(cli *cli.Context,  sourceRoot string, toolkitRoot string, opts *Options,) error {
```

tariq1890: A **nice-to-have**
```suggestion
func Install(cli *cli.Context,  sourceRoot string, toolkitRoot string, opts *Options,) error {
```

tools/container/toolkit/toolkit.go
elezar: Done.

tools/container/toolkit/toolkit.go
elezar: > Can we add a small note stating the purpose of sourceRoot here? We could also mention what the toolkitRoot is for as well

I've added a comment on the purpose of `sourceRoot`. I would rather keep the `sourceRoot` after the options here. I have some follow-up commits (see https://github.com/NVIDIA/nvidia-container-toolkit/pull/765/commits/830904f950bef183242c6e1051a81c701e77d11e) that further refactors how these arguments are specified, and would rahter address this there.

tariq1890 (approved):

Issue #772, opened 2024-11-04
[v1.17.0, mode = "cdi"] OCI runtime create failed: 'failed to create link'/'failed to check if link exists: unexpected link target'


benz0li:
Full error message:

```
failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #1: error running hook: exit status 1, stdout: , stderr: time="yyyy-mm-ddThh:mm:ss+hh:mm" level=error msg="failed to create link [libGLX_nvidia.so.535.183.01 /usr/lib/x86_64-linux-gnu/libGLX_indirect.so.0]: failed to check if link exists: unexpected link target: libGLX_mesa.so.0": unknown"
```

Further information:

```
ls -al /usr/lib/x86_64-linux-gnu
```

```
...
lrwxrwxrwx  1 root root        27 month  dd hh:mm libGLX_indirect.so.0 -> libGLX_nvidia.so.535.183.01
lrwxrwxrwx  1 root root        27 month  dd hh:mm libGLX_nvidia.so.0 -> libGLX_nvidia.so.535.183.01
-rwxr-xr-x  1 root root   1195552 month  dd hh:mm libGLX_nvidia.so.535.183.01
...
```

elezar:
@benz0li could you provide information on the image that you're running? 

I would assume that a symlink `libGLX_indirect.so.0 -> libGLX_mesa.so.0` already exists in the image in this case.

SurenNihalani:
This happened to us as well. I think we should rollback this image to avoid more people from hitting this

benz0li:
> [@benz0li](https://github.com/benz0li) could you provide information on the image that you're running?

Image: `glcr.b-data.ch/jupyterlab/cuda/r/verse:4.4.2`

Source: https://github.com/b-data/jupyterlab-r-docker-stack/blob/4786ca20927a2d34a8d4dcdb2f2b4fdfcc55660e/verse/latest.Dockerfile

> I would assume that a symlink `libGLX_indirect.so.0 -> libGLX_mesa.so.0` already exists in the image in this case.

Yes. Due to the installation of [`libgl1-mesa-dev`](https://packages.ubuntu.com/jammy-updates/libgl1-mesa-dev) and thus [`libglx-mesa0`](https://packages.ubuntu.com/jammy/libglx-mesa0).

[...]

PR #773, opened 2024-11-04
Force creation of symlinks in create-symlink hook

elezar:
This change updates the create-symlink hook to be equivalent to
```
ln -f -s target link
```
This ensures that links are updated even if they exist in the container being run.


klueska:
What if the symlink is wrong?

elezar:
> What if the existing symlink is wrong?

That's exactly the issue in #772. I have updated the implementation here to force create a symlink, but I think we should talk through the implications.

cdesiniotis (approved):

PR #777, opened 2024-11-07
Fallback to file for runtime config

elezar:
This change ensures that we fall back to the previous behaviour of reading the existing config from the specified config file if extracting the current config from the command line fails. This fixes use cases where the containerd / crio executables are not available.

### DIFF
--- a/tools/container/runtime/containerd/containerd.go
+++ b/tools/container/runtime/containerd/containerd.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/NVIDIA/nvidia-container-toolkit/pkg/config/engine"
 	"github.com/NVIDIA/nvidia-container-toolkit/pkg/config/engine/containerd"
+	"github.com/NVIDIA/nvidia-container-toolkit/pkg/config/toml"
 	"github.com/NVIDIA/nvidia-container-toolkit/tools/container"
 )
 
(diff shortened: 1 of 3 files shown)


elezar:
> I think we need the same fallback logic [here](https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/tools/container/runtime/runtime.go#L170) as well

You're right. I have updated the implementation. I also missed the logic in the `Setup` calls.

tools/container/runtime/containerd/containerd.go
```diff
@@ -84,13 +85,7 @@ func Flags(opts *Options) []cli.Flag {
...
 	log.Infof("Starting 'setup' for %v", c.App.Name)
 
-	cfg, err := containerd.New(
-		containerd.WithPath(o.Config),
-		containerd.WithConfigSource(containerd.CommandLineSource(o.HostRootMount)),
-		containerd.WithRuntimeType(co.runtimeType),
-		containerd.WithUseLegacyConfig(co.useLegacyConfig),
-		containerd.WithContainerAnnotations(co.containerAnnotationsFromCDIPrefixes()...),
-	)
+	cfg, err := getConfigurer(o, co)
```
tariq1890, line 88: nit: can we choose a name that is more specific here? `getConfigurer` sounds a bit too vague to me

tariq1890 (approved):

tools/container/runtime/containerd/containerd.go
```diff
@@ -84,13 +85,7 @@ func Flags(opts *Options) []cli.Flag {
...
 	log.Infof("Starting 'setup' for %v", c.App.Name)
 
-	cfg, err := containerd.New(
-		containerd.WithPath(o.Config),
-		containerd.WithConfigSource(containerd.CommandLineSource(o.HostRootMount)),
-		containerd.WithRuntimeType(co.runtimeType),
-		containerd.WithUseLegacyConfig(co.useLegacyConfig),
-		containerd.WithContainerAnnotations(co.containerAnnotationsFromCDIPrefixes()...),
-	)
+	cfg, err := getConfig(o, co)
```
tariq1890, line 88:
```suggestion
	cfg, err := getRuntimeConfig(o, co)
```

PR #778, opened 2024-11-08
Fix fsnotify.Remove logic function.

zhanluxianshen:
1 Use simple grammar in fsnotify package.
```
This makes checking events a lot easier; for example:

  if event.Op&Write == Write && !(event.Op&Remove == Remove) {
  }
Becomes:

  if event.Has(Write) && !event.Has(Remove) {
  }
``` 

 According [fsnotify notes](https://github.com/fsnotify/fsnotify/releases/tag/v1.6.0)

2.
Fix Remove event.


tariq1890 (approved):

elezar (approved):

zhanluxianshen:
thanks.

PR #782, opened 2024-11-08
Fix logger error append logic.

zhanluxianshen:
Fix logger error append logic.

### DIFF
--- a/internal/runtime/logger.go
+++ b/internal/runtime/logger.go
@@ -74,7 +74,9 @@ func (l *Logger) Update(filename string, logLevel string, argv []string) {
 		if argLogFile != nil {
 			logFiles = append(logFiles, argLogFile)
 		}
-		argLogFileError = errors.Join(argLogFileError, err)
+		if err != nil {
+			argLogFileError = errors.Join(argLogFileError, err)
+		}
 	}
 	defer func() {
 		if argLogFileError != nil {


internal/runtime/logger.go
```diff
@@ -74,7 +74,9 @@ func (l *Logger) Update(filename string, logLevel string, argv []string) {
 		if argLogFile != nil {
 			logFiles = append(logFiles, argLogFile)
 		}
-		argLogFileError = errors.Join(argLogFileError, err)
+		if err != nil {
+			argLogFileError = errors.Join(argLogFileError, err)
+		}
```
elezar, line 79: Could you explain why this is required? `errors.Join` should ignore `nil` errors.

elezar:
```// Join returns an error that wraps the given errors.
// Any nil error values are discarded.
// Join returns nil if every value in errs is nil.
// The error formats as the concatenation of the strings obtained
// by calling the Error method of each element of errs, with a newline
// between each string.
//
// A non-nil error returned by Join implements the Unwrap() []error method.
func Join(errs ...error) error {
	n := 0
	for _, err := range errs {
		if err != nil {
			n++
		}
	}
	if n == 0 {
		return nil
	}
	e := &joinError{
		errs: make([]error, 0, n),
	}
	for _, err := range errs {
		if err != nil {
			e.errs = append(e.errs, err)
		}
	}
	return e
}
```

zhanluxianshen:
Thanks for quick review.
you are right , its no need to check nil.

Issue #797, opened 2024-11-14
Parsing default IMEX info fails for legacy images


astefanutti:
Since the latest 1.17.x versions, containers with images considered "legacy" and that do not have the `NVIDIA_IMEX_CHANNELS` environment variable set fail to start with the following error:

```
Error: container create failed: time="2024-11-13T16:24:41Z" level=error msg="runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli.real: error parsing IMEX info: unsupported IMEX channel value: all\n" 
```

It seems the `NVIDIA_IMEX_CHANNELS` environment variable is defaulted to `all` here for "legacy" images:

https://github.com/NVIDIA/nvidia-container-toolkit/blob/1995925a7df644ead7afb767608841d9a08bcbc4/internal/config/image/cuda_image.go#L145

Which cannot be parsed by https://github.com/NVIDIA/libnvidia-container/blob/63d366ee3b4183513c310ac557bf31b05b83328f/src/cli/common.c#L446.

An occurrence of that issue has been reported here for example: https://github.com/pytorch/test-infra/pull/5852.

That case should ideally be more gracefully handled.

higi:
can anyone help me with this?:

2024-11-15T11:55:12Z create container gshaibi/gpu-burn:latest
2024-11-15T11:55:13Z latest Pulling from gshaibi/gpu-burn
2024-11-15T11:55:13Z Digest: sha256:ed07993b0581228c2bd7113fae0ed214549547f0fa91ba50165bc2473cfaf979
2024-11-15T11:55:13Z Status: Image is up to date for gshaibi/gpu-burn:latest
2024-11-15T11:55:14Z start container for gshaibi/gpu-burn:latest: begin
2024-11-15T11:55:14Z error starting container: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: Auto-detected mode as 'legacy'
nvidia-container-cli: error parsing IMEX info: unsupported IMEX channel value: all: unknown
2024-11-15T11:55:30Z start container for gshaibi/gpu-burn:latest: begin
[...]

PR #799, opened 2024-11-14
Fix NVIDIA_IMEX_CHANNELS handling on legacy images

elezar:
For legacy images (images with a CUDA_VERSION set but no CUDA_REQUIRES set), the default behaviour for device envvars is to treat non-existence as all.

This change ensures that the NVIDIA_IMEX_CHANNELS envvar is not treated in the same way, instead returning no devices if the envvar is not set.

See #797


klueska (commented):
It's a little weird that unset == all, but I see how this change works, and its fine for now.

elezar:
> It's a little weird that unset == all, but I see how this change works, and its fine for now.

Note that this is for legacy images only. We could look into deprecating support for those to simplify our reasoning about these cases.

klueska (approved):

astefanutti:
@elezar @klueska thank you for the quick fix!

PR #805, opened 2024-11-18
Add support for containerd v3 configs

alam0rt:
With the release of containerd 2.0 comes a new default config version. This PR attempts to support the new version 3 of the containerd configuration.

Version 3 is fairly similar to 2 and for all intents and purposes, a valid version 3 config is a valid version 2 config (bar the version of course). So for now I've just gone and copied the V2 logic and changed things here and there.

### DIFF
--- a/pkg/config/engine/containerd/config_v1.go
+++ b/pkg/config/engine/containerd/config_v1.go
@@ -70,18 +70,20 @@ func (c *ConfigV1) AddRuntime(name string, path string, setAsDefault bool) error
 	config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "options", "BinaryName"}, path)
 	config.SetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "options", "Runtime"}, path)
 
-	if setAsDefault && c.UseDefaultRuntimeName {
-		config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime_name"}, name)
-	} else if setAsDefault {
-		// Note: This is deprecated in containerd 1.4.0 and will be removed in 1.5.0
-		if config.GetPath([]string{"plugins", "cri", "containerd", "default_runtime"}) == nil {
-			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "runtime_type"}, c.RuntimeType)
-			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "runtime_root"}, "")
-			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "runtime_engine"}, "")
-			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "privileged_without_host_devices"}, false)
+	if setAsDefault {
+		if !c.UseLegacyConfig {
+			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime_name"}, name)
+		} else {
+			// Note: This is deprecated in containerd 1.4.0 and will be removed in 1.5.0
+			if config.GetPath([]string{"plugins", "cri", "containerd", "default_runtime"}) == nil {
+				config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "runtime_type"}, c.RuntimeType)
+				config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "runtime_root"}, "")
+				config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "runtime_engine"}, "")
+				config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "privileged_without_host_devices"}, false)
+			}
+			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "options", "BinaryName"}, path)
+			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "options", "Runtime"}, path)
 		}
-		config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "options", "BinaryName"}, path)
-		config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "options", "Runtime"}, path)
 	}
 
 	*c.Tree = config
--- a/pkg/config/engine/containerd/config_v2_test.go
+++ b/pkg/config/engine/containerd/config_v2_test.go
@@ -46,7 +46,7 @@ func TestAddRuntime(t *testing.T) {
 					privileged_without_host_devices = false
 					runtime_engine = ""
 					runtime_root = ""
-					runtime_type = ""
+					runtime_type = "io.containerd.runc.v2"
 					[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.test.options]
 						BinaryName = "/usr/bin/test"
 			`,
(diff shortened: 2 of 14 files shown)


pkg/config/engine/containerd/config_v3.go
```diff
@@ -0,0 +1,159 @@
...
+
+import (
+	"fmt"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/pkg/config/engine"
+	"github.com/NVIDIA/nvidia-container-toolkit/pkg/config/toml"
+)
+
+// ConfigV3 represents a version 3 containerd config
+type ConfigV3 Config
```
elezar, line 27: If the v3 config is the same as the v2 config then we should not duplicate the logic for the functions here.

Are there any differences that are relevant to the runtime sections?

pkg/config/engine/containerd/containerd.go
```diff
@@ -89,14 +91,16 @@ func New(opts ...Option) (engine.Interface, error) {
...
 		return cfg, nil
+	case 3:
+		return (*ConfigV3)(cfg), nil
 	}
 
 	return nil, fmt.Errorf("unsupported config version: %v", version)
 }
 
 // parseVersion returns the version of the config
 func (c *Config) parseVersion(useLegacyConfig bool) (int, error) {
```
elezar, line 102: We've essentially replaced the `useLegacyConfig` variable with the `configVersion` option. Would it make sense to update the implementation here to something like:

```
defaultVersion := 3
if c.configVersion != 0 {
    defaultVersion = c.configVersion
}
```
and drop the `useLegacyConfig` argument?

pkg/config/engine/containerd/option.go
```diff
@@ -30,7 +30,7 @@ type builder struct {
 	configSource         toml.Loader
 	path                 string
 	runtimeType          string
-	useLegacyConfig      bool
+	configVersion        int64
```
elezar, line 33: This is technically the `defaultVersion` since the version in the config file overwrites this.

elezar:
@cdesiniotis @tariq1890 this is something that will be required for the next GPU Operator release and we could consider backporting this too.

alam0rt:
> @alam0rt I have taken these changes and reworked them. I hope you don't have an issue with me rebasing your commits to get them a little more streamlined. Let me know what you think.

I absolutely do not mind. Thanks for taking a look! Have been a bit inundated so I haven't had a chance to rework it myself

cdesiniotis:
It looks like `io.containerd.grpc.v1.cri` was renamed to `io.containerd.cri.v1.runtime` in the v3 configuration file. I was able to get things to work with this commit: https://github.com/cdesiniotis/nvidia-container-toolkit/commit/16b5376a907c35a32f012afd4ba371b327a70fc6.

cdesiniotis:
@alam0rt I have gone ahead and updated this PR with my commit.

@elezar @tariq1890 please take a look.

pkg/config/engine/containerd/config_v1.go
```diff
@@ -70,18 +70,20 @@ func (c *ConfigV1) AddRuntime(name string, path string, setAsDefault bool) error
...
-		config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime_name"}, name)
-	} else if setAsDefault {
-		// Note: This is deprecated in containerd 1.4.0 and will be removed in 1.5.0
-		if config.GetPath([]string{"plugins", "cri", "containerd", "default_runtime"}) == nil {
-			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "runtime_type"}, c.RuntimeType)
-			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "runtime_root"}, "")
-			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "runtime_engine"}, "")
-			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "privileged_without_host_devices"}, false)
+	if setAsDefault {
+		if !c.UseLegacyConfig {
```
tariq1890, line 74: Can we invert the if-else block here? It would look more readable if we avoid the NOT statement when possible

pkg/config/engine/api.go
```diff
@@ -18,12 +18,13 @@ package engine
...
-	DefaultRuntime() string
 	AddRuntime(string, string, bool) error
-	Set(string, interface{})
+	DefaultRuntime() string
+	GetRuntimeConfig(string) (RuntimeConfig, error)
 	RemoveRuntime(string) error
 	Save(string) (int64, error)
-	GetRuntimeConfig(string) (RuntimeConfig, error)
+	Set(string, interface{})
+	String() string
```
tariq1890, line 27: I see that we are only using this in unit tests, do we really need to add a new interface method for this?

pkg/config/engine/containerd/config_v1.go
alam0rt: inverted it

pkg/config/engine/containerd/containerd.go
```diff
@@ -72,45 +88,50 @@ func New(opts ...Option) (engine.Interface, error) {
...
+		UseLegacyConfig:      b.useLegacyConfig,
+		ContainerAnnotations: b.containerAnnotations,
 	}
-	switch version {
+
+	switch configVersion {
 	case 1:
 		return (*ConfigV1)(cfg), nil
 	case 2:
+		cfg.CRIRuntimePluginName = "io.containerd.grpc.v1.cri"
```
elezar, line 109: The intent of this switch statement is just to cast the config to the correct type, so instead of doing this here, should we set some local based on the version above and set `CRIRuntimePluginName` in the instantiation directly?

I think we could add `b.criRuntimePluginName(version)` or something to make this cleaner.

pkg/config/engine/containerd/option.go
```diff
@@ -21,16 +21,13 @@ import (
...
 )
 
-const (
-	defaultRuntimeType = "io.containerd.runc.v2"
-)
-
 type builder struct {
 	logger               logger.Interface
 	configSource         toml.Loader
+	configVersion        int
```
elezar, line 27: Did we want to allow the `CRIRuntimePluginName` to be overridden?

pkg/config/engine/containerd/config_v1.go
elezar: @tariq1890 I kept the ordering of the blocks the same to reduce the diff against `main`. @alam0rt this means that I reverted your commit that inverts the block (which was not signed-off in any case).

pkg/config/engine/containerd/option.go
elezar: I didn't update this to allow it to be overridden. If we see this as a requirement, plumbing a value through should be relatively straightforward.

pkg/config/engine/containerd/config_v2_test.go
cdesiniotis, line 1: nit: should we rename this file to `config_test.go` to align with the renaming of `config_v2.go` to `config.go`?

cdesiniotis (approved):

pkg/config/engine/containerd/config_v2_test.go
elezar: Yes, that's a good idea. Updated.

pkg/config/engine/api.go
elezar: Sorry @tariq1890, I missed this. I will have a look again and see if we can clean up the interface as a follow-up.

PR #807, opened 2024-11-21
Add imex mode to CDI spec generation

elezar:
This change adds a imex mode to CDI spec generation. This mode detected generates CDI specifications for existing IMEX channels. By default these devices have the fully qualified CDI device names:

`nvidia.com/imex-channel={{ .ID }}`

### DIFF
--- a/cmd/nvidia-ctk/cdi/generate/generate.go
+++ b/cmd/nvidia-ctk/cdi/generate/generate.go
@@ -104,10 +104,12 @@ func (m command) build() *cli.Command {
 			Destination: &opts.format,
 		},
 		&cli.StringFlag{
-			Name:        "mode",
-			Aliases:     []string{"discovery-mode"},
-			Usage:       "The mode to use when discovering the available entities. One of [auto | nvml | wsl]. If mode is set to 'auto' the mode will be determined based on the system configuration.",
-			Value:       nvcdi.ModeAuto,
+			Name:    "mode",
+			Aliases: []string{"discovery-mode"},
+			Usage: "The mode to use when discovering the available entities. " +
+				"One of [" + strings.Join(nvcdi.AllModes[string](), " | ") + "]. " +
+				"If mode is set to 'auto' the mode will be determined based on the system configuration.",
+			Value:       string(nvcdi.ModeAuto),
 			Destination: &opts.mode,
 		},
 		&cli.StringFlag{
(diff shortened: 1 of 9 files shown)


pkg/nvcdi/lib-imex.go
```diff
@@ -0,0 +1,118 @@
...
+
+var _ Interface = (*imexlib)(nil)
+
+const (
+	classImexChannel = "imex-channel"
+)
+
+// GetSpec returns a CDI spec for all available IMEX channels.
+func (l *imexlib) GetSpec() (spec.Interface, error) {
+	return nil, nil
```
cdesiniotis, line 45: Why is this not implemented?

pkg/nvcdi/lib-imex_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	moduleRoot, err := test.GetModuleRoot()
+	require.NoError(t, err)
+	hostRoot := filepath.Join(moduleRoot, "testdata", "lookup", "rootfs-1")
+
+	expectedSpec := `---
+cdiVersion: 0.5.0
+containerEdits:
+  env:
+  - NVIDIA_VISIBLE_DEVICES=void
```
cdesiniotis, line 44: Question -- where in the code is this set? Are we always adding this envvar to our specs regardless of the mode (i.e. nvml, wsl, imex)?

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -183,17 +183,6 @@ func (m command) validateFlags(c *cli.Context, opts *options) error {
 		return fmt.Errorf("invalid output format: %v", opts.format)
 	}
 
-	opts.mode = strings.ToLower(opts.mode)
```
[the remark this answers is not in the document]
elezar: This was removed because I didn't want to maintain the list in two places. Let me pull in some changes that I have in another commit to improve things

pkg/nvcdi/lib-imex.go
elezar: We actually never call GetSpec at this level and always call GetSpec of the wrapper. Definitely needs a cleanup, but that refactor is out of scope for this PR.

pkg/nvcdi/lib-imex_test.go
elezar: This is set in the wrapper GetSpec in lib.go. I think these changes call out that this may not be the best place to do it since it seems a little out of place. The intent is to ENSURE that we don't inadvertently trigger runtime or hook injection when we're already applying changes using CDI.

cdesiniotis (approved):

PR #825, opened 2024-12-06
Refactor driver discovery

elezar:


internal/nvsandboxutils/gen/nvsandboxutils/nvsandboxutils.yml
```diff
@@ -49,6 +49,7 @@ TRANSLATOR:
     const:
       - {action: accept, from: "^NVSANDBOXUTILS_"}
       - {action: accept, from: "^nvSandboxUtils"}
+      - {action: replace, from: "^NVSANDBOXUTILS_255_MASK_", to: "MASK255_" }
```
elezar, line 52: I don't think this was required.

PR #826, opened 2024-12-06
Fix create-device-node test when devices exist

elezar:
This changes fixes the TestCreateControlDevices test on systems where device nodes exist.


elezar:
cc @xiaoyaoyuyi

PR #838, opened 2024-12-18
Add option in toolkit container to enable CDI in runtime

cdesiniotis:
These changes add an option to enable CDI in container engines (containerd, crio, docker) from the toolkit contaienr.

This can be triggered through the `--enable-cdi-in-runtime` command line option or `RUNTIME_ENABLE_CDI` environment variable.

### DIFF
--- a/cmd/nvidia-ctk-installer/container/runtime/runtime.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/runtime.go
@@ -25,6 +25,7 @@ import (
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/container/runtime/containerd"
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/container/runtime/crio"
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/container/runtime/docker"
+	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/container/toolkit"
 )
 
 const (
@@ -104,9 +105,13 @@ func Flags(opts *Options) []cli.Flag {
 }
 
 // ValidateOptions checks whether the specified options are valid
-func ValidateOptions(opts *Options, runtime string, toolkitRoot string) error {
+func ValidateOptions(c *cli.Context, opts *Options, runtime string, toolkitRoot string, to *toolkit.Options) error {
 	// We set this option here to ensure that it is available in future calls.
 	opts.RuntimeDir = toolkitRoot
+
+	if !c.IsSet("enable-cdi-in-runtime") {
+		opts.EnableCDI = to.CDI.Enabled
+	}
 
 	// Apply the runtime-specific config changes.
 	switch runtime {
--- a/cmd/nvidia-ctk-installer/container/runtime/runtime.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/runtime.go
@@ -66,6 +66,12 @@ func Flags(opts *Options) []cli.Flag {
 			Destination: &opts.RestartMode,
 			EnvVars:     []string{"RUNTIME_RESTART_MODE"},
 		},
+		&cli.BoolFlag{
+			Name:        "enable-cdi-in-runtime",
+			Usage:       "Enable CDI in the configured runt	ime",
+			Destination: &opts.EnableCDI,
+			EnvVars:     []string{"RUNTIME_ENABLE_CDI"},
+		},
 		&cli.StringFlag{
 			Name:        "host-root",
 			Usage:       "Specify the path to the host root to be used when restarting the runtime using systemd",
--- a/pkg/config/engine/api.go
+++ b/pkg/config/engine/api.go
@@ -24,7 +24,6 @@ type Interface interface {
 	GetRuntimeConfig(string) (RuntimeConfig, error)
 	RemoveRuntime(string) error
 	Save(string) (int64, error)
-	Set(string, interface{})
 	String() string
 }
 
--- a/pkg/config/engine/api.go
+++ b/pkg/config/engine/api.go
@@ -20,6 +20,7 @@ package engine
 type Interface interface {
 	AddRuntime(string, string, bool) error
 	DefaultRuntime() string
+	EnableCDI()
 	GetRuntimeConfig(string) (RuntimeConfig, error)
 	RemoveRuntime(string) error
 	Save(string) (int64, error)
--- a/pkg/config/engine/containerd/config.go
+++ b/pkg/config/engine/containerd/config.go
@@ -111,6 +111,11 @@ func (c Config) DefaultRuntime() string {
 	return ""
 }
 
+// EnableCDI sets the enable_cdi field in the Containerd config to true.
+func (c *Config) EnableCDI() {
+	c.Set("enable_cdi", true)
+}
+
 // RemoveRuntime removes a runtime from the docker config
 func (c *Config) RemoveRuntime(name string) error {
 	if c == nil || c.Tree == nil {
--- a/pkg/config/engine/containerd/config.go
+++ b/pkg/config/engine/containerd/config.go
@@ -96,13 +96,6 @@ func (c *Config) getRuntimeAnnotations(path []string) ([]string, error) {
 	return annotations, nil
 }
 
-// Set sets the specified containerd option.
-func (c *Config) Set(key string, value interface{}) {
-	config := *c.Tree
-	config.SetPath([]string{"plugins", c.CRIRuntimePluginName, key}, value)
-	*c.Tree = config
-}
-
 // DefaultRuntime returns the default runtime for the cri-o config
 func (c Config) DefaultRuntime() string {
 	if runtime, ok := c.GetPath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"}).(string); ok {
@@ -113,7 +106,9 @@ func (c Config) DefaultRuntime() string {
 
 // EnableCDI sets the enable_cdi field in the Containerd config to true.
 func (c *Config) EnableCDI() {
-	c.Set("enable_cdi", true)
+	config := *c.Tree
+	config.SetPath([]string{"plugins", c.CRIRuntimePluginName, "enable_cdi"}, true)
+	*c.Tree = config
 }
 
 // RemoveRuntime removes a runtime from the docker config
--- a/pkg/config/engine/containerd/config_v1.go
+++ b/pkg/config/engine/containerd/config_v1.go
@@ -143,13 +143,6 @@ func (c *ConfigV1) RemoveRuntime(name string) error {
 	return nil
 }
 
-// Set sets the specified containerd option.
-func (c *ConfigV1) Set(key string, value interface{}) {
-	config := *c.Tree
-	config.SetPath([]string{"plugins", "cri", "containerd", key}, value)
-	*c.Tree = config
-}
-
 // Save writes the config to a file
 func (c ConfigV1) Save(path string) (int64, error) {
 	return (Config)(c).Save(path)
@@ -167,5 +160,7 @@ func (c *ConfigV1) GetRuntimeConfig(name string) (engine.RuntimeConfig, error) {
 }
 
 func (c *ConfigV1) EnableCDI() {
-	c.Set("enable_cdi", true)
+	config := *c.Tree
+	config.SetPath([]string{"plugins", "cri", "containerd", "enable_cdi"}, true)
+	*c.Tree = config
 }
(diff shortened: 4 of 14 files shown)


tools/container/runtime/runtime.go
```diff
@@ -89,6 +90,13 @@ func Flags(opts *Options) []cli.Flag {
 			EnvVars:     []string{"NVIDIA_RUNTIME_SET_AS_DEFAULT", "CONTAINERD_SET_AS_DEFAULT", "DOCKER_SET_AS_DEFAULT"},
 			Hidden:      true,
 		},
+		&cli.BoolFlag{
+			Name:        "runtime-enable-cdi",
```
tariq1890, line 94: Let the flag be passed as `--enable-cdi`. The envar and the Option variable can have the `Runtime/RUNTIME` prefix. This is consistent with the other flags like `socket`, `restart-mode` etc
```suggestion
			Name:        "enable-cdi",
```

tools/container/runtime/runtime.go
```diff
@@ -89,6 +90,13 @@ func Flags(opts *Options) []cli.Flag {
...
+			Usage:       "Enable CDI in the configured runtime",
```
tariq1890, line 95:
```suggestion
			Usage:       "Enable Container Device Interface (CDI) in the configured runtime",
```

pkg/config/engine/containerd/config_v1.go
```diff
@@ -163,3 +163,7 @@ func (c *ConfigV1) GetRuntimeConfig(name string) (engine.RuntimeConfig, error) {
 		tree: runtimeData,
 	}, nil
 }
+
+func (c *ConfigV1) EnableCDI() {
+	c.Set("enable_cdi", true)
```
elezar, line 168: I think we should be able to cast `c` to a `Config` type and call `EnableCDI` here instead of reimplementing it.

pkg/config/engine/api.go
```diff
@@ -24,6 +24,7 @@ type Interface interface {
 	RemoveRuntime(string) error
 	Save(string) (int64, error)
 	GetRuntimeConfig(string) (RuntimeConfig, error)
+	EnableCDI()
```
elezar, line 27: I think `Set` is only used to enable CDI. Let's remove that from the interface if that's the case.

tools/container/container.go
```diff
@@ -36,13 +36,14 @@ const (
...
-	RestartMode   string
-	HostRootMount string
+	Config           string
+	Socket           string
+	RuntimeName      string
+	RuntimeDir       string
+	SetAsDefault     bool
+	RestartMode      string
+	HostRootMount    string
+	RuntimeEnableCDI bool
```
elezar, line 46: I don't agree with the `Runtime*` prefix here. `Runtime` in this context refers to the NVIDIA runtime. I would just use `EnableCDI` here.

tools/container/runtime/runtime.go
elezar: Let's try to be consistent with the option for the `nvidia-ctk runtime configure` command.

elezar (dismissed):
Some minor comments.

tools/container/runtime/runtime.go
cdesiniotis: The problem is that we already have a `cdi-enabled` / `enable-cdi` flag defined in the toolkit options which triggers the generation of a CDI spec: https://github.com/NVIDIA/nvidia-container-toolkit/blob/8d869acce5643ed0de0a358f605bf14f8bc05fca/tools/container/toolkit/toolkit.go#L173-L179

Any ideas on what the name of the new flag should be?

elezar: Thanks for the clarification.

One question I would have is whether we need a separate flag? Would it not make sense for the same flag / envvar (e.g. `CDI_ENABLED`) to be use to also trigger enabling CDI in the runtime? Do older versions of containerd complain about invalid config options?

If we feel we want finer control, then I'm ok to use `RUNTIME_ENABLE_CDI` or `ENABLE_CDI_IN_RUNTIME` to be used. (maybe the latter).

tools/container/runtime/docker/docker.go
```diff
@@ -55,6 +55,10 @@ func Setup(c *cli.Context, o *container.Options) error {
 		return fmt.Errorf("unable to configure docker: %v", err)
 	}
 
+	if o.RuntimeEnableCDI {
+		cfg.Set("features", map[string]bool{"cdi": true})
+	}
+
```
elezar, line 61: Note that we're already enabling CDI in the `Configure` call, so there is no need to do it here too.
```suggestion
```

tools/container/runtime/runtime.go
```diff
@@ -124,6 +132,9 @@ func ValidateOptions(opts *Options, runtime string, toolkitRoot string) error {
 		if opts.RestartMode == runtimeSpecificDefault {
 			opts.RestartMode = crio.DefaultRestartMode
 		}
+		if opts.RuntimeEnableCDI {
+			opts.RuntimeEnableCDI = false
+		}
```
elezar, line 137: We already handle the case that `EnableCDI` is a no-op in the case of CRI-O.
```suggestion
```

pkg/config/engine/docker/docker.go
```diff
@@ -103,6 +103,11 @@ func (c Config) DefaultRuntime() string {
 	return r
 }
 
+// EnableCDI sets features.cdi to true in the docker config.
+func (c *Config) EnableCDI() {
+	c.Set("features", map[string]bool{"cdi": true})
```
elezar, line 108: Out of scope of this PR: I've just thought about the fact that we would overrwrite existing features here.

elezar:
@klueska as discussed today, I have rebased this PR.

elezar:
As discussed yesterday, we should have the `cdi.enabled = true` option in the GPU Operator CRD also trigger enabling CDI in runtimes where relevant (e.g. Containerd). In most cases this should not conflict with using the `nvidia-container-runtime.cdi` runtime due to the fact that:
* The `nvidia-container-runtime.cdi` is configured to react to the `nvidia.cdi.k8s.io/` annotation prefixes.
* The `k8s-device-plugin` uses the `nvidia.cdi.k8s.io/` annotation prefix.
* The `nvidia-container-runtime.cdi` is configured to use the `management.nvidia.com/gpu` CDI kind by default.

Note that the following edge case exists:
* If the `k8s-device-plugin` is configured to inject CDI devices using the CRI fields (not possible using the current GPU Operator API), certain kubelet versions will map these requests to `cdi.k8s.io/` annotations in addition to passing the requested CDI devices in the CRI field. This should still not be an issue unless a user has explicitly added `cdi.k8s.io/` annotation to the list of allowe annotations.

I think it should be sufficient to have `CDI_ENABLED` imply `RUNTIME_ENABLE_CDI` with an explicit setting of `RUNTIME_ENABLE_CDI` overriding the inferred value.

pkg/config/engine/containerd/config.go
```diff
@@ -111,6 +104,13 @@ func (c Config) DefaultRuntime() string {
 	return ""
 }
 
+// EnableCDI sets the enable_cdi field in the Containerd config to true.
+func (c *Config) EnableCDI() {
+	config := *c.Tree
+	config.SetPath([]string{"plugins", c.CRIRuntimePluginName, "enable_cdi"}, true)
```
guptaNswati, line 110: nitpick: to add an explicit log for this enablement. something like 
`c.Logger.Infof("enabled CDI in %v ", c.CRIRuntimePluginName)`

pkg/config/engine/containerd/config_v1.go
```diff
@@ -165,3 +158,9 @@ func (c *ConfigV1) GetRuntimeConfig(name string) (engine.RuntimeConfig, error) {
 		tree: runtimeData,
 	}, nil
 }
+
+func (c *ConfigV1) EnableCDI() {
+	config := *c.Tree
+	config.SetPath([]string{"plugins", "cri", "containerd", "enable_cdi"}, true)
```
guptaNswati, line 164: here too c.Logger.Infof("enabled CDI in cri and containerd")

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -66,6 +66,12 @@ func Flags(opts *Options) []cli.Flag {
 			Destination: &opts.RestartMode,
 			EnvVars:     []string{"RUNTIME_RESTART_MODE"},
 		},
+		&cli.BoolFlag{
+			Name:        "enable-cdi-in-runtime",
+			Usage:       "Enable CDI in the configured runt	ime",
+			Destination: &opts.EnableCDI,
+			EnvVars:     []string{"RUNTIME_ENABLE_CDI"},
```
cdesiniotis, line 73: nit: do we want to update the envvar to match the name of the flag, `ENABLE_CDI_IN_RUNTIME`? I don't have a strong opinion though since the other options all start with the `RUNTIME_` prefix.

pkg/config/engine/containerd/config.go
cdesiniotis: If we want to add a debug or log statement, it should be at the call site.

pkg/config/engine/containerd/config_v1.go
cdesiniotis: If we want to add a debug or log statement, it should be at the call site.

cdesiniotis:
> As discussed yesterday, we should have the `cdi.enabled = true` option in the GPU Operator CRD also trigger enabling CDI in runtimes where relevant (e.g. Containerd). In most cases this should not conflict with using the `nvidia-container-runtime.cdi` runtime due to the fact that:
> 
> * The `nvidia-container-runtime.cdi` is configured to react to the `nvidia.cdi.k8s.io/` annotation prefixes.
> * The `k8s-device-plugin` uses the `nvidia.cdi.k8s.io/` annotation prefix.
> * The `nvidia-container-runtime.cdi` is configured to use the `management.nvidia.com/gpu` CDI kind by default.
> 
> Note that the following edge case exists:
> 
> * If the `k8s-device-plugin` is configured to inject CDI devices using the CRI fields (not possible using the current GPU Operator API), certain kubelet versions will map these requests to `cdi.k8s.io/` annotations in addition to passing the requested CDI devices in the CRI field. This should still not be an issue unless a user has explicitly added `cdi.k8s.io/` annotation to the list of allowe annotations.
> 
> I think it should be sufficient to have `CDI_ENABLED` imply `RUNTIME_ENABLE_CDI` with an explicit setting of `RUNTIME_ENABLE_CDI` overriding the inferred value.

@elezar thanks for writing this up. I am aligned with this and agree that we can have the `cdi.enabled=true` option also trigger the enablement of CDI in runtimes. I will work on the operator changes.

pkg/config/engine/docker/docker.go
cdesiniotis: I see you fixed this in a later commit.

cmd/nvidia-ctk-installer/container/runtime/runtime.go
elezar: I purposfully chose this to match the other `RUNTIME_` envvars here.

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -104,10 +105,14 @@ func Flags(opts *Options) []cli.Flag {
...
 
 // ValidateOptions checks whether the specified options are valid
-func ValidateOptions(opts *Options, runtime string, toolkitRoot string) error {
+func ValidateOptions(c *cli.Context, opts *Options, runtime string, toolkitRoot string, to *toolkit.Options) error {
 	// We set this option here to ensure that it is available in future calls.
 	opts.RuntimeDir = toolkitRoot
 
+	if !c.IsSet("enable-cdi-in-runtime") {
+		opts.EnableCDI = to.CDI.Enabled
+	}
```
cdesiniotis, line 114: Now that I think about it again, we don't need any operator change with this in place. As you said in a prior comment:
> I think it should be sufficient to have CDI_ENABLED imply RUNTIME_ENABLE_CDI with an explicit setting of RUNTIME_ENABLE_CDI overriding the inferred value.

So by default, if a user sets `cdi.enabled=true` in the operator, we enable CDI in the runtime (e.g. containerd) while allowing them to opt-out of this behavior by manually configuring the `RUNTIME_ENABLE_CDI` environment variable in the toolkit.

tariq1890: So `cdi.default` becomes no-op?

cdesiniotis:
@elezar the changes you added lgtm. Can we merge this PR?

elezar:
> @elezar the changes you added lgtm. Can we merge this PR?

Sure. Let's get it in!

elezar (approved):

cmd/nvidia-ctk-installer/container/runtime/runtime.go
elezar: Yes, the default behaviour is to enable CDI in the runtime if CDI is enabled. This can be opted out of if the user sets the envvar -- or opted into if the toolkit option is not set. The Operator API is a separate concern to this.

elezar:
@cdesiniotis is there a reason to backport this, or are we ok targeting `v1.18.0`?

cdesiniotis:
@elezar I am okay to target `v1.18.0` so that we can include all dependency updates as well in the next operator release.

PR #873, opened 2025-01-21
Rename test folder to tests

ArangoGutierrez:
Rename the `test` folder to `tests` to be in line with other project conventions


internal/oci/spec_test.go
```diff
@@ -19,14 +19,14 @@ func TestMaintainSpec(t *testing.T) {
 	}
 
 	for _, f := range files {
-		inputSpecPath := filepath.Join(moduleRoot, "test/input", f)
+		inputSpecPath := filepath.Join(moduleRoot, "tests/input", f)
```
elezar, line 22: out of scope for this PR, but we should consider moving this to `testdata`.

elezar (changes requested):
I think you forgot to delete the existing folder.

ArangoGutierrez:
> I think you forgot to delete the existing folder.

Thanks for noticing, I have removed the folder, but I am confused, I did a simple `mv` ... I wonder what caused the folder duplication.

internal/oci/spec_test.go
ArangoGutierrez: I'll do as a follow up to this PR

elezar (approved):

PR #874, opened 2025-01-22
Skip graphics modifier in CSV mode

elezar:
In CSV mode the CSV files at /etc/nvidia-container-runtime/host-files-for-container.d/ should be the source of truth for container modifications. This change skips graphics modifications to a container. This prevents conflicts when handling files such as vulkan icd files which are already defined in the CSV file.

Applying both modifications leads to errors such as:
```
time="2025-01-21T18:55:11Z" level=info msg="Symlinking /var/lib/docker/overlay2/8592b399d9308cdbef42010765f81af3f921ed86aa07c7e686b59f1e72fc3e2d/merged/etc/vulkan/icd.d/nvidia_icd.json to /usr/lib/aarch64-linux-gnu/nvidia/nvidia_icd.json"
time="2025-01-21T18:55:11Z" level=error msg="failed to create link [/usr/lib/aarch64-linux-gnu/nvidia/nvidia_icd.json /etc/vulkan/icd.d/nvidia_icd.json]: failed to create symlink: failed to remove existing file: remove /var/lib/docker/overlay2/8592b399d9308cdbef42010765f81af3f921ed86aa07c7e686b59f1e72fc3e2d/merged/etc/vulkan/icd.d/nvidia_icd.json: device or resource busy": unknown.
```

### DIFF
--- a/internal/modifier/list.go
+++ b/internal/modifier/list.go
@@ -22,14 +22,12 @@ import (
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/oci"
 )
 
-type list struct {
-	modifiers []oci.SpecModifier
-}
+type List []oci.SpecModifier
 
 // Merge merges a set of OCI specification modifiers as a list.
 // This can be used to compose modifiers.
 func Merge(modifiers ...oci.SpecModifier) oci.SpecModifier {
-	var filteredModifiers []oci.SpecModifier
+	var filteredModifiers List
 	for _, m := range modifiers {
 		if m == nil {
 			continue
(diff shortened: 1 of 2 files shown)


klueska:
Seems reasonable, albeit a little hacky

elezar:
> Seems reasonable, albeit a little hacky

I have refactored this a bit based on our discussions. Note that the intent here is to have CSV mode in the NVIDIA Container Runtime behave the same way as what CDI mode would if the spec was generated from the CSV files. This is not currently the case due to the "graphics" modifiers being triggered when `NVIDIA_DRIVER_CAPABILITIES` includes graphics.

I understand that we don't want to "silently" ignore options that the user passes, but since this is already the case in CDI mode, I feel that doing the same in the case of CSV mode is justified.

klueska:
Thanks for the rework. It reads much better now.

klueska (approved):

PR #876, opened 2025-01-23
Add remote-test option for E2E

ArangoGutierrez:


ArangoGutierrez:
Next step is to have this to run during PR's, merge events and Tag cuts

test/e2e/e2e_test.go
```diff
@@ -46,7 +69,16 @@ var _ = BeforeSuite(func() {
...
+// runScript runs a script either locally or remotely
+// if the sshKey is provided, it runs the script remotely
 func runScript(script string) (string, error) {
+	if sshKey != "" {
+		return runRemoteScript(script)
+	}
+	return localRunScript(script)
+}
+
+func localRunScript(script string) (string, error) {
```
ArangoGutierrez, line 81: I wonder if I should move everything from this line forward to the `utils.go` file

test/e2e/utils.go
```diff
@@ -0,0 +1,75 @@
...
+	"bytes"
+	"fmt"
+	"text/template"
+)
+
+const Shebang = `#! /usr/bin/env bash
+set -xe
+`
+
+var dockerInstallTemplate = `
```
ArangoGutierrez, line 29: The idea here is that in the future, we will have different methods of installation to test, for example, CDI vs non-CDI.

test/e2e/utils.go
```diff
@@ -0,0 +1,75 @@
...
+
+: ${IMAGE:={{.Image}}}
+
+sudo ln -s /var/run/nvidia-container-toolkit/toolkit/nvidia-container-runtime-hook /usr/bin/nvidia-container-runtime-hook
```
elezar, line 33: Do we want to add a comment as to why this is required?

ArangoGutierrez: sure, more documentation never hurts

test/e2e/Makefile
```diff
@@ -18,9 +18,28 @@ include $(CURDIR)/versions.mk
...
-test:
+E2E_IMAGE_NAME ?= ghcr.io/nvidia/container-toolkit
+E2E_IMAGE_TAG ?= latest
+
+E2E_SSH_KEY ?= 
+E2E_SSH_USER ?= 
+E2E_SSH_HOST ?= 
+
+.PHONY: local-test remote-test
+# Local test assumes that the container-toolkit is already installed on the host
```
elezar, line 29: Do we need the distinction?

test/e2e/e2e_test.go
```diff
@@ -30,8 +35,26 @@ import (
 // Test context
 var (
 	ctx context.Context
+
+	imageRepo string
+	imageTag  string
```
elezar, line 40: Does this warrant a type? (same for the `ssh` info below)?

test/e2e/e2e_test.go
elezar: I think we could move all the script specifics to a separate file, but not `utils.go`.

test/e2e/nvidia-container-toolkit_test.go
```diff
@@ -18,13 +18,25 @@ package e2e
...
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
 // Integration tests for Docker runtime
-var _ = Describe("docker", func() {
+var _ = Describe("docker", Ordered, func() {
+	// Install the NVIDIA Container Toolkit
+	BeforeAll(func(ctx context.Context) {
+		if sshKey != "" {
```
elezar, line 31: As I mentioned, we should probably allow the isntallation logic to be run locally too.

test/e2e/nvidia-container-toolkit_test.go
```diff
@@ -18,13 +18,25 @@ package e2e
...
+			installScript, err := getInstallScript(dockerInstallTemplate, fmt.Sprintf("%s:%s", imageRepo, imageTag))
```
elezar, line 32: does it make sense to return an `Installer` instead which implements `Install() error`?

test/e2e/utils.go
```diff
@@ -0,0 +1,75 @@
...
+	${IMAGE}	\
+	--root /var/run/nvidia-container-toolkit	\
+	--runtime=docker	\
+	--config=/config-root/daemon.json	\
+	--driver-root=/	\
+	--no-daemon	\
+	--restart-mode=systemd
+`
+
+type DockerInstall struct {
```
elezar, line 49: This type isn't used.

test/e2e/utils.go
```diff
@@ -0,0 +1,75 @@
...
+var dockerInstallTemplate = `
+
+: ${IMAGE:={{.Image}}}
+
+sudo ln -s /var/run/nvidia-container-toolkit/toolkit/nvidia-container-runtime-hook /usr/bin/nvidia-container-runtime-hook
+
+docker run --pid=host --rm -i --privileged	\
+	-v /:/host	\
+	-v /var/run/docker.sock:/var/run/docker.sock	\
+	-v /var/run/nvidia-container-toolkit:/var/run/nvidia-container-toolkit	\
```
elezar, line 38: I think we should use a temporary path here since we ideally want to remove the config again after the tests.

test/e2e/e2e_test.go
elezar: Sorry, I should have been more clear. I think we should separate logic according to functionality / domain and name the files accordingly. Using `tools.go` as a catch-all does not achieve this.

tests/e2e/Makefile
```diff
@@ -18,9 +18,27 @@ include $(CURDIR)/versions.mk
 
 E2E_RUNTIME ?= docker
 
+E2E_INSTALL_CTK ?= true
+
+E2E_IMAGE_NAME ?= ghcr.io/nvidia/container-toolkit
+E2E_IMAGE_TAG ?=
+ifeq ($(E2E_IMAGE_TAG),)
+$(error E2E_IMAGE_TAG is not set)
+endif
```
elezar, line 27: Would it not make sense to use the tag that would be generated when running:
```
make -f deployments/container/Makefile build-ubuntu20.04
```
in the root by default.

i.e:
```
        --tag nvidia/container-toolkit:1.17.3-ubuntu20.04 \
```

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -24,7 +24,27 @@ import (
 )
 
 // Integration tests for Docker runtime
-var _ = Describe("docker", func() {
+var _ = Describe("docker", Ordered, func() {
+	var s Script
+
+	// Install the NVIDIA Container Toolkit
+	BeforeAll(func(ctx context.Context) {
+		switch {
```
elezar, line 32: Instead of using a switch statement here, let's add a constructor for the `Script` (or `Runner`) so that we can update this file as shown in the diff below. Note that we only construct a `Runner` once and pass it to the `Installer` that we also create.

```
diff --git a/tests/e2e/nvidia-container-toolkit_test.go b/tests/e2e/nvidia-container-toolkit_test.go
index cd76b7b8..c2dc3a10 100644
--- a/tests/e2e/nvidia-container-toolkit_test.go
+++ b/tests/e2e/nvidia-container-toolkit_test.go
@@ -25,19 +25,22 @@ import (
 
 // Integration tests for Docker runtime
 var _ = Describe("docker", Ordered, func() {
-	var s Script
+	var s Runner
 
 	// Install the NVIDIA Container Toolkit
 	BeforeAll(func(ctx context.Context) {
-		switch {
-		case sshKey == "":
-			s = localRun{}
-		default:
-			s = remoteRun{sshKey, sshUser, remoteHost}
-		}
+		s := NewRunner(
+			WithHost(remoteHost),
+			WithSshKey(sshKey),
+			WithSshUser(sshUser),
+		)
 
 		if installCTK {
-			installer, err := newInstaller(dockerInstallTemplate, imageRepo+":"+imageTag, sshKey, sshUser, remoteHost)
+			installer, err := NewToolkitInstaller(
+				WithRunner(s),
+				WithImageRepo(imageRepo),
+				WithImageTag(imageTag),
+			)
 
 			err = installer.Install()
```

tests/e2e/tools.go
```diff
@@ -0,0 +1,207 @@
...
+	--restart-mode=systemd
+`
+
+type Installer struct {
+	Image    string
+	Template string
+
+	SshKey     string
+	SshUser    string
+	RemoteHost string
```
elezar, line 66: These should not be `installer` members. The installer doesn't care if we're running remotely or locally. It should have a reference to a `Runner` that it uses to run the required commands.

tests/e2e/tools.go
```diff
@@ -0,0 +1,207 @@
...
+		SshUser:    sshUser,
+		RemoteHost: remoteHost,
+	}
+
+	return i, nil
+}
+
+func (i *Installer) Install() error {
+	// Parse the combined template
+	tmpl, err := template.New("dockerScript").Parse(i.Template)
```
elezar, line 83: nit: The template name need not be `dockerScript`.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -24,7 +24,27 @@ import (
 )
 
 // Integration tests for Docker runtime
-var _ = Describe("docker", func() {
+var _ = Describe("docker", Ordered, func() {
+	var s Script
```
elezar, line 28: This should probably be a `Runner` and not a `Script`.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -24,7 +24,27 @@ import (
...
+			s = remoteRun{sshKey, sshUser, remoteHost}
+		}
+
+		if installCTK {
+			installer, err := newInstaller(dockerInstallTemplate, imageRepo+":"+imageTag, sshKey, sshUser, remoteHost)
+			Expect(err).ToNot(HaveOccurred())
+
+			err = installer.Install()
+			Expect(err).ToNot(HaveOccurred())
+		}
```
elezar, line 45: The logic on which `Script` (or `Runner`) should be used should be pulled into a constructor with functional arguments. This instance should also be reused in the `Installer` so as to not duplicate the logic there.
```suggestion
		s := NewRunner(
			WithHost(remoteHost),
			WithSshKey(sshKey),
			WithSshUser(sshUser),
		)

		if installCTK {
			installer, err := NewToolkitInstaller(
				WithRunner(s),
				WithVersion(version),
			)
			Expect(err).ToNot(HaveOccurred())

			err = installer.Install()
			Expect(err).ToNot(HaveOccurred())
		}
```

tests/e2e/tools.go
```diff
@@ -0,0 +1,207 @@
...
+		s = localRun{}
+	default:
+		s = remoteRun{i.SshKey, i.SshUser, i.RemoteHost}
+	}
+
+	_, err = s.Run(renderedScript.String())
+	return err
+}
+
+type Script interface {
```
elezar, line 107: Since this inteface only has a single function `Run` the idiomatic Go name for it would be `Runner`. Let's also move it to a separate file.

tests/e2e/tools.go
```diff
@@ -0,0 +1,207 @@
...
+	Run(script string) (string, error)
+}
+
+type localRun struct{}
```
elezar, line 111:
```suggestion
// A localRunner runs scripts locally.
type localRun struct{}
```

tests/e2e/tools.go
```diff
@@ -0,0 +1,207 @@
...
+type remoteRun struct {
```
elezar, line 112:
```suggestion

// A remoteRunner runs scripts over SSH.
type remoteRunner struct {
```
(also wondering whether `sshRunner` isn't more accurate).

tests/e2e/tools.go
```diff
@@ -0,0 +1,207 @@
...
+	default:
+		s = remoteRun{i.SshKey, i.SshUser, i.RemoteHost}
+	}
+
+	_, err = s.Run(renderedScript.String())
+	return err
+}
+
+type Script interface {
+	Run(script string) (string, error)
```
elezar, line 108: Just wondering about this. Would we expect this to return `(string, string, error)` instead so that we can always access `STDOUT` and `STDERR`?

tests/e2e/tools.go
```diff
@@ -0,0 +1,207 @@
...
+		User: sshUser,
+		Auth: []ssh.AuthMethod{
+			ssh.PublicKeys(signer),
+		},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+	}
+
+	connectionFailed := false
+	for i := 0; i < 20; i++ {
+		client, err = ssh.Dial("tcp", remoteHost+":22", sshConfig)
```
elezar, line 192: Nit: Does it make sense to make the port configurable?

tests/e2e/tools.go
```diff
@@ -0,0 +1,207 @@
...
+		return "", fmt.Errorf("script execution failed: %v\nSTDOUT: %s\nSTDERR: %s", err, stdout.String(), stderr.String())
+	}
+
+	// Return the captured stdout and nil error
+	return stdout.String(), nil
+}
+
+func (r remoteRun) Run(script string) (string, error) {
+	// Create a new SSH connection
+	client, err := connectOrDie(r.SshKey, r.SshUser, r.RemoteHost)
```
elezar, line 142: Should we create a new SSH session every time we `Run` a script?

tests/e2e/tools.go
```diff
@@ -0,0 +1,207 @@
...
+	sshConfig := &ssh.ClientConfig{
+		User: sshUser,
+		Auth: []ssh.AuthMethod{
+			ssh.PublicKeys(signer),
+		},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+	}
+
+	connectionFailed := false
+	for i := 0; i < 20; i++ {
```
elezar, line 191: I don't think we need a retry at this stage.

tests/e2e/Makefile
elezar: I just coppied a snippet from the shell output when running the make command. Do we think the following is a reasonable mode of operation:

1. Run `./scripts/build-packages.sh ubuntu18.04-amd64` to build the packages.
2. Run `make -f deployment/container/Makefile build` to build the required image.
3. Run `make -f tests/e2e/Makefile test` to run the tests for the built version locally?

ArangoGutierrez:
@elezar PTAL

PR #877, opened 2025-01-23
 Disable mounting of compat libs from container by default

elezar:
This change passes the `--no-cntlibs` argument to the `nvidia-container-cli` from the nvidia-container-runtime-hook to disable overwriting host drivers with the compat libs from a container being started.

Note that this may be a breaking change for some applications. For that reason an `allow-cuda-compat-libs-from-container` feature flag is provided to opt-in to the previous behaviour.


ArangoGutierrez (approved):
lgtm

PR #881, opened 2025-01-24
Add IMEX binaries to CDI discovery

elezar:
This change adds the nvidia-imex and nivdia-imex-ctl binaries to the list of driver binaries that are searched when using CDI.


klueska (approved):

klueska:
Simple, yet effective

ArangoGutierrez (approved):

PR #905, opened 2025-02-06
[no-relnote] fix step holodeck_public_dns_name variable name E2E

ArangoGutierrez:


.github/workflows/e2e.yaml
```diff
@@ -71,7 +71,7 @@ jobs:
           echo "$SSH_KEY" > "$e2e_ssh_key"
           chmod 600 "$e2e_ssh_key"
           export E2E_SSH_KEY="$e2e_ssh_key"
-          export E2E_SSH_HOST="${{ steps.get_public_dns_name.outputs.result }}"
+          export E2E_SSH_HOST="${{ steps.holodeck_public_dns_name.outputs.result }}"
```
elezar, line 74: Question: Could we set this in the `env` section instead?

ArangoGutierrez: yes

ArangoGutierrez:
> /ok-to-test

In order for E2E to run during PR's we might need to change 
https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/.github/workflows/e2e.yaml#L30

ArangoGutierrez:
> > > /ok-to-test
> > 
> > 
> > In order for E2E to run during PR's we might need to change https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/.github/workflows/e2e.yaml#L30
> 
> Does pr-bot not trigger a push?

we need to remove this bit `${{ github.event.workflow_run.event == 'push' }}`

elezar (approved):

PR #906, opened 2025-02-06
Add CUDA forward compatibility hook

elezar:
With #877 the default behaviour of the NVIDIA Container Runtime / NVIDIA Container Runtime Hook was changed to *not* mount compat libraries from the container into the container. This removed "automatic" support for CUDA Forward compatibility.

This change attempts to address this by adding a `createContainerHook` that will create a file in `/etc/ld.so.conf.d/` in the container to ensure that the `/usr/local/cuda/compat` libraries are added to the ldcache over the libraries mounted from the host. The provided host diver version is compared to the version of the compat libraries in the container and the config update is only performed if the compat libraries are newer than the host drivers.

Note that the hook *only creates a file* in the container's file system and does not perform any mount operations. This means that this mechanism is not present the same vulnerabilities causing CVE-2024-0132 and CVE-2025-23359.

In the case of the legacy runtime, this behaviour is only triggered if the `allow-cuda-compat-libs-from-container` feature flag is not enabled. The CDI spec generation has also been extended to include this hook.

### DIFF
--- a/cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
+++ b/cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
@@ -1,0 +1,221 @@
+/**
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package cudacompat
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/oci"
+)
+
+const (
+	cudaCompatPath = "/usr/local/cuda/compat"
+	// cudaCompatLdsoconfdFilenamePattern specifies the pattern for the filename
+	// in ld.so.conf.d that includes a reference to the CUDA compat path.
+	// The 00-compat prefix is chosen to ensure that these libraries have a
+	// higher precedence than other libraries on the system.
+	cudaCompatLdsoconfdFilenamePattern = "00-compat-*.conf"
+)
+
+type command struct {
+	logger logger.Interface
+}
+
+type options struct {
+	hostDriverVersion string
+	containerSpec     string
+}
+
+// NewCommand constructs a cuda-compat command with the specified logger
+func NewCommand(logger logger.Interface) *cli.Command {
+	c := command{
+		logger: logger,
+	}
+	return c.build()
+}
+
+// build the enable-cuda-compat command
+func (m command) build() *cli.Command {
+	cfg := options{}
+
+	// Create the 'enable-cuda-compat' command
+	c := cli.Command{
+		Name:  "enable-cuda-compat",
+		Usage: "This hook ensures that the folder containing the CUDA compat libraries is added to the ldconfig search path if required.",
+		Before: func(c *cli.Context) error {
+			return m.validateFlags(c, &cfg)
+		},
+		Action: func(c *cli.Context) error {
+			return m.run(c, &cfg)
+		},
+	}
+
+	c.Flags = []cli.Flag{
+		&cli.StringFlag{
+			Name:        "host-driver-version",
+			Usage:       "Specify the host driver version. If the CUDA compat libraries detected in the container do not have a higher MAJOR version, the hook is a no-op.",
+			Destination: &cfg.hostDriverVersion,
+		},
+		&cli.StringFlag{
+			Name:        "container-spec",
+			Hidden:      true,
+			Category:    "testing-only",
+			Usage:       "Specify the path to the OCI container spec. If empty or '-' the spec will be read from STDIN",
+			Destination: &cfg.containerSpec,
+		},
+	}
+
+	return &c
+}
+
+func (m command) validateFlags(_ *cli.Context, cfg *options) error {
+	return nil
+}
+
+func (m command) run(_ *cli.Context, cfg *options) error {
+	if cfg.hostDriverVersion == "" {
+		return nil
+	}
+
+	s, err := oci.LoadContainerState(cfg.containerSpec)
+	if err != nil {
+		return fmt.Errorf("failed to load container state: %w", err)
+	}
+
+	containerRootDir, err := s.GetContainerRoot()
+	if err != nil {
+		return fmt.Errorf("failed to determined container root: %w", err)
+	}
+
+	containerForwardCompatDir, err := m.getContainerForwardCompatDir(containerRoot(containerRootDir), cfg.hostDriverVersion)
+	if err != nil {
+		return fmt.Errorf("failed to get container forward compat directory: %w", err)
+	}
+	if containerForwardCompatDir == "" {
+		return nil
+	}
+
+	return m.createLdsoconfdFile(containerRoot(containerRootDir), cudaCompatLdsoconfdFilenamePattern, containerForwardCompatDir)
+}
+
+func (m command) getContainerForwardCompatDir(containerRoot containerRoot, hostDriverVersion string) (string, error) {
+	if hostDriverVersion == "" {
+		m.logger.Debugf("Host driver version not specified")
+		return "", nil
+	}
+	if !containerRoot.hasPath(cudaCompatPath) {
+		m.logger.Debugf("No CUDA forward compatibility libraries directory in container")
+		return "", nil
+	}
+	if !containerRoot.hasPath("/etc/ld.so.cache") {
+		m.logger.Debugf("The container does not have an LDCache")
+		return "", nil
+	}
+
+	libs, err := containerRoot.globFiles(filepath.Join(cudaCompatPath, "libcuda.so.*.*"))
+	if err != nil {
+		m.logger.Warningf("Failed to find CUDA compat library: %w", err)
+		return "", nil
+	}
+
+	if len(libs) == 0 {
+		m.logger.Debugf("No CUDA forward compatibility libraries container")
+		return "", nil
+	}
+
+	if len(libs) != 1 {
+		m.logger.Warningf("Unexpected number of CUDA compat libraries in container: %v", libs)
+		return "", nil
+	}
+
+	compatDriverVersion := strings.TrimPrefix(filepath.Base(libs[0]), "libcuda.so.")
+	compatMajor, err := extractMajorVersion(compatDriverVersion)
+	if err != nil {
+		return "", fmt.Errorf("failed to extract major version from %q: %v", compatDriverVersion, err)
+	}
+
+	driverMajor, err := extractMajorVersion(hostDriverVersion)
+	if err != nil {
+		return "", fmt.Errorf("failed to extract major version from %q: %v", hostDriverVersion, err)
+	}
+
+	if driverMajor >= compatMajor {
+		m.logger.Debugf("Compat major version is not greater than the host driver major version (%v >= %v)", hostDriverVersion, compatDriverVersion)
+		return "", nil
+	}
+
+	resolvedCompatDir := strings.TrimPrefix(filepath.Dir(libs[0]), string(containerRoot))
+	return resolvedCompatDir, nil
+}
+
+// createLdsoconfdFile creates a file at /etc/ld.so.conf.d/ in the specified root.
+// The file is created at /etc/ld.so.conf.d/{{ .pattern }} using `CreateTemp` and
+// contains the specified directories on each line.
+func (m command) createLdsoconfdFile(in containerRoot, pattern string, dirs ...string) error {
+	if len(dirs) == 0 {
+		m.logger.Debugf("No directories to add to /etc/ld.so.conf")
+		return nil
+	}
+
+	ldsoconfdDir, err := in.resolve("/etc/ld.so.conf.d")
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(ldsoconfdDir, 0755); err != nil {
+		return fmt.Errorf("failed to create ld.so.conf.d: %w", err)
+	}
+
+	configFile, err := os.CreateTemp(ldsoconfdDir, pattern)
+	if err != nil {
+		return fmt.Errorf("failed to create config file: %w", err)
+	}
+	defer configFile.Close()
+
+	m.logger.Debugf("Adding directories %v to %v", dirs, configFile.Name())
+
+	added := make(map[string]bool)
+	for _, dir := range dirs {
+		if added[dir] {
+			continue
+		}
+		_, err = configFile.WriteString(fmt.Sprintf("%s\n", dir))
+		if err != nil {
+			return fmt.Errorf("failed to update config file: %w", err)
+		}
+		added[dir] = true
+	}
+
+	// The created file needs to be world readable for the cases where the container is run as a non-root user.
+	if err := configFile.Chmod(0644); err != nil {
+		return fmt.Errorf("failed to chmod config file: %w", err)
+	}
+
+	return nil
+}
+
+// extractMajorVersion parses a version string and returns the major version as an int.
+func extractMajorVersion(version string) (int, error) {
+	majorString := strings.SplitN(version, ".", 2)[0]
+	return strconv.Atoi(majorString)
+}
--- a/cmd/nvidia-cdi-hook/cudacompat/cudacompat_test.go
+++ b/cmd/nvidia-cdi-hook/cudacompat/cudacompat_test.go
@@ -1,0 +1,173 @@
+/*
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+*/
+
+package cudacompat
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	testlog "github.com/sirupsen/logrus/hooks/test"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCompatLibs(t *testing.T) {
+	logger, _ := testlog.NewNullLogger()
+
+	testCases := []struct {
+		description                       string
+		contents                          map[string]string
+		hostDriverVersion                 string
+		expectedContainerForwardCompatDir string
+	}{
+		{
+			description:       "empty root",
+			hostDriverVersion: "222.55.66",
+		},
+		{
+			description: "compat lib is newer; no ldcache",
+			contents: map[string]string{
+				"/usr/local/cuda/compat/libcuda.so.333.88.99": "",
+			},
+			hostDriverVersion: "222.55.66",
+		},
+		{
+			description: "compat lib is newer; ldcache",
+			contents: map[string]string{
+				"/etc/ld.so.cache": "",
+				"/usr/local/cuda/compat/libcuda.so.333.88.99": "",
+			},
+			hostDriverVersion:                 "222.55.66",
+			expectedContainerForwardCompatDir: "/usr/local/cuda/compat",
+		},
+		{
+			description: "compat lib is older; ldcache",
+			contents: map[string]string{
+				"/etc/ld.so.cache": "",
+				"/usr/local/cuda/compat/libcuda.so.111.88.99": "",
+			},
+			hostDriverVersion:                 "222.55.66",
+			expectedContainerForwardCompatDir: "",
+		},
+		{
+			description: "compat lib has same major version; ldcache",
+			contents: map[string]string{
+				"/etc/ld.so.cache": "",
+				"/usr/local/cuda/compat/libcuda.so.222.88.99": "",
+			},
+			hostDriverVersion:                 "222.55.66",
+			expectedContainerForwardCompatDir: "",
+		},
+		{
+			description: "driver version empty; ldcache",
+			contents: map[string]string{
+				"/etc/ld.so.cache": "",
+				"/usr/local/cuda/compat/libcuda.so.222.88.99": "",
+			},
+			hostDriverVersion: "",
+		},
+		{
+			description: "symlinks are followed",
+			contents: map[string]string{
+				"/etc/ld.so.cache": "",
+				"/etc/alternatives/cuda/compat/libcuda.so.333.88.99": "",
+				"/usr/local/cuda": "symlink=/etc/alternatives/cuda",
+			},
+			hostDriverVersion:                 "222.55.66",
+			expectedContainerForwardCompatDir: "/etc/alternatives/cuda/compat",
+		},
+		{
+			description: "symlinks stay in container",
+			contents: map[string]string{
+				"/etc/ld.so.cache":             "",
+				"/compat/libcuda.so.333.88.99": "",
+				"/usr/local/cuda":              "symlink=../../../../../../",
+			},
+			hostDriverVersion:                 "222.55.66",
+			expectedContainerForwardCompatDir: "/compat",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			containerRootDir := t.TempDir()
+			for name, contents := range tc.contents {
+				target := filepath.Join(containerRootDir, name)
+				require.NoError(t, os.MkdirAll(filepath.Dir(target), 0755))
+
+				if strings.HasPrefix(contents, "symlink=") {
+					require.NoError(t, os.Symlink(strings.TrimPrefix(contents, "symlink="), target))
+					continue
+				}
+
+				require.NoError(t, os.WriteFile(target, []byte(contents), 0600))
+			}
+
+			c := command{
+				logger: logger,
+			}
+			containerForwardCompatDir, err := c.getContainerForwardCompatDir(containerRoot(containerRootDir), tc.hostDriverVersion)
+			require.NoError(t, err)
+			require.EqualValues(t, tc.expectedContainerForwardCompatDir, containerForwardCompatDir)
+		})
+	}
+}
+
+func TestUpdateLdconfig(t *testing.T) {
+	logger, _ := testlog.NewNullLogger()
+	testCases := []struct {
+		description      string
+		folders          []string
+		expectedContents string
+	}{
+		{
+			description: "no folders; have no contents",
+		},
+		{
+			description:      "single folder is added",
+			folders:          []string{"/usr/local/cuda/compat"},
+			expectedContents: "/usr/local/cuda/compat\n",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			containerRootDir := t.TempDir()
+			c := command{
+				logger: logger,
+			}
+			err := c.createLdsoconfdFile(containerRoot(containerRootDir), cudaCompatLdsoconfdFilenamePattern, tc.folders...)
+			require.NoError(t, err)
+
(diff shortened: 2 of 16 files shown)


cmd/nvidia-cdi-hook/compat-libs/compat-libs.go
```diff
@@ -0,0 +1,208 @@
...
+	return c.build()
+}
+
+// build the compat-libs command
+func (m command) build() *cli.Command {
+	cfg := options{}
+
+	// Create the 'compat-libs' command
+	c := cli.Command{
+		Name: "compat-libs",
```
elezar, line 55: Should this hook have a better name?

Currently it reads:
```
nvidia-cdi-hook compat-libs --driver-version 999.88.77
```

elezar: Changed this to:
```
nvidia-cdi-hook cuda-compat --drvier-version 999.88.77
```
We could use this hook for other compat functions such as enhanced compatibility.

internal/modifier/gated.go
```diff
@@ -78,5 +79,10 @@ func NewFeatureGatedModifier(logger logger.Interface, cfg *config.Config, image
 		discoverers = append(discoverers, d)
 	}
 
+	if !cfg.Features.AllowCUDACompatLibsFromContainer.IsEnabled() {
```
cdesiniotis, line 82: Maybe I am confused, but shouldn't this be negated? I would expect this hook to be added only when `allow-cuda-compat-libs-from-container` is enabled.

elezar: The purpose of this hook is to provide the same functionality as is currently in `libnvidia-container`. This means that if we disable that (as we do by default in this branch), we would have to inject the hook. If a user opts-in to the older behaviour, then the hook is not required.

internal/modifier/gated.go
```diff
@@ -78,5 +79,22 @@ func NewFeatureGatedModifier(logger logger.Interface, cfg *config.Config, image
...
+		compatLibHookDiscoverer := discover.NewCUDACompatHookDiscoverer(logger, cfg.NVIDIACTKConfig.Path, driver)
+		discoverers = append(discoverers, compatLibHookDiscoverer)
+		if cfg.NVIDIAContainerRuntimeConfig.Mode == "legacy" {
+			ldcacheIpdateHookDiscoverer, err := discover.NewLDCacheUpdateHook(
```
elezar, line 86: Note: Since the `nvidia-container-runtime-hook` is invoked as a `prestart` hook, this is done BEFORE the `createContainer` hook that we insert above. This means that we need to once again run the update ldcache hook.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,238 @@
+/**
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
```
tariq1890, line 2:
```suggestion
# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	if !containerRoot.hasPath("/etc/ld.so.cache") {
+		// If there is no ldcache in the container, the hook is a no-op.
+		return nil
+	}
+	if !containerRoot.hasPath("/etc/ld.so.conf.d") {
+		// If the /etc/ld.so.conf.d folder does not exist in the container, the hook is a no-op.
+		return nil
+	}
```
cdesiniotis, line 117: Question -- what is the rationale behind this condition? I noticed in `createConfig()` we create this directory before attempting to create a file under it.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,238 @@
...
+	m.logger.Debugf("Adding folders %v to %v", folders, configFile.Name())
+
+	configured := make(map[string]bool)
+	for _, folder := range folders {
+		if configured[folder] {
+			continue
+		}
+		_, err = configFile.WriteString(fmt.Sprintf("%s\n", folder))
+		if err != nil {
+			return fmt.Errorf("failed to update ld.so.conf.d: %w", err)
```
cdesiniotis, line 227:
```suggestion
			return fmt.Errorf("failed to update config file: %w", err)
```

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,238 @@
...
+	if len(folders) == 0 {
+		m.logger.Debugf("No folders to add to /etc/ld.so.conf")
+		return nil
+	}
+
+	if err := os.MkdirAll(filepath.Join(root, "/etc/ld.so.conf.d"), 0755); err != nil {
+		return fmt.Errorf("failed to create ld.so.conf.d: %w", err)
+	}
+
+	configFile, err := os.CreateTemp(filepath.Join(root, "/etc/ld.so.conf.d"), "00-compat-*.conf")
```
cdesiniotis, line 212: Question -- since the file we are creating is not a temporary file, I am assuming `CreateTemp` is used here for generating the random string suffix?

cmd/nvidia-cdi-hook/cudacompat/cudacompat_test.go
```diff
@@ -0,0 +1,131 @@
...
+				"/etc/ld.so.conf.d/.hidden":                   "",
+				"/usr/local/cuda/compat/libcuda.so.222.88.99": "",
+			},
+			options: options{
+				hostDriverVersion: "222.55.66",
+			},
+			expectedCompatContents: "",
+		},
+		{
+			description: "driver version empty; ldcache",
```
cdesiniotis, line 89: Question -- I am just trying to grok this scenario. IIUC this means we always use the compat libraries if we cannot discover the driver version of the libraries present on the host. Does libnvidia-container have the same behavior? Taking a step back, shouldn't failing to detect the driver version from the host not result in an error?

internal/modifier/gated.go
cdesiniotis: Thanks for the clarification.

internal/modifier/gated.go
cdesiniotis: If we create the file in `nvidia-container-runtime-hook`, then wouldn't that remove the need to add the `cuda-compat` createContainer hook altogether?

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: The intent is to not create this in containers that don't already have ldconfig set up to consume config files in that folder. I think we could skip this check though.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: Yes we're leveraging that for the random string. I'm sure we could come up with an alternative -- or even just use a timestamp -- if you prefer.

internal/modifier/gated.go
elezar: I did try it, and my initial assumption was incorrect. Because we invoke ldconfig as:
```
        argv = (char * []){cnt->cfg.ldconfig, "-f", "/etc/ld.so.conf", "-C", "/etc/ld.so.cache", cnt->cfg.libs_dir, cnt->cfg.libs32_dir, NULL};
```
in `libnvidia-container`, this means that the `cnt->cfg.libs_dir` and `cnt->cfg.libs32_dir` folders *take precedence* over the files in `/etc/ld.so.conf.d` and the CUDA libraries present there are used.

We could rework the `libnvidia-container` implementation further, but the intent of this change is also to provide the functionality for #910 so that we can deprecate the legacy code path by default.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: I have removed this check to simplify the hook.

cmd/nvidia-cdi-hook/cudacompat/cudacompat_test.go
elezar: Yes, failing to detect the driver version will cause a failure in `libnvidia-container`. That behaviour is not captured here since the driver version extraction happens earlier in the process.

The reason that I didn't want to raise an error in this case is that I didn't want to cause a container to fail to start due to an error in the runtime -- more drastically changing the behaviour. I think I can change the behaviour of the hook to be a no-op if the driver version is not specified though.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: Should be fixed.

cmd/nvidia-cdi-hook/cudacompat/cudacompat_test.go
elezar: I have change the behaviour to a no-op if the driver root is empty.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
cdesiniotis: Let's stick with the implementation as is.

internal/modifier/gated.go
cdesiniotis: Got it. So for legacy mode, we are adding two createContainer hooks ( the `cuda-compat` hook and the `update-ldcache`) to ensure the compat libs are used, the caveat being that we execute ldconfig twice.

internal/runtime/runtime_factory.go
```diff
@@ -128,8 +128,8 @@ func supportedModifierTypes(mode string) []string {
 		return []string{"nvidia-hook-remover", "mode"}
 	case "csv":
 		// For CSV mode we support mode and feature-gated modification.
-		return []string{"nvidia-hook-remover", "mode", "feature-gated"}
+		return []string{"nvidia-hook-remover", "feature-gated", "mode"}
```
cdesiniotis, line 131: Question -- why does the ordering matter?

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -166,4 +168,51 @@ var _ = Describe("docker", Ordered, func() {
 			Expect(referenceOutput).To(Equal(out4))
 		})
 	})
+
+	Describe("CUDA Forward compatibility", Ordered, func() {
```
cdesiniotis, line 172: Nice! I notice the test cases get skipped since our holodeck installs the latest driver: https://github.com/NVIDIA/nvidia-container-toolkit/actions/runs/13331020349/job/37235261007?pr=906#step:10:95

Any plans to downgrade the version of the driver we install in our e2e tests?

internal/modifier/gated.go
cdesiniotis, line 82: Question -- are `features` in the config only applicable to libnvidia-container? If yes, how can we make this more clear to users? From the user's perspective, it may be a bit confusing that compat libs do end up being used even if `features.allow-cuda-compat-libs-from-container=false` is set in the config.

cdesiniotis (commented):
I left a few questions. But overall this LGTM. Would be good to get another pair of eyes on this.

tests/e2e/nvidia-container-toolkit_test.go
elezar: There are plans, but I have not been able to persuade holodeck to install a lower driver version yet. I need to figure out which driver versions are available on the instance as a starting point.

internal/modifier/gated.go
elezar: `features` are not strictly limited to `libnvidia-container`. Note that `libnvidia-container` does not process any of these features and they are ONLY checked from the `nvidia-container-runtime` or `nvidia-container-runtime-hook`. I think this is a side-effect of us not being specific enough when naming the feature flag in this case. Since features are not listed in the config by default, and only listed if toggled explicitly, I think we should be ok with this slight confusion.

internal/runtime/runtime_factory.go
elezar: In the CSV case, I don't think it's strictly needed, but the intent is to ensure that the `ldconfig` hook that is injected as part of the `mode` modifications is *after* the hook to configure the compat libraries. Note that since we may also add libraries in the `feature-gated` modifiers, we may have been missing these libraries from the ldcache in the past.

internal/modifier/gated.go
elezar: Yes, that is correct.

alexeldeib:
👋 I wanted to avoid cluttering this PR but I'm impacted by the problem this PR intends to fix and following along closely, opened an issue over here to discuss https://github.com/NVIDIA/nvidia-container-toolkit/issues/934

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/oci"
+)
+
+const (
+	cudaCompatPath = "/usr/local/cuda/compat"
+	// cudaCompatLdsoconfdFilenamePattern specifies the pattern for the filename
+	// in ld.so.conf.d that includes a reference to the CUDA compat path.
+	// The 00-compat prefix is chosen to ensure that these libraries have a
+	// higher precedence than other libraries on the system.
+	cudaCompatLdsoconfdFilenamePattern = "00-compat-*.conf"
```
klueska, line 38: I don't see this being a problem in practice. But it is worth noting that it's not *guaranteed* to come first.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,247 @@
...
+type command struct {
+	logger logger.Interface
+}
+
+type options struct {
+	hostDriverVersion string
+	containerSpec     string
+}
+
+// NewCommand constructs an cuda-compat command with the specified logger
```
klueska, line 50:
```suggestion
// NewCommand constructs a cuda-compat command with the specified logger
```

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,247 @@
...
+func NewCommand(logger logger.Interface) *cli.Command {
+	c := command{
+		logger: logger,
+	}
+	return c.build()
+}
+
+// build the cuda-compat command
+func (m command) build() *cli.Command {
```
klueska, line 59: why `m` as the variable? I would think `c` would be more fitting.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,247 @@
...
+		return fmt.Errorf("failed to determined container root: %w", err)
+	}
+
+	return m.updateCUDACompatLibs(root(containerRoot), cfg.hostDriverVersion)
+}
+
+func (m command) updateCUDACompatLibs(containerRoot root, hostDriverVersion string) error {
+	if hostDriverVersion == "" {
+		return nil
+	}
```
klueska, line 116: Is this redundant with the same check above? Is there a case where we call this out of band from `run()`?

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,247 @@
...
+
+func (m command) validateFlags(c *cli.Context, cfg *options) error {
+	return nil
+}
+
+func (m command) run(c *cli.Context, cfg *options) error {
+	if cfg.hostDriverVersion == "" {
+		return nil
+	}
+	s, err := oci.LoadContainerState(cfg.containerSpec)
```
klueska, line 100: nit:
```suggestion

	s, err := oci.LoadContainerState(cfg.containerSpec)
```

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,247 @@
...
+
+	containerRoot, err := s.GetContainerRoot()
+	if err != nil {
+		return fmt.Errorf("failed to determined container root: %w", err)
+	}
+
+	return m.updateCUDACompatLibs(root(containerRoot), cfg.hostDriverVersion)
+}
+
+func (m command) updateCUDACompatLibs(containerRoot root, hostDriverVersion string) error {
```
klueska, line 113: Can we transform this function into one that returns a `(bool, error)` whose job at the end of the day is to check if `(driverMajor < compatMajor)`.

We can then use the result of this to decide if we want to call `m.createLdsoconfdFile(...)` or not.

Separating it like this makes it clearer in the `run()` call what the noop path is and what the yesop path is.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,247 @@
...
+// Symlinks are resolved, but are guaranteed to resolve in the root.
+func (r root) resolve(path string) (string, error) {
+	absolute := filepath.Clean(filepath.Join(string(r), path))
+	return symlink.FollowSymlinkInScope(absolute, string(r))
+}
+
+// createLdsoconfdFile creates a file at /etc/ld.so.conf.d/ in the specified root.
+// The file is created at /etc/ld.so.conf.d/{{ .pattern }} using `CreateTemp` and
+// contains the specified folders on each line.
+func (m command) createLdsoconfdFile(in root, pattern string, folders ...string) error {
```
klueska, line 207: Can you move this up to group it under the other calls for this type.

klueska, line 207: nit: can we call the variable `dirs` instead of folders (and adjust the comment appropriately too)?

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil
+	}
+
+	ldsoconfdDir, err := in.resolve("/etc/ld.so.conf.d")
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(ldsoconfdDir, 0755); err != nil {
+		return fmt.Errorf("failed to create ld.so.conf.d: %w", err)
+	}
```
klueska, line 219: What is happening here? You try and resolve `ldsoconfdDir` and only if there is not error, you attempt to create it?

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,247 @@
...
+
+	configFile, err := os.CreateTemp(ldsoconfdDir, pattern)
+	if err != nil {
+		return fmt.Errorf("failed to create config file: %w", err)
+	}
+	defer configFile.Close()
+
+	m.logger.Debugf("Adding folders %v to %v", folders, configFile.Name())
+
+	configured := make(map[string]bool)
```
klueska, line 229:
```suggestion
	added := make(map[string]bool)
```

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -269,6 +274,8 @@ func (m command) generateSpec(opts *options) (spec.Interface, error) {
 		nvcdi.WithLibrarySearchPaths(opts.librarySearchPaths.Value()),
 		nvcdi.WithCSVFiles(opts.csv.files.Value()),
 		nvcdi.WithCSVIgnorePatterns(opts.csv.ignorePatterns.Value()),
+		// We set the following to allow for dependency injection:
+		nvcdi.WithNvmlLib(opts.nvmllib),
```
klueska, line 278: Is this going to cause problems? I have a vague memory of us purposefully *not* relying on NVML in this subcommand. I also don't quite see why it's necessary yet. Maybe that will become clearer further down.

internal/modifier/gated.go
```diff
@@ -78,5 +79,24 @@ func NewFeatureGatedModifier(logger logger.Interface, cfg *config.Config, image
 		discoverers = append(discoverers, d)
 	}
 
+	if !cfg.Features.AllowCUDACompatLibsFromContainer.IsEnabled() && !cfg.Features.DisableCUDACompatLibHook.IsEnabled() {
```
klueska, line 82: So the check for `!cfg.Features.AllowCUDACompatLibsFromContainer.IsEnabled()` is what's key here -- to ensure taht we don't do double work in this new hook AND in libnvidiacontainer.

cmd/nvidia-ctk/cdi/generate/generate.go
klueska: It never became clear. What am I missing?

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: To clarify, these don't have to come first, they just have to come before the folder contianing the driver libraries mounted from the host. 

* In the CDI case, the host driver libraries have their own config file named [`00-nvcr-*.conf`](https://github.com/NVIDIA/nvidia-container-toolkit/blob/65b575fa96023cdcb849aca088cb42a03c4fd4d3/cmd/nvidia-cdi-hook/update-ldcache/update-ldcache.go#L156-L161) which is guaranteed to come after `00-compat-*.conf`.
* In the legacy case, the host driver libraries are mounted to the standard libraries which have config files which do not have `00-` prefixes. (e.g. `/etc/ld.so.conf.d/x86_64-linux-gnu.conf` or `/etc/ld.so.conf.d/aarch64-linux-gnu.conf` on ubuntu) or are implied as trusted folders in [`ldconfig`](https://man7.org/linux/man-pages/man8/ldconfig.8.html).

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: The only place we call it out of band is tests. The check above is to avoid even decoding the container config and reading the bundle to extract the root.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: `containerRoot.resolve` resolves a path in the container root -- following any symlinks and ensuring that we do not escape the container filesystem. The final path does not have to exsit which is why we create it. (`MkdirAll` is a no-op if the directory exists).

internal/modifier/gated.go
elezar: Yes, I want `allow-cuda-compat-libs-from-container` to revert to the "old" `libnvidia-container`-based mechanism entirely and use the new hook-based mechanism by default. I don't want two mechanisms being used at the same time.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: I'll move the container root logic to a separate file to ensure that these are grouped properly.

elezar: Sure.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: Yes, `c` would work better, but I've been using `m` in other commands since we already have the `cli.Context` parameter as `c`. I would rather keep this as `m` for this PR and then update across all our commands for consistency.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: I have updated to flow to first determine the `compatDir` in the container and only to create the file if this is non-empty. I thought this was cleaner than passing `(bool, string, error)` back from the function.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: Done.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// createLdsoconfdFile creates a file at /etc/ld.so.conf.d/ in the specified root.
+// The file is created at /etc/ld.so.conf.d/{{ .pattern }} using `CreateTemp` and
+// contains the specified directories on each line.
+func (m command) createLdsoconfdFile(in containerRoot, pattern string, dirs ...string) error {
+	if len(dirs) == 0 {
+		m.logger.Debugf("No directories to add to /etc/ld.so.conf")
+		return nil
+	}
```
ArangoGutierrez, line 172: same check as in line 114, do we intend to perform this double check?

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,208 @@
...
+
+	return m.createLdsoconfdFile(containerRoot(containerRootDir), cudaCompatLdsoconfdFilenamePattern, containerForwardCompatDir)
+}
+
+func (m command) getContainerForwardCompatDir(containerRoot containerRoot, hostDriverVersion string) (string, error) {
+	if hostDriverVersion == "" {
+		m.logger.Debugf("Host driver version not specified")
+		return "", nil
+	}
```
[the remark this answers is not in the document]
elezar: This is intentional. See https://github.com/NVIDIA/nvidia-container-toolkit/pull/906#discussion_r1966497622

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: Yes. This is intentional: https://github.com/NVIDIA/nvidia-container-toolkit/pull/906#discussion_r1966497622

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,208 @@
...
+		m.logger.Warningf("Unexpected number of CUDA compat libraries in container: %v", libs)
+		return "", nil
+	}
+
+	compatDriverVersion := strings.TrimPrefix(filepath.Base(libs[0]), "libcuda.so.")
+	compatMajor := strings.SplitN(compatDriverVersion, ".", 2)[0]
+
+	driverMajor := strings.SplitN(hostDriverVersion, ".", 2)[0]
+
+	if driverMajor >= compatMajor {
```
rajatchopra, line 156: The string comparison shouldn't be done. There should be an Atoi before this.
Example: driverMajor = "11", compatMajor = "101"
The current code will show that driverMajor is greater than compatMajor, which is wrong. Can add a testcase too.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -0,0 +1,208 @@
...
+	return c.build()
+}
+
+// build the cuda-compat command
+func (m command) build() *cli.Command {
+	cfg := options{}
+
+	// Create the 'cuda-compat' command
+	c := cli.Command{
+		Name:  "cuda-compat",
```
rajatchopra, line 63: Grammar review for fun.
If its a command, then it should be a verb. 'enable-cuda-compat'?

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: Yes, that's true. This relies on the current state of our driver versions where the branch (major) version is always some three digit number. This is also why string comparison is used in `libnvidia-container` at present, but is not future proof and will break when we release a `1xxx` driver branch. I will update to use `Atoi` instead.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
klueska: Seems good enough. I would still have preferred the check against the driver version up in the `run()` call itself, but its fine as is.

klueska (approved):

ArangoGutierrez (approved):

internal/config/features.go
```diff
@@ -25,6 +25,12 @@ type features struct {
 	// If this feature flag is not set to 'true' only host-rooted config paths
 	// (i.e. paths starting with an '@' are considered valid)
 	AllowLDConfigFromContainer *feature `toml:"allow-ldconfig-from-container,omitempty"`
+	// DisableCUDACompatLibHook, when enabled skips the injection of a specific
+	// hook to process CUDA compatibility libraries.
+	//
+	// Note: Since this mechanism replaces the logic in the `nvidia-container-cli`,
+	// toggling this feature has no effect if `allow-cuda-compat-libs-from-container` is enabled.
+	DisableCUDACompatLibHook *feature `toml:"disable-cuda-compat-lib-hook,omitempty"`
```
ssst0n3, line 33: This feature flag seems not used in the cdi mode. Is this expected?

pkg/nvcdi/driver-nvml.go
```diff
@@ -97,6 +97,8 @@ func NewDriverLibraryDiscoverer(logger logger.Interface, driver *root.Driver, nv
 		libraryPaths,
 	)
 
+	// TODO: The following should use the version directly.
+	cudaCompatLibHookDiscoverer := discover.NewCUDACompatHookDiscoverer(logger, nvidiaCDIHookPath, driver)
```
ssst0n3, line 101: Should the `disable-cuda-compat-lib-hook` feature flag  be used here?

PR #915, opened 2025-02-10
[no-relnote] Configure all GitHub actions as reusable workflow

ArangoGutierrez:


.github/workflows/code_scanning.yaml
```diff
@@ -15,17 +15,7 @@
 name: "CodeQL"
 
 on:
-  pull_request:
-    types:
-      - opened
-      - synchronize
-    branches:
-      - main
-      - release-*
```
elezar, line 24: Is there any reason to not use the "standard" triggers for this?

.github/workflows/e2e.yaml
```diff
@@ -67,8 +73,8 @@ jobs:
 
       - name: Run e2e tests
         env:
-          IMAGE_NAME: ghcr.io/${LOWERCASE_REPO_OWNER}/container-toolkit
-          VERSION: ${COMMIT_SHORT_SHA}
+          IMAGE_NAME: ghcr.io/${{ inputs.lowercase_repo_owner }}/container-toolkit
```
elezar, line 76: Question: Isn't this *always* `nvidia`?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,79 @@
...
+          if [[ "${{ github.ref_name != 'main' && !startsWith(github.ref_name, 'release-') }}" == "true" ]]; then
+            echo "build_multi_arch_images=false" >> $GITHUB_OUTPUT
+          else
+            echo "build_multi_arch_images=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Set Docker Build Arguments
+        id: docker_build_args
+        run: |
+          echo "push_on_build=true" >> $GITHUB_OUTPUT
```
elezar, line 57: This is now always true.

.github/workflows/image.yaml
```diff
@@ -84,18 +93,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         name: Check out code
```
elezar, line 95: What version is checked out here? Should we align this with the short sha?

.github/workflows/code_scanning.yaml
ArangoGutierrez: I think is a good first step to run the code scan before performing any extra steps.

.github/workflows/image.yaml
ArangoGutierrez: would make sense in a follow-up PR. The defaults are ok, but moving forward when writing more reusable workflows, it will make more sense 

https://github.com/actions/checkout?tab=readme-ov-file#usage

.github/workflows/ci.yaml
ArangoGutierrez: done

.github/workflows/image.yaml
```diff
@@ -16,11 +16,20 @@
...
-  push:
-    branches:
-      - "pull-request/[0-9]+"
-      - main
-      - release-*
+  workflow_call:
+    inputs:
+      commit_short_sha:
+        required: true
+        type: string
```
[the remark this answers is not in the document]
ArangoGutierrez: done

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      build_multi_arch_images: ${{ steps.build_multi_arch_images.outputs.value }}
+    steps:
+      - name: Generate Commit Short SHA
+        id: version
+        run: echo "version=$(echo $GITHUB_SHA | cut -c1-8)" >> "$GITHUB_OUTPUT"
+
+      - name: Convert Repository Owner to Lowercase
+        id: lowercase_repo_owner
```
elezar, line 40: Is this ever used? I thought we said that this is always `nvidia`?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+        run: echo "lowercase_repo_owner=$(echo "${GITHUB_REPOSITORY_OWNER}" | awk '{print tolower($0)}')" >> $GITHUB_OUTPUT
+
+      - name: Determine Multi-Arch Build Setting
+        id: build_multi_arch_images
+        run: |
+          if [[ "${{ github.ref_name != 'main' && !startsWith(github.ref_name, 'release-') }}" == "true" ]]; then
+            echo "build_multi_arch_images=false" >> $GITHUB_OUTPUT
+          else
+            echo "build_multi_arch_images=true" >> $GITHUB_OUTPUT
+          fi
```
elezar, line 50: It may be clearer to invert the conditional:
```suggestion
          if [[ "${{ github.ref_name == 'main' || startsWith(github.ref_name, 'release-') }}" == "true" ]]; then
            echo "build_multi_arch_images=true" >> $GITHUB_OUTPUT
          else
            echo "build_multi_arch_images=false" >> $GITHUB_OUTPUT
          fi
```

Actually, could we simplify this further to:
```suggestion
          multi_arch_required="${{ github.ref_name == 'main' || startsWith(github.ref_name, 'release-') }}" 
          echo "build_multi_arch_images=${{ multi_arch_required }}" >> $GITHUB_OUTPUT          
```

.github/workflows/ci.yaml
ArangoGutierrez: my bad deleted almost everything about this, forgot this bit

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,66 @@
...
+jobs:
+  code-scanning:
+    uses: ./.github/workflows/code_scanning.yaml
+
+  prepare-variables:
+    needs: code-scanning
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      build_multi_arch_images: ${{ steps.build_multi_arch_images.outputs.value }}
```
elezar, line 33: Could this just be:
```suggestion
      build_multi_arch_images: ${{ github.ref_name == 'main' || startsWith(github.ref_name, 'release-') }}
```
and then we don't need the scripts below?

.github/workflows/image.yaml
elezar: Shure. Could you please clarify which verison is checked out by default.

.github/workflows/image.yaml
```diff
@@ -115,16 +117,13 @@ jobs:
...
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Build image
         env:
-          IMAGE_NAME: ghcr.io/${LOWERCASE_REPO_OWNER}/container-toolkit
-          VERSION: ${COMMIT_SHORT_SHA}
+          IMAGE_NAME: ghcr.io/nvidia/container-toolkit
+          VERSION: ${{ inputs.version }}
+          PUSH_ON_BUILD: ${{ inputs.push_on_build }}
```
elezar, line 125: What about?
```suggestion
          PUSH_ON_BUILD: true
```

.github/workflows/image.yaml
```diff
@@ -16,11 +16,17 @@
...
+    inputs:
+      version:
+        required: true
+        type: string
+      build_multi_arch_images:
+        required: true
+        type: string
+      push_on_build:
+        required: true
+        type: string
```
elezar, line 29: This is always `true`. Could we remove it?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,63 @@
...
+        run: echo "version=$(echo $GITHUB_SHA | cut -c1-8)" >> "$GITHUB_OUTPUT"
+
+      - name: Determine Multi-Arch Build Setting
+        id: build_multi_arch_images
+        run: |
+          multi_arch_required="${{ github.ref_name == 'main' || startsWith(github.ref_name, 'release-') }}" 
+          echo "build_multi_arch_images=$multi_arch_required" >> $GITHUB_OUTPUT  
+
+  golang:
+    needs: code-scanning
```
elezar, line 46: What is the risk in running the `golang` tests and the `code-scanning` in parallel?

.github/workflows/image.yaml
ArangoGutierrez: I was going for reusability of the workflow file, but default to true is ok

.github/workflows/image.yaml
ArangoGutierrez: yes

.github/workflows/ci.yaml
ArangoGutierrez: none. let me re-arrange

.github/workflows/image.yaml
ArangoGutierrez:
```yaml
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
```

.github/workflows/ci.yaml
ArangoGutierrez: image

ArangoGutierrez: now we run the basic checks in parallel

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - release-*
+
+jobs:
+  code-scanning:
+    uses: ./.github/workflows/code_scanning.yaml
+
+  prepare-variables:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
```
elezar, line 31: Does the following work?
```suggestion
      version: ${{ github.sha.substring(0,8) }}
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,53 @@
...
+    branches:
+      - "pull-request/[0-9]+"
+      - main
+      - release-*
+
+jobs:
+  code-scanning:
+    uses: ./.github/workflows/code_scanning.yaml
+
+  prepare-variables:
```
elezar, line 28: Does it make sense to just call this `variables`?

.github/workflows/code_scanning.yaml
elezar: That wasn't my question. Does it make sense to *also* run this on PRs?

.github/workflows/e2e.yaml
```diff
@@ -67,8 +70,8 @@ jobs:
 
       - name: Run e2e tests
         env:
-          IMAGE_NAME: ghcr.io/${LOWERCASE_REPO_OWNER}/container-toolkit
-          VERSION: ${COMMIT_SHORT_SHA}
+          IMAGE_NAME: ghcr.io/nvidia/container-toolkit
+          VERSION: ${{ steps.vars.outputs.version }}
```
elezar, line 74:
```suggestion
          VERSION: ${{ inputs.version }}
```

.github/workflows/e2e.yaml
```diff
@@ -67,8 +70,8 @@ jobs:
 
       - name: Run e2e tests
         env:
-          IMAGE_NAME: ghcr.io/${LOWERCASE_REPO_OWNER}/container-toolkit
-          VERSION: ${COMMIT_SHORT_SHA}
+          IMAGE_NAME: ghcr.io/nvidia/container-toolkit
```
elezar, line 73: If we do move `e2e.yaml` to a central repo, we would have to have this as an input.

(Out of scope for this PR).

.github/workflows/image.yaml
elezar: If needed, we can add `push_on_build` to the `inputs` *and* set a default value.

.github/workflows/ci.yaml
ArangoGutierrez: no, it doesn't

.github/workflows/e2e.yaml
ArangoGutierrez: agree

.github/workflows/ci.yaml
elezar: As a matter of interest, what error does it give? Do we have a reference of which functions are available to us? We do use `startsWith`, for example.

elezar:
If I recall correctly, the action results are tracked by SHA and I was wondering whether it makes sense to trigger the `CodeQL` and `Golang` checks on PRs as well? The motivation being that these would provide early feedback to contributors without requireing an `/ok-to-test` to trigger the PR copy.

.github/workflows/code_scanning.yaml
```diff
@@ -15,17 +15,7 @@
...
-  push:
-    branches:
-      - main
-      - release-*
+  workflow_call:
```
elezar, line 18: Given that we want these to run on PRs so that they trigger for dependabot we may want to update this to:
```suggestion
  workflow_call: {}
  pull_request:
    types:
      - opened
      - synchronize
    branches:
      - main
      - release-*
```

.github/workflows/golang.yaml
```diff
@@ -15,17 +15,7 @@
...
-      - opened
-      - synchronize
-    branches:
-      - main
-      - release-*
-  push:
-    branches:
-      - main
-      - release-*
+  workflow_call:
```
elezar, line 18: Given that we want these to run on PRs so that they trigger for dependabot we may want to update this to:
```suggestion
  workflow_call: {}
  pull_request:
    types:
      - opened
      - synchronize
    branches:
      - main
      - release-*
```

.github/workflows/image.yaml
ArangoGutierrez: is the SHA of the commit that triggered the action, unregarding of the branch (that's why copy-pr-bot works)

elezar (approved):
Thanks @ArangoGutierrez

Looks good.

PR #927, opened 2025-02-13
Disable device node creation in CDI mode

elezar:
This change adds a hook to disable device node creation for FULL GPUs (i.e. non-MIG devices) or modification in a container by updating the `ModifyDeviceFiles` driver parameter.

(This does not include `nvidia-caps` devices that are required by MIG devices).

The presence of "extra" device nodes in a container are largely cosmetic since the container *should* not have the required cgroup access for the additional devices. This does not affect the device nodes on the host.

Without this change running a command like `nvidia-smi` in a creates the device nodes as follows:
```
elezar@dgx0126:~/src/container-toolkit$ docker run --rm -ti -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=0 --runtime=nvidia ubuntu bash -c "ls -al /dev/nvidia*; echo ""; nvidia-smi -L; echo ""; ls -al /dev/nvidia*"                                                                                                                                                                                            
crw-rw-rw- 1 root root 195, 254 Feb 13 14:29 /dev/nvidia-modeset
crw-rw-rw- 1 root root 511,   0 Feb 13 14:29 /dev/nvidia-uvm
crw-rw-rw- 1 root root 511,   1 Feb 13 14:29 /dev/nvidia-uvm-tools
crw-rw-rw- 1 root root 195,   0 Feb 13 14:29 /dev/nvidia0
crw-rw-rw- 1 root root 195, 255 Feb 13 14:29 /dev/nvidiactl

GPU 0: Tesla V100-SXM2-16GB-N (UUID: GPU-edfee158-11c1-52b8-0517-92f30e7fac88)

crw-rw-rw- 1 root root 195, 254 Feb 13 14:29 /dev/nvidia-modeset
crw-rw-rw- 1 root root 511,   0 Feb 13 14:29 /dev/nvidia-uvm
crw-rw-rw- 1 root root 511,   1 Feb 13 14:29 /dev/nvidia-uvm-tools
crw-rw-rw- 1 root root 195,   0 Feb 13 14:29 /dev/nvidia0
crw-rw-rw- 1 root root 195,   1 Feb 13 14:29 /dev/nvidia1
crw-rw-rw- 1 root root 195,   2 Feb 13 14:29 /dev/nvidia2
crw-rw-rw- 1 root root 195,   3 Feb 13 14:29 /dev/nvidia3
crw-rw-rw- 1 root root 195,   4 Feb 13 14:29 /dev/nvidia4
crw-rw-rw- 1 root root 195,   5 Feb 13 14:29 /dev/nvidia5
crw-rw-rw- 1 root root 195,   6 Feb 13 14:29 /dev/nvidia6
crw-rw-rw- 1 root root 195,   7 Feb 13 14:29 /dev/nvidia7
crw-rw-rw- 1 root root 195, 255 Feb 13 14:29 /dev/nvidiactl

/dev/nvidia-caps:
total 0
drwxr-xr-x 2 root root     80 Feb 13 14:29 .
drwxr-xr-x 7 root root    640 Feb 13 14:29 ..
cr-------- 1 root root 238, 1 Feb 13 14:29 nvidia-cap1
cr--r--r-- 1 root root 238, 2 Feb 13 14:29 nvidia-cap2
```

With the change applied we see:
```
elezar@dgx0126:~/src/container-toolkit$ docker run --rm -ti -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=0 --runtime=nvidia ubuntu bash -c "ls -al /dev/nvidia*; echo ""; nvidia-smi -L; echo ""; ls -al /dev/nvidia*"
crw-rw-rw- 1 root root 195, 254 Feb 13 14:27 /dev/nvidia-modeset
crw-rw-rw- 1 root root 511,   0 Feb 13 14:27 /dev/nvidia-uvm
crw-rw-rw- 1 root root 511,   1 Feb 13 14:27 /dev/nvidia-uvm-tools
crw-rw-rw- 1 root root 195,   0 Feb 13 14:27 /dev/nvidia0
crw-rw-rw- 1 root root 195, 255 Feb 13 14:27 /dev/nvidiactl

GPU 0: Tesla V100-SXM2-16GB-N (UUID: GPU-edfee158-11c1-52b8-0517-92f30e7fac88)

crw-rw-rw- 1 root root 195, 254 Feb 13 14:27 /dev/nvidia-modeset
crw-rw-rw- 1 root root 511,   0 Feb 13 14:27 /dev/nvidia-uvm
crw-rw-rw- 1 root root 511,   1 Feb 13 14:27 /dev/nvidia-uvm-tools
crw-rw-rw- 1 root root 195,   0 Feb 13 14:27 /dev/nvidia0
crw-rw-rw- 1 root root 195, 255 Feb 13 14:27 /dev/nvidiactl

/dev/nvidia-caps:
total 0
drwxr-xr-x 2 root root     80 Feb 13 14:27 .
drwxr-xr-x 7 root root    500 Feb 13 14:27 ..
cr-------- 1 root root 238, 1 Feb 13 14:27 nvidia-cap1
cr--r--r-- 1 root root 238, 2 Feb 13 14:27 nvidia-cap2
```

due to the parameter being updated accordingly:
```
elezar@dgx0126:~/src/container-toolkit$ docker run --rm -ti -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=0 --runtime=nvidia ubuntu bash -c "cat /proc/driver/nvidia/params | grep ModifyDeviceFiles"
ModifyDeviceFiles: 0
```

### DIFF
--- a/cmd/nvidia-cdi-hook/disable-device-node-modification/disable-device-node-modification.go
+++ b/cmd/nvidia-cdi-hook/disable-device-node-modification/disable-device-node-modification.go
@@ -1,0 +1,144 @@
+/**
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package disabledevicenodemodification
+
+import (
+	"bufio"
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/oci"
+)
+
+const (
+	nvidiaDriverParamsPath = "/proc/driver/nvidia/params"
+)
+
+type options struct {
+	containerSpec string
+}
+
+// NewCommand constructs an disable-device-node-modification subcommand with the specified logger
+func NewCommand(logger logger.Interface) *cli.Command {
+	cfg := options{}
+
+	c := cli.Command{
+		Name:  "disable-device-node-modification",
+		Usage: "Ensure that the /proc/driver/nvidia/params file present in the container does not allow device node modifications.",
+		Before: func(c *cli.Context) error {
+			return validateFlags(c, &cfg)
+		},
+		Action: func(c *cli.Context) error {
+			return run(c, &cfg)
+		},
+	}
+
+	c.Flags = []cli.Flag{
+		&cli.StringFlag{
+			Name:        "container-spec",
+			Hidden:      true,
+			Usage:       "Specify the path to the OCI container spec. If empty or '-' the spec will be read from STDIN",
+			Destination: &cfg.containerSpec,
+		},
+	}
+
+	return &c
+}
+
+func validateFlags(c *cli.Context, cfg *options) error {
+	return nil
+}
+
+func run(_ *cli.Context, cfg *options) error {
+	modifiedParamsFileContents, err := getModifiedNVIDIAParamsContents()
+	if err != nil {
+		return fmt.Errorf("failed to get modified params file contents: %w", err)
+	}
+	if len(modifiedParamsFileContents) == 0 {
+		return nil
+	}
+
+	s, err := oci.LoadContainerState(cfg.containerSpec)
+	if err != nil {
+		return fmt.Errorf("failed to load container state: %w", err)
+	}
+
+	containerRootDirPath, err := s.GetContainerRoot()
+	if err != nil {
+		return fmt.Errorf("failed to determined container root: %w", err)
+	}
+
+	return createParamsFileInContainer(containerRootDirPath, modifiedParamsFileContents)
+}
+
+func getModifiedNVIDIAParamsContents() ([]byte, error) {
+	hostNvidiaParamsFile, err := os.Open(nvidiaDriverParamsPath)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to load params file: %w", err)
+	}
+	defer hostNvidiaParamsFile.Close()
+
+	modifiedContents, err := getModifiedParamsFileContentsFromReader(hostNvidiaParamsFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get modfied params file contents: %w", err)
+	}
+
+	return modifiedContents, nil
+}
+
+// getModifiedParamsFileContentsFromReader returns the contents of a modified params file from the specified reader.
+func getModifiedParamsFileContentsFromReader(r io.Reader) ([]byte, error) {
+	var modified bytes.Buffer
+	scanner := bufio.NewScanner(r)
+
+	var requiresModification bool
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "ModifyDeviceFiles: ") {
+			if line == "ModifyDeviceFiles: 0" {
+				return nil, nil
+			}
+			if line == "ModifyDeviceFiles: 1" {
+				line = "ModifyDeviceFiles: 0"
+				requiresModification = true
+			}
+		}
+		if _, err := modified.WriteString(line + "\n"); err != nil {
+			return nil, fmt.Errorf("failed to create output buffer: %w", err)
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("failed to read params file: %w", err)
+	}
+
+	if !requiresModification {
+		return nil, nil
+	}
+
+	return modified.Bytes(), nil
+}
--- a/cmd/nvidia-cdi-hook/commands/commands.go
+++ b/cmd/nvidia-cdi-hook/commands/commands.go
@@ -22,6 +22,7 @@ import (
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-cdi-hook/chmod"
 	symlinks "github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-cdi-hook/create-symlinks"
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-cdi-hook/cudacompat"
+	disabledevicenodemodification "github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-cdi-hook/disable-device-node-modification"
 	ldcache "github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-cdi-hook/update-ldcache"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
 )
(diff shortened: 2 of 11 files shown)


cmd/nvidia-cdi-hook/update-nvidia-params/update-nvidia-params.go
```diff
@@ -0,0 +1,199 @@
...
+	return c.build()
+}
+
+// build the update-nvidia-params command
+func (m command) build() *cli.Command {
+	cfg := options{}
+
+	// Create the 'update-nvidia-params' command
+	c := cli.Command{
+		Name:  "update-nvidia-params",
```
elezar, line 61: I think we should call this `disable-device-node-modification` or something more descriptive of the intent.

cc @klueska what do you think?

jgehrcke:
> for FULL GPUs

What does "full" mean here? Fully occupied? Entire?

jgehrcke:
> Without this change running a command like nvidia-smi in a creates the device nodes as follows:

Thanks for the detailed output. I wonder: why is the creation of those device nodes a problem? Can you add that to the PR description? It's probably a simple answer, for me it's not obvious at all. Is this maybe cosmetic? Does the modification also apply on the host, or it only 'visible' in the container? That is, does running nvidia-smi _in_ the container mutate host state? Is that something we want to prevent from happening?

cmd/nvidia-cdi-hook/update-nvidia-params/update-nvidia-params.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// getModifiedParamsFileContentsFromReader returns the contents of a modified params file from the specified reader.
+func (m command) getModifiedParamsFileContentsFromReader(r io.Reader) ([]byte, error) {
+	var modified bytes.Buffer
+	scanner := bufio.NewScanner(r)
+
+	var requiresModification bool
+	for scanner.Scan() {
+		line := scanner.Text()
```
jgehrcke, line 130: One of our favorite topics. As I learn about Golang, I naturally got curious here. Text processing and ingest or emission of 'text data' is always a fun topic. Because typically assumptions about text encodings are built into code, and those assumptions are often not obvious.

So, I just tried to understand what's happening here.

The scanner operates on a raw byte stream. Here we do the magic conversion to something called text. Text typically means: a sequence of abstract characters (unicode code points). 

Going from a sequence of bytes to a sequence of unicode code points is called decoding, and it assumes a specific codec.

Which codec is used here?

Or `Text()` is a bad naming choice in bufio and what's happening here is just splitting/tokenization of a byte sequence (output: sequence of byte sequences).

But then we compare with string literals below, as in e.g. `strings.HasPrefix(line, "ModifyDeviceFiles: ")`.

To be precise, this tries to prefix-match the bytes in `line` with the _bytes_ underlying the string literal. String literals in Golang are stored as bytes using the UTF-8 code:

>  string literals always contain UTF-8 text as long as they have no byte-level escapes.

(source: https://go.dev/blog/strings)

And that's how this code assumes that the byte sequence entering the system represents text encoded with UTF-8 (or a subset thereof, which ascii is). And that is typically a valid assumption :).

cmd/nvidia-cdi-hook/update-nvidia-params/update-nvidia-params.go
```diff
@@ -0,0 +1,199 @@
...
+		return fmt.Errorf("failed to create temporary parms file mount: %w", err)
+	}
+
+	return nil
+}
+
+// createFileInTempfs creates a file with the specified name, contents, and mode in a tmpfs.
+// A tmpfs is created at /tmp/nvct-emtpy-dir* with a size sufficient for the specified contents.
+func createFileInTempfs(name string, contents []byte, mode os.FileMode) (string, error) {
+	tmpRoot, err := os.MkdirTemp("", "nvct-empty-dir*")
```
jgehrcke, line 176: Not sure I understand the naming choice "nvct-empty-dir". Is this always empty? :)

Does this generate a directory visible on the host filesystem when I do `ls /tmp`?

jgehrcke (approved):
Thank you, Evan! 

Given the timeline and goal that we have I think we should land this asap. I have only looked at this very high-level and left high-level questions. I trust that code is good enough for a first release. :rocket: Please merge at your own discretion (from my point of view).

cmd/nvidia-cdi-hook/update-nvidia-params/update-nvidia-params.go
cdesiniotis: Alternatively, I would also be in favor of keeping the `update-nvidia-params` name and adding subcommands or command-line-flags specific to the `update-nvidia-params` command that indicate which parameter in the params file we are updating. This way, if we ever need to update other parameters we have a natural way to express that.

cmd/nvidia-cdi-hook/update-nvidia-params/update-nvidia-params.go
cdesiniotis: > Does this generate a directory visible on the host filesystem when I do ls /tmp?

A direct quote from Evan:

"Note that since the hook is run as a createContainer hook, the mount operations are performed in the mount namespace of the container. This means that the mounts are not visible from the host."

cmd/nvidia-cdi-hook/update-nvidia-params/update-nvidia-params.go
elezar: I don't want to have multi-modal hooks. We already have a two or three level nesting from the `nvidia-cdi-hook` or `nvidia-ctk hook` commands.

elezar:
> > for FULL GPUs
> 
> What does "full" mean here? Fully occupied? Entire?

We use "FULL GPU" do refer to a GPU that has NOT been partitioned using MIG.

cmd/nvidia-cdi-hook/update-nvidia-params/update-nvidia-params.go
elezar: Yes, this directory is always empty. The `empty-dir` inffix is something that I have seen in the context of k8s, but I would have to dig for the exact source. (on a system where I was testing this we have around 42 of these folders).

In the current implementation this directory is created at `/tmp/nvct-empty-dir*` *on the host* meaning that it is visible on the host when you run `ls /tmp`:
```
elezar@dgx0126:~$ ls /tmp/nvct-empty-dir*
ls: cannot open directory '/tmp/nvct-empty-dir230315467': Permission denied
elezar@dgx0126:~$ sudo ls /tmp/nvct-empty-dir*
elezar@dgx0126:~$
```
As seen above, this directory is not accessible to regular users and also does not contain the `nvct-params` file that we create in the tmpfs.

The mount operation that follows, however, creates the tmpfs mount in the container since this hook is running in the container's namespace:
```
root@8cda1ee3028a:/# mount | grep params
tmpfs on /proc/driver/nvidia/params type tmpfs (rw,relatime,size=4k)
```

cmd/nvidia-cdi-hook/disable-device-node-modification/disable-device-node-modification.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return &c
+}
+
+func (m command) validateFlags(c *cli.Context, cfg *options) error {
+	return nil
+}
+
+func (m command) run(c *cli.Context, cfg *options) error {
+	// TODO: Do we need to prefix the driver root?
```
ArangoGutierrez, line 87: Is my understanding that the NVIDIA `/proc` filesystem is locked, so hardcoding it to the var `nvidiaDriverParamsPath` looks good and safe

ArangoGutierrez (approved):
LGTM
Just one comment.

elezar:
@klueska one think I was considering here, was whether we want to make this more general purpose than `disable-device-node-creation`? Maybe `update-nvidia-proc` or something so that we can also use this to perform functions like we have for `nvkind`?

klueska:
> @klueska one think I was considering here, was whether we want to make this more general purpose than `disable-device-node-creation`? Maybe `update-nvidia-proc` or something so that we can also use this to perform functions like we have for `nvkind`?

Maybe .... would it then become something like:
```
update-nvidia-proc --disable-device-node-creation ...
```

elezar:
> > @klueska one think I was considering here, was whether we want to make this more general purpose than `disable-device-node-creation`? Maybe `update-nvidia-proc` or something so that we can also use this to perform functions like we have for `nvkind`?
> 
> Maybe .... would it then become something like:
> 
> ```
> update-nvidia-proc --disable-device-node-creation ...
> ```

I'm sticking to a single hook. We can reevaluate as part of the rc.

PR #935, opened 2025-02-21
Disable nvsandboxutils in nvcdi API

elezar:
Repeated calls to nvsandboxutils.Init and Shutdown are causing segmentation violations. Here we disabled nvsandbox utils unless explicitly specified.

We will reenable the interaction with this library once the required checks and tests are in place to ensure that we don't introduce breakages.

This was tested by updating `nvidia-ctk cdi generate` to allow for multiple modes and validating that removing this initialisation fixes the failure.

### DIFF
--- a/pkg/nvcdi/lib.go
+++ b/pkg/nvcdi/lib.go
@@ -99,19 +99,24 @@ func New(opts ...Option) (Interface, error) {
 		}
 		l.nvmllib = nvml.New(nvmlOpts...)
 	}
-	if l.nvsandboxutilslib == nil {
-		var nvsandboxutilsOpts []nvsandboxutils.LibraryOption
-		// Set the library path for libnvidia-sandboxutils
-		candidates, err := l.driver.Libraries().Locate("libnvidia-sandboxutils.so.1")
-		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/nvcdi/lib.go
```diff
@@ -99,19 +99,22 @@ func New(opts ...Option) (Interface, error) {
...
-		if err != nil {
-			l.logger.Warningf("Ignoring error in locating libnvidia-sandboxutils.so.1: %v", err)
-		} else {
-			libNvidiaSandboxutilsPath := candidates[0]
-			l.logger.Infof("Using %v", libNvidiaSandboxutilsPath)
-			nvsandboxutilsOpts = append(nvsandboxutilsOpts, nvsandboxutils.WithLibraryPath(libNvidiaSandboxutilsPath))
-		}
-		l.nvsandboxutilslib = nvsandboxutils.New(nvsandboxutilsOpts...)
-	}
+	// TODO: Repeated calls to nvsandboxutils.Init and Shutdown are causing
```
ArangoGutierrez, line 102: `TODO` message points to why we are disabling this, not to a goal or road map idea on how to fix/enhance the issue.

elezar:
> Non blocking comment

Does that mean that you approve?

ArangoGutierrez (approved):

PR #941, opened 2025-02-25
Use memfd when running ldconfig

elezar:
This change copies ldconfig into a memfd before executing it from the createContainer hook.

See https://github.com/NVIDIA/libnvidia-container/commit/ff83efb92cde1f3211b097dc6d1e6c0b921e66a2 where this was done for libnvidia-container.


cmd/nvidia-cdi-hook/update-ldcache/container-root.go
klueska, line 1: This commit (`[no-relnote] Move root to separate file`) doesn't seem to do what it says -- the next commit is the one that introduces a new file for all the root stuff

elezar: I've updated that.

cdesiniotis (approved):

PR #943, opened 2025-02-26
Add ignore-imex-channel-requests feature flag

elezar:
This allows the NVIDIA Container Toolkit to ignore IMEX channel requests through the NVIDIA_IMEX_CHANNELS envvar or volume mounts and ensures that the NVIDIA Container Toolkit cannot be used to provide out-of-band access to an IMEX channel by simply specifying an environment variable, possibly bypassing other checks by an orchestration system such as kubernetes.

To enable this feature add the following to the `config.toml`:
```
[features]
ignore-imex-channel-requests = true
```
This can also be done by running:
```
sudo nvidia-ctk config --set features.ignore-imex-channel-requests
```

In the case of the toolkit container ensure that the `NVIDIA_CONTAINER_TOOLKIT_OPT_IN_FEATURES` includes `ignore-imex-channel-requests`.

### DIFF
--- a/cmd/nvidia-container-runtime-hook/container_config.go
+++ b/cmd/nvidia-container-runtime-hook/container_config.go
@@ -198,6 +198,10 @@ func getMigDevices(image image.CUDA, envvar string) *string {
 }
 
 func (hookConfig *hookConfig) getImexChannels(image image.CUDA, privileged bool) []string {
+	if hookConfig.Features.IgnoreImexChannelRequests.IsEnabled() {
+		return nil
+	}
+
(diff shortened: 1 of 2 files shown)


jgehrcke (commented):
Nice.

This needs to be explicitly opted in? How can this be enabled? (for a newcomer like me this is not obvious!). Maybe you can clarify that in the PR descripion.

I think for our future selves we should also maybe add a code comment about the choice and relevance of the default behavior (default "insecure" because ... or default "secure" because ...). Probably using the trade-off between backwards compat and security (which is typically a bit of a mean trade-off to make).

PR #947, opened 2025-02-27
Ensure that libcuda.so is in the ldcache

elezar:


tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -166,4 +167,29 @@ var _ = Describe("docker", Ordered, func() {
...
+		})
+
+		It("should include libcuda.so in the ldcache", func(ctx context.Context) {
+			ldcacheOutput, _, err := r.Run("docker run --rm -ti --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=all ubuntu bash -c \"ldconfig -p | grep 'libcuda.so'\"")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ldcacheOutput).ToNot(BeEmpty())
+
+			ldcacheLines := strings.Split(ldcacheOutput, "\n")
+			Expect(ldcacheLines).To(HaveLen(2))
+
```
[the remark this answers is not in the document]
elezar: Thanks. The command that is executed to capture these lines is tailored to always capture 2 lines. If this is not the case, then it's already a failure.  Leaving as is.

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
```diff
@@ -0,0 +1,139 @@
+/**
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
```
ArangoGutierrez, line 2: 2025?

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -166,4 +167,29 @@ var _ = Describe("docker", Ordered, func() {
...
 	})
+
+	When("A container is run using CDI", Ordered, func() {
+		BeforeAll(func(ctx context.Context) {
+			_, _, err := r.Run("docker pull ubuntu")
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should include libcuda.so in the ldcache", func(ctx context.Context) {
+			ldcacheOutput, _, err := r.Run("docker run --rm -ti --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=all ubuntu bash -c \"ldconfig -p | grep 'libcuda.so'\"")
```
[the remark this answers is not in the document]
elezar: I have updated this to use only `-i`. Note that we do need to capture the stdout in this command.

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
```diff
@@ -0,0 +1,139 @@
...
+	return c.build()
+}
+
+// build the create-soname-symlinks command
+func (m command) build() *cli.Command {
+	cfg := options{}
+
+	// Create the 'create-soname-symlinks' command
+	c := cli.Command{
+		Name:  "create-soname-symlinks",
```
elezar, line 57: @klueska I elected to add a new hook entirely instead of modifying the existing `update-ldcache`. This is in keeping with "purpose-built hooks" and also means that the hook name can be used to indicate the intent.

ArangoGutierrez (commented):
lgtm, just one nit

ArangoGutierrez (approved):

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
```diff
@@ -0,0 +1,131 @@
...
+	args = append(args,
+		// Specify the containerRoot to use.
+		"-r", containerRoot,
+		// Specify -n to only process the specified folders.
+		"-n",
+		// Explicitly disable updating the LDCache.
+		"-N",
+	)
+	// Explicitly specific the directories to add.
+	args = append(args, dirs...)
```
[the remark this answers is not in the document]
elezar: This is not the `.so` symlinks, but the `SONAME` symlinks i.e. `libcuda.so.1` -> `libcuda.so.RM_VERSION` in the case of `libcuda`. The `.so` symlinks are created using the "standard" `create-symlinks` hook.

elezar:
@cdesiniotis I am removing the must-backport label for this. Although this was triggered by a customer request, there is a workaround available and I would rather not backport another new hook to the `release-1.17` branch.

jgehrcke:
>  and I would rather not backport another new hook to the release-1.17 branch.

Sounds good.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -107,33 +107,34 @@ func (m command) run(_ *cli.Context, cfg *options) error {
...
 }
 
-func (m command) getContainerForwardCompatDir(containerRoot containerRoot, hostDriverVersion string) (string, error) {
+func (m command) getContainerForwardCompatDir(containerRoot utils.ContainerRoot, hostDriverVersion string) (string, error) {
 	if hostDriverVersion == "" {
 		m.logger.Debugf("Host driver version not specified")
 		return "", nil
 	}
-	if !containerRoot.hasPath(cudaCompatPath) {
+	if !containerRoot.HasPath(cudaCompatPath) {
```
jgehrcke, line 128: I have picked a more or less arbitrary place in the code to put the following comment. Having more expressive variable names would make review for me easier. You can postpone looking at this comment until after merge, that's fine with me! In the spirit of learning what we do here however I'd appreciate an answer at some point :)

In my software engineering life I have always deeply cared about file system terminology and operations, and about making related code readable and self-expressive. For example,

- I like to distinguish a "relative path" from an "absolute path" via variable name (if possible)
- I like to distinguish a file object from a file path via variable name
- I like to distinguish a path to a file from a path to a directory (expressing intent, of course these are technically the same)
- I like to call things a "base name" for expressing intent a well (when there are no path separators, and this is supposed to be a "file name" or "directory name").

What is `containerRoot`  in canonical unix file path terminology?

- Is it guaranteed to point to a directory?
- Is it an absolute path?
- Is it always just `/`? 

What are some properties about `cudaCompatPath` that we know/guarantee, that we could also express in the variable name or in the type?

- Is it always a relative path (not starting with `/)?
- Is it always a path to a directory?

cmd/nvidia-cdi-hook/utils/container-root.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/moby/sys/symlink"
+)
+
+// A ContainerRoot represents the root filesystem of a container.
+type ContainerRoot string
```
jgehrcke, line 28: Oh, it's just a string.

Relates to my previous comment: I'd love to see the intent expressed in the variable name: path to a directory?

And this is probably really just my lack of knowledge: I wonder: when is the container's root file system _not_ located at just `/`?

jgehrcke (approved):
Thanks for all the work! I understand very little but you have my full emotional support! And I know you will repair everything you destroy. In that sense: approved with honor. :rocket:

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: Thanks for calling this out. I think we could do a pass through this and improve readability greatly. There is no rush on this particular PR (I've elected to hold the changes back from the v1.17.5 release) and as such, I think we can address these concerns here instead of as a follow-up.

The `containerRootDir` is the absolute path on the host filesystem to the root (`/`) of the container filesystem. It is specified in the OCI Runtime Specification as [`Root`](https://github.com/opencontainers/runtime-spec/blob/main/config.md#root) which informs the slightly ambiguous name.  The `containerRoot` variable is the typed representation of this directory that allows us to attached helper functions such as `HasPath` / `Resolve` / `Glob` to it.

The `cudaCompatPath` is the absolute path to the directory containing the CUDA compat libraries in the container (if it exists). It defined as a [*constant* with value `/usr/local/cuda/compat`](https://github.com/NVIDIA/nvidia-container-toolkit/pull/947/files#diff-7ac7f6fb88a2b00361d500447c04582f3c805bd729238b9bfaeaf24cbccc0487R33). That is to say, it is always an absolute path to a directory in the container, and we're confirming that this exists, but have to calculate the path to this folder on the host filesystem to do this check.

cmd/nvidia-cdi-hook/utils/container-root.go
elezar: I have added som context above. Let me know if it's not clear and I can add more information.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
jgehrcke: >  I think we can address these concerns here instead of as a follow-up.

Thank you for that!

jgehrcke: > The containerRootDir is the absolute path on the host filesystem to the root (/) of the container filesystem

Is the container filesystem always accessible from within the host filesystem?

> It is specified in the OCI Runtime Specification as [Root](https://github.com/opencontainers/runtime-spec/blob/main/config.md#root) which informs the slightly ambiguous name.

gotcha. Thanks for that background.

> The cudaCompatPath is the absolute path to the directory containing the CUDA compat libraries in the container (if it exists).

Thank you for that precision. Now, that is quickly&easily understandable.

> but have to calculate the path to this folder on the host filesystem to do this check

meaning that it's always mounted _into_ the container, and never baked into it?

cmd/nvidia-cdi-hook/chmod/chmod.go
```diff
@@ -113,15 +113,15 @@ func (m command) run(c *cli.Context, cfg *config) error {
 		return fmt.Errorf("failed to load container state: %v", err)
 	}
 
-	containerRoot, err := s.GetContainerRoot()
+	containerRoot, err := s.GetContainerRootDirPath()
```
jgehrcke, line 116: :heart:  Thanks for adding clarity here.

And now that I have read your explanation.. this is a path that is valid in the _host_ filesystem, and points to the container filesystem's root..? :)

cmd/nvidia-cdi-hook/chmod/chmod.go
```diff
@@ -140,6 +140,7 @@ func (m command) run(c *cli.Context, cfg *config) error {
 }
 
 // getPaths updates the specified paths relative to the root.
+// TODO(elezar): This function should be updated to make use of the oci.ContainerRoot type.
```
jgehrcke, line 143: :)

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
```diff
@@ -0,0 +1,132 @@
...
+	return nil
+}
+
+func (m command) run(c *cli.Context, cfg *options) error {
+	s, err := oci.LoadContainerState(cfg.containerSpec)
+	if err != nil {
+		return fmt.Errorf("failed to load container state: %v", err)
+	}
+
+	containerRoot, err := s.GetContainerRootDirPath()
```
jgehrcke, line 103: Below I see we sometimes have a nice `OnHost` suffix. Could this be applicable here, too?

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -103,53 +102,57 @@ func (m command) run(_ *cli.Context, cfg *options) error {
...
 	}
 
-	containerRootDir, err := s.GetContainerRoot()
+	containerRootDirPath, err := s.GetContainerRootDirPath()
 	if err != nil {
 		return fmt.Errorf("failed to determined container root: %w", err)
 	}
 
-	containerForwardCompatDir, err := m.getContainerForwardCompatDir(containerRoot(containerRootDir), cfg.hostDriverVersion)
+	containerForwardCompatDir, err := m.getContainerForwardCompatDirPathInContainer(containerRootDirPath, cfg.hostDriverVersion)
```
jgehrcke, line 110: Thanks for having added the `InContainer` suffix here. I know these things get lengthy but the information density is really good and this makes stuff no-brainer readable, which was the goal.

And again, the `containerRootDirPath` can be thought of as a `containerRootDirPathOnHost`, right?

jgehrcke (approved):
Thank you for all the iteration. Let's land this. I have left a few more questions while browsing this, but please don't slow down landing this because of my remarks.

cmd/nvidia-cdi-hook/chmod/chmod.go
elezar: Yes. It is the absolute path on the host where the root of the container's filesystem is located.

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
elezar: Do you mean in the for the fuction or the variable name? (or both).

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to determined container root: %v", err)
+	}
+	if containerRoot == "" {
+		m.logger.Warningf("No container root detected")
```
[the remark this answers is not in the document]
elezar: It may be that someone has launched the container runtime with a malformed container specification or that there is a mismatch between what we expect and (future) OCI runtime specification versions. This is more of a failsafe for an unlikely event than something we generally expect.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: Yes, the variable should be `containerRootDirPathOnHost`.

internal/discover/ldconfig.go
```diff
@@ -51,28 +51,22 @@ func (d ldconfig) Hooks() ([]Hook, error) {
...
-}
-
-// createLDCacheUpdateHook locates the NVIDIA Container Toolkit CLI and creates a hook for updating the LD Cache
-func createLDCacheUpdateHook(hookCreator HookCreator, ldconfig string, libraries []string) *Hook {
 	var args []string
 
-	if ldconfig != "" {
-		args = append(args, "--ldconfig-path", ldconfig)
+	if d.ldconfigPath != "" {
+		args = append(args, "--ldconfig-path", d.ldconfigPath)
```
jgehrcke, line 57: :muscle:

internal/ldconfig/ldconfig.go
```diff
@@ -0,0 +1,203 @@
...
+		"-f", "/etc/ld.so.conf",
+	}
+
+	if l.ldcacheExists() {
+		args = append(args, "-C", "/etc/ld.so.cache")
+	} else {
+		args = append(args, "-N")
+	}
+
+	// If we did not create the
```
jgehrcke: If we did not create the rest of the sentence, then

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
```diff
@@ -0,0 +1,166 @@
...
+const (
+	reexecUpdateLdCacheCommandName = "reexec-create-soname-symlinks"
+)
+
+type command struct {
+	logger logger.Interface
+}
+
+type options struct {
+	folders       cli.StringSlice
```
jgehrcke, line 43: `directoryPaths`?

internal/ldconfig/ldconfig.go
elezar: Updated.

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
ArangoGutierrez: agree with Jan-P Directory is a much nicer call to Folder.

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
```diff
@@ -0,0 +1,166 @@
...
+		},
+		Action: func(c *cli.Context) error {
+			return m.run(c, &cfg)
+		},
+	}
+
+	c.Flags = []cli.Flag{
+		&cli.StringSliceFlag{
+			Name:        "folder",
+			Usage:       "Specify a folder to generate soname symlinks in",
```
ArangoGutierrez: if we end up deciding for directory over folder on the discussion above, let's make sure we update this string

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
```diff
@@ -0,0 +1,166 @@
...
+			Destination: &cfg.folders,
```
ArangoGutierrez, line 83: and here

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
```diff
@@ -0,0 +1,166 @@
...
+
+// createSonameSymlinksHandler wraps createSonameSymlinks with error handling.
+func createSonameSymlinksHandler() {
+	if err := createSonameSymlinks(os.Args); err != nil {
+		log.Printf("Error updating ldcache: %v", err)
+		os.Exit(1)
+	}
+}
+
+// createSonameSymlinks runs ldconfig enusures that soname symlinks are created
```
ArangoGutierrez:
```suggestion
// createSonameSymlinks runs ldconfig ensures that soname symlinks are created
```

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
```diff
@@ -0,0 +1,166 @@
...
+// in the specified directories.
+// It is invoked from a reexec'd handler and provides namespace isolation for
+// the operations performed by this hook. At the point where this is invoked,
+// we are in a new mount namespace that is cloned from the parent.
+//
+// args[0] is the reexec initializer function name
+// args[1] is the path of the ldconfig binary on the host
+// args[2] is the container root directory
```
ArangoGutierrez, line 148: we use directory here, I guess this is a good clue on how to define the discussion above.

cmd/nvidia-cdi-hook/update-ldcache/update-ldcache.go
```diff
@@ -143,113 +135,30 @@ func updateLdCacheHandler() {
...
+// libraries that are mounted from the host.
+// It is invoked from a reexec'd handler and provides namespace isolation for
+// the operations performed by this hook. At the point where this is invoked,
+// we are in a new mount namespace that is cloned from the parent.
 //
 // args[0] is the reexec initializer function name
 // args[1] is the path of the ldconfig binary on the host
 // args[2] is the container root directory
-// The remaining args are folders that need to be added to the ldcache.
+// The remaining args are folders where soname symlinks need to be created.
```
ArangoGutierrez, line 147: jeje folders vs dir conversation. I guess is was a nice Nit catch by @jgehrcke

internal/ldconfig/ldconfig.go
```diff
@@ -0,0 +1,206 @@
...
+		return nil, fmt.Errorf("an ldconfig path must be specified")
+	}
+	if inRoot == "" || inRoot == "/" {
+		return nil, fmt.Errorf("ldconfig must be run in the non-system root")
+	}
+	return l, nil
+}
+
+// CreateSonameSymlinks uses ldconfig to create the soname symlinks in the
+// specified folders.
```
ArangoGutierrez: folder vs dir

internal/ldconfig/ldconfig.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if l.ldcacheExists() {
+		args = append(args, "-C", "/etc/ld.so.cache")
+	} else {
+		args = append(args, "-N")
+	}
+
+	// If the ld.so.conf.d directory exists, we create a config file there
+	// containing the required directories, otherwise we add the specified
+	// directories to the ldconfig command directly.
```
ArangoGutierrez, line 116: here we use dir but some lines above we use folder

ArangoGutierrez (changes requested):
Some quick nits, I feel I lean into more towards `directory` than `folder`. 
PR Looks good overall just naming convetion discussion and one misSpell

elezar:
> Some quick nits, I feel I lean into more towards `directory` than `folder`.

Please note that the API for the hook uses `--folder` as a command line argument and this is not something that I want to change at the moment. I will update the local content to be more consistent though.

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
elezar: These options are filled by the CLI. A convention that I use is that the member matches (with the exception of plurality if required) the primary flag. Since the CLI argument is `--folder` (and can be repeated), I want to keep this `folders`.

cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
elezar: This directory is different from the other directories that we're talking about.

PR #965, opened 2025-03-06
Bump version for v1.17.5 release

elezar:
These changes bump the version and update the changelog for the v1.17.5 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # NVIDIA Container Toolkit Changelog
 
+## v1.17.5
+
+- Fix bug that may overwrite docker feature flags when configuring CDI from the `nvidia-ctk runtime configure` command.
+- Remove the unused `Set()` function from engine config API.
+- Add an `EnableCDI()` method to engine config API.
+- Add an `ignore-imex-channel-requests` feature flag. This ensures that the NVIDIA Container Runtime can be configured to ignore IMEX channel requests when these should be managed by another component.
+- Update the `update-ldcache` hook to run the host `ldconfig` from a MEMFD.
+- Add support for CUDA Forward Compatibility (removed by default in v1.17.4) using a dedicated `enable-cuda-compat` hook.  This can be disabled using a `disable-cuda-compat-lib-hook` feature flag.
+- Disable nvsandboxutils in the `nvcdi` API. This prevents a segmentation violation with NVIDIA GPU Drivers from the 565 branch.
+- Fix a bug where `cdi` mode would not work with the `--gpus` flag even if the NVIDIA Container Runtime was used.
+
+### Changes in the Toolkit Container
+
+- Enable CDI in container engine (Containerd, Cri-o, Docker) if CDI_ENABLED is set.
+- Bump CUDA base image version to 12.8.0
+
 ## v1.17.4
 - Disable mounting of compat libs from container by default
 - Add allow-cuda-compat-libs-from-container feature flag
--- a/versions.mk
+++ b/versions.mk
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 LIB_NAME := nvidia-container-toolkit
-LIB_VERSION := 1.17.4
+LIB_VERSION := 1.17.5
 LIB_TAG :=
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # NVIDIA Container Toolkit Changelog
 
+## v1.17.5
+
+- Fix bug that may overwrite docker feature flags when configuring CDI from the `nvidia-ctk runtime configure` command.
+- Remove the unused `Set()` function from engine config API.
+- Add an `EnableCDI()` method to engine config API.
+- Add an `ignore-imex-channel-requests` feature flag. This ensures that the NVIDIA Container Runtime can be configured to ignore IMEX channel requests when these should be managed by another component.
```
jgehrcke, line 8: :muscle:

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+- Update the `update-ldcache` hook to run the host `ldconfig` from a MEMFD.
+- Add support for CUDA Forward Compatibility (removed by default in v1.17.4) using a dedicated `enable-cuda-compat` hook.  This can be disabled using a `disable-cuda-compat-lib-hook` feature flag.
+- Disable nvsandboxutils in the `nvcdi` API. This prevents a segmentation violation with NVIDIA GPU Drivers from the 565 branch.
+- Fix a bug where `cdi` mode would not work with the `--gpus` flag even if the NVIDIA Container Runtime was used.
```
jgehrcke, line 12: Sometimes we do "CDI", and here we do `cdi`, but that may be on purpose.

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # NVIDIA Container Toolkit Changelog
 
+## v1.17.5
```
jgehrcke, line 3: Thanks for adding release notes / changelog entries.

Thought: in the future it would be nice if especially "fix a bug.." would link to a ticket describing the bug.

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+- Add an `EnableCDI()` method to engine config API.
+- Add an `ignore-imex-channel-requests` feature flag. This ensures that the NVIDIA Container Runtime can be configured to ignore IMEX channel requests when these should be managed by another component.
+- Update the `update-ldcache` hook to run the host `ldconfig` from a MEMFD.
+- Add support for CUDA Forward Compatibility (removed by default in v1.17.4) using a dedicated `enable-cuda-compat` hook.  This can be disabled using a `disable-cuda-compat-lib-hook` feature flag.
+- Disable nvsandboxutils in the `nvcdi` API. This prevents a segmentation violation with NVIDIA GPU Drivers from the 565 branch.
+- Fix a bug where `cdi` mode would not work with the `--gpus` flag even if the NVIDIA Container Runtime was used.
+
+### Changes in the Toolkit Container
+
+- Enable CDI in container engine (Containerd, Cri-o, Docker) if CDI_ENABLED is set.
```
jgehrcke, line 16: Does this introduce a new option CDI_ENABLED? Is CDI_ENABLED an environment variable?

The way it's worded it sounds like previously when CDI_ENABLED was set then CDI was not actually enabled (which sounds like a bug).

jgehrcke (approved):
Cool. Thanks. I left a few comments for inspiration, but this LGTM. Approving.

ArangoGutierrez (approved):

CHANGELOG.md
elezar: `cdi` is an actual mode. CDI is the general mechanism.

CHANGELOG.md
elezar: Yes, we could get better at tracking these things. Sometimes these are internal only and I try to avoid referencing those here.

CHANGELOG.md
elezar: When `CDI_ENABLED` was set, we would do things related to CDI for the NVIDIA Container Runtime but not the container engine. The change was to *also* enable CDI in the container engine. This is required for DRA.

PR #968, opened 2025-03-07
Allow enable-cuda-compat hook to be disabled in CDI spec generation

elezar:
This change adds support to the nvcdi package to opt out of specific hooks.

Currently only the `enable-cuda-compat` hook is supported. This allows clients to generate a CDI spec that is compatible with older nvidia-cdi-hook CLIs.

This is required in cases where a tool such as the k8s-device-plugin or k8s-dra-driver-gpu needs to generate a CDI specification for a version of the NVIDIA Container Toolkit that does not support this hook.

### DIFF
--- a/pkg/nvcdi/api.go
+++ b/pkg/nvcdi/api.go
@@ -35,3 +35,13 @@ type Interface interface {
 	GetMIGDeviceSpecs(int, device.Device, int, device.MigDevice) ([]specs.Device, error)
 	GetDeviceSpecsByID(...string) ([]specs.Device, error)
 }
+
+// A HookName refers to one of the predefined set of CDI hooks that may be
+// included in the generated CDI specification.
+type HookName string
+
+const (
+	// HookEnableCudaCompat refers to the hook used to enable CUDA Forward Compatibility.
+	// This was added with v1.17.5 of the NVIDIA Container Toolkit.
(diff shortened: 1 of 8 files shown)


pkg/nvcdi/api.go
```diff
@@ -35,3 +35,13 @@ type Interface interface {
...
 }
+
+// A HookName refers to one of the predefined set of CDI hooks that may be
+// included in the generated CDI specification.
+type HookName string
+
+const (
+	// HookEnableCudaCompat refers to the hook used to enable CUDA Forward Compatibility.
+	// This was added with v1.17.5 of the NVIDIA Container Toolkit.
+	HookEnableCudaCompat = HookName("enable-cuda-compat")
```
ArangoGutierrez, line 46: Not a blocker, but we should have as TODO to have a central place to define this internal names to make it easier to  edit/change a hook name in the future.

ArangoGutierrez (approved):

pkg/nvcdi/api.go
elezar: Yes, we should refactor the hook handling to have them defined more consistently. Some of this falls in the scope of pulling the CDI spec generation into `go-nvlib` instead of implementing it here in the NVIDIA Container Toolkit. This would also mean that the device plugin and the dra driver do not need to depend on the `nvidia-container-toolkit`.

PR #970, opened 2025-03-07
Allow enable-cuda-compat hook to be disabled in CDI spec generation

elezar:
This change adds support to the nvcdi package to opt out of specific hooks.

Currently only the `enable-cuda-compat` hook is supported. This allows clients to generate a CDI spec that is compatible with older nvidia-cdi-hook CLIs.

This is required in cases where a tool such as the k8s-device-plugin or k8s-dra-driver-gpu needs to generate a CDI specification for a version of the NVIDIA Container Toolkit that does not support this hook.

Backports #968


ArangoGutierrez (approved):

jgehrcke (approved):
:rocket: Thank you!

PR #976, opened 2025-03-09
Create device nodes

elezar:


internal/lookup/root/root.go
```diff
@@ -125,3 +128,20 @@ func xdgDataDirs() []string {
 
 	return []string{"/usr/local/share", "/usr/share"}
 }
+
+// LoadKmods loads the specified kernel modules in the driver root.
```
ArangoGutierrez, line 132:
```suggestion
// LoadKernelModules loads the specified kernel modules in the driver root.
```

internal/nvcaps/nvcaps.go
```diff
@@ -36,27 +36,62 @@ const (
...
 )
 
-// MigMinor represents the minor number of a MIG device
-type MigMinor int
+// An Index represents a gpu, ci, or gi index.
+// We use uint32 as this typically maps to a device minor number.
+type Index uint32
 
-// MigCap represents the path to a MIG cap file
+// MigMinor represents the minor number of a MIG device
```
ArangoGutierrez, line 43: Should we have consistency on the `// doc` I see for Index we have `// A VarName` but for the rest we use the regular `// VarName`

ArangoGutierrez (commented):
just 2 minor comments, I'll run a manual test now

ArangoGutierrez:
> Note
From https://download.nvidia.com/XFree86/Linux-x86_64/396.54/README/README.txt

```
Q. How and when are the NVIDIA device files created?

A. When a user-space NVIDIA driver component needs to communicate with the
   NVIDIA kernel module, and the NVIDIA character device files do not yet
   exist, the user-space component will first attempt to load the kernel
   module and create the device files itself.

   Device file creation and kernel module loading generally require root
   privileges. The X driver, running within a setuid root X server, will have
   these privileges, but not, e.g., the CUDA driver within the environment of
   a normal user.

   If the user-space NVIDIA driver component cannot load the kernel module or
   create the device files itself, it will attempt to invoke the setuid root
   nvidia-modprobe utility, which will perform these operations on behalf of
   the non-privileged driver.
```

We have found that with the current `nvidia-ctk` , calling `nvidia-ctk cdi generate` causes the device nodes to be created.

elezar:
Closing this one. Will reopen if required.

PR #990, opened 2025-03-13
Refactor toolkit installer

elezar:
These changes refactor the toolkit installer used from the toolkit contianer. 

The goal is to prepare for ##602 where a single container image is used for the toolkit instead of multiple container images.


cmd/nvidia-ctk-installer/toolkit/installer/artifact-root.go
```diff
@@ -0,0 +1,84 @@
+/**
```
ArangoGutierrez, line 1: is there a reason this license header is different?

ArangoGutierrez (commented):
First pass looks good. will give it a try

cmd/nvidia-ctk-installer/toolkit/installer/artifact-root.go
elezar: Different from what? The reason is that these changes have been on a branch associated with #602 for some time and I am factoring them out. Will update the header.

PR #1000, opened 2025-03-19
Issue warning on unsupported CDI hook

elezar:
To allow for CDI hooks to be added gradually we provide a generic no-op hook for unrecognised subcommands. This will log a warning (to STDERR) instead of erroring out.

For example:
```
$ ./nvidia-cdi-hook foo
WARN[0000] Unsupported CDI hook: foo
$ echo $?
0
$ ./nvidia-cdi-hook
WARN[0000] No CDI hook specified
$ echo $?
0
```

```
./nvidia-ctk hook foo
WARN[0000] Unsupported CDI hook: foo
$ echo $?
0
./nvidia-ctk hook
WARN[0000] No CDI hook specified
$ echo $?
0
```

### DIFF
--- a/cmd/nvidia-cdi-hook/main.go
+++ b/cmd/nvidia-cdi-hook/main.go
@@ -51,6 +51,18 @@ func main() {
 	c.Usage = "Command to structure files for usage inside a container, called as hooks from a container runtime, defined in a CDI yaml file"
 	c.Version = info.GetVersionString()
 
+	// We set the default action for the `nvidia-cdi-hook` command to issue a
+	// warning and exit with no error.
+	// This means that if an unsupported hook is run, a container will not fail
+	// to launch. An unsupported hook could be the result of a CDI specification
+	// referring to a new hook that is not yet supported by an older NVIDIA
+	// Container Toolkit version or a hook that has been removed in newer
+	// version.
+	c.Action = func(ctx *cli.Context) error {
+		commands.IssueUnsupportedHookWarning(logger, ctx)
+		return nil
+	}
+
 	// Setup the flags for this command
 	c.Flags = []cli.Flag{
 		&cli.BoolFlag{
--- a/cmd/nvidia-ctk/hook/hook.go
+++ b/cmd/nvidia-ctk/hook/hook.go
@@ -27,7 +27,7 @@ type hookCommand struct {
 	logger logger.Interface
 }
 
-// NewCommand constructs a hook command with the specified logger
+// NewCommand constructs CLI subcommand for handling CDI hooks.
 func NewCommand(logger logger.Interface) *cli.Command {
 	c := hookCommand{
 		logger: logger,
@@ -37,10 +37,21 @@ func NewCommand(logger logger.Interface) *cli.Command {
 
 // build
 func (m hookCommand) build() *cli.Command {
-	// Create the 'hook' command
+	// Create the 'hook' subcommand
 	hook := cli.Command{
 		Name:  "hook",
 		Usage: "A collection of hooks that may be injected into an OCI spec",
+		// We set the default action for the `hook` subcommand to issue a
+		// warning and exit with no error.
+		// This means that if an unsupported hook is run, a container will not fail
+		// to launch. An unsupported hook could be the result of a CDI specification
+		// referring to a new hook that is not yet supported by an older NVIDIA
+		// Container Toolkit version or a hook that has been removed in newer
+		// version.
+		Action: func(ctx *cli.Context) error {
+			commands.IssueUnsupportedHookWarning(m.logger, ctx)
+			return nil
+		},
 	}
 
 	hook.Subcommands = commands.New(m.logger)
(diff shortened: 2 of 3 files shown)


ArangoGutierrez (approved):

cmd/nvidia-ctk/hook/hook.go
```diff
@@ -41,6 +41,9 @@ func (m hookCommand) build() *cli.Command {
 	hook := cli.Command{
 		Name:  "hook",
```
jgehrcke, line 42: From the CLI libraries perspective -- does this define a subcommand named "hook"?

jgehrcke: if so, the code comment maybe should be "create 'hook' subcommand" (just noting this as I try to understand the code here)

cmd/nvidia-ctk/hook/hook.go
```diff
@@ -41,6 +41,9 @@ func (m hookCommand) build() *cli.Command {
...
 		Usage: "A collection of hooks that may be injected into an OCI spec",
+		Action: func(ctx *cli.Context) error {
+			return commands.WarnOnUnsupportedSubcommand(m.logger, ctx)
+		},
 	}
 
 	hook.Subcommands = commands.New(m.logger)
```
jgehrcke, line 49: oh. Hm maybe "hook" is the command, and sub-commands are one level deeper.

cmd/nvidia-cdi-hook/main.go
```diff
@@ -51,6 +51,11 @@ func main() {
 	c.Usage = "Command to structure files for usage inside a container, called as hooks from a container runtime, defined in a CDI yaml file"
 	c.Version = info.GetVersionString()
 
+	// We add an action to the hook
```
jgehrcke, line 54: The comment "We add an action to the hook" does not help me understand the code better :).

I think maybe instead of "We add an action to the hook" we can write "Add a default action to this CLI. It is executed when the user-provided sub-command is unknown"

I still wonder about the hierarchy provided by this CLI library. There is a CLI. The CLI has various commands. Each command has subcommands. Is that true? Or is the CLI the command? 

I ask because I wonder: what is a hook? A command? A subcommand?

Again, this is about code readability. I have built CLIs with various different libraries and ecosystems and I know everyone calls things differently... :)

cmd/nvidia-cdi-hook/main.go
```diff
@@ -51,6 +51,11 @@ func main() {
...
+	c.Action = func(ctx *cli.Context) error {
+		return commands.WarnOnUnsupportedSubcommand(logger, ctx)
```
jgehrcke, line 56: I suppose a decisive design decision here is "emit warning on stderr, and exit 0".

Is that right? I think we should maybe document that via code comment.

jgehrcke (approved):
Thank you @elezar! Merge this at will. We talked about offline, and of course it's ready to land.

In terms of code readability I have however left a few questions / remarks / thoughts. Please reply and/or adjust code at your own discretion! :)

cmd/nvidia-ctk/hook/hook.go
elezar: The CLI library that we use (urfave) has the following heirarchy:
A top-level CLI `App` can have one or more `Command`s. Each `Command` can have one or more `Subcommands`. 

Note that The `Subcommands` are themselves `Commands` which actually allows for arbitrary nesting.

I agree that from a documentation perspective, what we mean when we say "command" is "subcommand" and should use "command" for "app" (the top-level CLI) instead.

cmd/nvidia-cdi-hook/main.go
elezar: Yes, a hook as a general concept is an executable that is run by a low-level runtime at various points of a container's lifecycle. These are passed the container state via STDIN which we use to extract the information about the container -- for example the container root or even the environment variables in the container in the case of the `nvidia-container-runtime-hook`.

For CDI we have a number of smaller hooks that are supposed to do "one" thing (or at least have their intent clear from their name). We originally implemented these as subcommands of the `hook` subcommand of the `nvidia-ctk` command so as to simplify the distribution of the executable. In order to simplify discussions around hooks and what their intent are, we later decided to implement a separate command `nvidia-cdi-hook` that implements the same hooks as subcommands. With this done the following are equivalent:

```
./nvidia-cdi-hook [HOOK_NAME]
./nvidia-ctk hook [HOOK_NAME]
```

As I also mentioned on the call, we could take this further and implement *separate* executables for each of the hooks. This does have the disadvantage that we would not be able to implement "default" behaviour as we are doing here.

elezar:
> ```
> WARN[0000] Unsupported hook &[foo]
> ```
> 
> 1. does this get emitted to stderr?
> 2. why do we see `&[foo]` and not `"foo"`?

The standard logger that's used (`logrus.New()`) will print this to STDERR. The output is a result of how go is converting the arg(s) to a string for output.

cmd/nvidia-ctk/hook/hook.go
elezar: I hope my discussion above has cleared up the confusion here. We should make a pass of the existing code and update the comments / functions for consistency, but this has a low priority at present.

PR #1016, opened 2025-04-02
Allow container runtime executable path to be specified

### LINKED ISSUE #803: container-toolkit on k0s leads to unsupported config version: 3
Hello,

I'm trying to install nvidia helm in a k0s cluster:

```
[root@miriam ~]# k0s version
v1.31.2+k0s.0
[root@miriam ~]# /var/lib/k0s/bin/containerd -v
containerd github.com/containerd/containerd 1.7.22 7f7fdf5fed64eb6a7caf99b3e12efcf9d60e311c
[root@miriam ~]#
```

As stated possible at https://docs.k0sproject.io/v1.31.2+k0s.0/runtime/#using-nvidia-container-runtime, I've set the helm chart to following options at nvidia-container-toolkit:

```
toolkit:
  enabled: true
  repository: nvcr.io/nvidia/k8s
  image: container-toolkit
  version: v1.17.2-ubuntu20.04
  imagePullPolicy: IfNotPresent
  imagePullSecrets: []
  env:
    - name: CONTAINERD_CONFIG
      value: "/etc/k0s/containerd.d/nvidia.toml"
    - name: CONTAINERD_SOCKET
      value: "/run/k0s/containerd.sock"
    - name: CONTAINERD_RUNTIME_CLASS
      value: "nvidia"
    - name: CONTAINERD_SET_AS_DEFAULT
      value: "false"
    - name: CONTAINERD_USE_LEGACY_CONFIG
      value: "true"
  resources: {}
  installDir: "/usr/local/nvidia"
```

The usage of `CONTAINERD_USE_LEGACY_CONFIG` was an attempt after reading issue https://github.com/NVIDIA/nvidia-container-toolkit/pull/777 after the recommended way from k0s did not worked.

Anyways I run, what I get is:

```
IS_HOST_DRIVER=true
NVIDIA_DRIVER_ROOT=/
DRIVER_ROOT_CTR_PATH=/host
NVIDIA_DEV_ROOT=/
DEV_ROOT_CTR_PATH=/host
time="2024-11-16T02:28:31Z" level=info msg="Parsing arguments"
time="2024-11-16T02:28:31Z" level=info msg="Starting nvidia-toolkit"
time="2024-11-16T02:28:31Z" level=info msg="disabling device node creation since --cdi-enabled=false"
time="2024-11-16T02:28:31Z" level=info msg="Verifying Flags"
time="2024-11-16T02:28:31Z" level=info msg=Initializing
time="2024-11-16T02:28:31Z" level=info msg="Shutting Down"
time="2024-11-16T02:28:31Z" level=error msg="error running nvidia-toolkit: unable to determine runtime options: unable to load containerd config: unsupported config version: 3"
```

After checking some [source code](https://github.com/NVIDIA/nvidia-container-toolkit/blob/release-1.17/pkg/config/engine/containerd/containerd.go#L94), I guess legacyConfig does not get proper version to create the file,  but I haven't deeply read the source to understand (Also, not a good Go coder).

That said, what may be wrong to nvidia-toolkit does not create the legacyConfig in the specified folder properly?

Thanks!

elezar:
This change adds support for specifying the container runtime executable path. This can be used if, for example, there are two containerd executables and a specific one must be used.

Deploying to a k0s system with:
```
helm install gpu-operator -n gpu-operator --create-namespace \
  nvidia/gpu-operator $HELM_OPTIONS \
    --version=v25.3.0 \
    --set toolkit.repository=ghcr.io/nvidia \
    --set toolkit.version=ae385428-ubuntu20.04 \
    --set toolkit.env[0].name=RUNTIME_CONFIG \
    --set toolkit.env[0].value=/etc/k0s/containerd.d/nvidia.toml \
    --set toolkit.env[1].name=RUNTIME_SOCKET \
    --set toolkit.env[1].value=/run/k0s/containerd.sock \
    --set toolkit.env[2].name=RUNTIME_EXECUTABLE_PATH \
    --set toolkit.env[2].value=/var/lib/k0s/bin/containerd \
    --set toolkit.env[3].name=NVIDIA_RUNTIME_NAME \
    --set toolkit.env[3].value=nvidia
```

Allows the config to be extracted correctly and unblocks the deployment.

Fixes #803

### DIFF
--- a/cmd/nvidia-ctk-installer/container/runtime/runtime.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/runtime.go
@@ -26,6 +26,7 @@ import (
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/container/runtime/crio"
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/container/runtime/docker"
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/container/toolkit"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
 )
 
 const (
@@ -52,6 +53,12 @@ func Flags(opts *Options) []cli.Flag {
 			Value:       runtimeSpecificDefault,
 			Destination: &opts.Config,
 			EnvVars:     []string{"RUNTIME_CONFIG", "CONTAINERD_CONFIG", "DOCKER_CONFIG"},
+		},
+		&cli.StringFlag{
+			Name:        "executable-path",
+			Usage:       "The path to the runtime executable. This is used to extract the current config",
+			Destination: &opts.ExecutablePath,
+			EnvVars:     []string{"RUNTIME_EXECUTABLE_PATH"},
 		},
 		&cli.StringFlag{
 			Name:        "socket",
@@ -104,13 +111,18 @@ func Flags(opts *Options) []cli.Flag {
 	return flags
 }
 
-// ValidateOptions checks whether the specified options are valid
-func ValidateOptions(c *cli.Context, opts *Options, runtime string, toolkitRoot string, to *toolkit.Options) error {
+// Validate checks whether the specified options are valid
+func (opts *Options) Validate(logger logger.Interface, c *cli.Context, runtime string, toolkitRoot string, to *toolkit.Options) error {
 	// We set this option here to ensure that it is available in future calls.
 	opts.RuntimeDir = toolkitRoot
 
 	if !c.IsSet("enable-cdi-in-runtime") {
 		opts.EnableCDI = to.CDI.Enabled
+	}
+
+	if opts.ExecutablePath != "" && opts.RuntimeName == docker.Name {
+		logger.Warningf("Ignoring executable-path=%q flag for %v", opts.ExecutablePath, opts.RuntimeName)
+		opts.ExecutablePath = ""
 	}
 
 	// Apply the runtime-specific config changes.
--- a/cmd/nvidia-ctk/runtime/configure/configure.go
+++ b/cmd/nvidia-ctk/runtime/configure/configure.go
@@ -68,6 +68,7 @@ type config struct {
 	dryRun         bool
 	runtime        string
 	configFilePath string
+	executablePath string
 	configSource   string
 	mode           string
 	hookFilePath   string
@@ -117,6 +118,11 @@ func (m command) build() *cli.Command {
 			Name:        "config",
 			Usage:       "path to the config file for the target runtime",
 			Destination: &config.configFilePath,
+		},
+		&cli.StringFlag{
+			Name:        "executable-path",
+			Usage:       "The path to the runtime executable. This is used to extract the current config",
+			Destination: &config.executablePath,
 		},
 		&cli.StringFlag{
 			Name:        "config-mode",
@@ -204,6 +210,11 @@ func (m command) validateFlags(c *cli.Context, config *config) error {
 			m.logger.Warningf("Ignoring cdi.enabled flag for %v", config.runtime)
 		}
 		config.cdi.enabled = false
+	}
+
+	if config.executablePath != "" && config.runtime == "docker" {
+		m.logger.Warningf("Ignoring executable-path=%q flag for %v", config.executablePath, config.runtime)
+		config.executablePath = ""
 	}
 
 	switch config.configSource {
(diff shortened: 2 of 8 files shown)


cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -53,6 +53,11 @@ func Flags(opts *Options) []cli.Flag {
 			Destination: &opts.Config,
 			EnvVars:     []string{"RUNTIME_CONFIG", "CONTAINERD_CONFIG", "DOCKER_CONFIG"},
 		},
+		&cli.StringFlag{
+			Name:        "container-runtime-executable-path",
```
tariq1890, line 57: Why not just `executable-path` or `runtime-executable-path`? This is consistent with how `socket` is named

pkg/config/engine/containerd/containerd.go
```diff
@@ -162,8 +162,11 @@ func (c *Config) GetRuntimeConfig(name string) (engine.RuntimeConfig, error) {
 }
 
 // CommandLineSource returns the CLI-based containerd config loader
-func CommandLineSource(hostRoot string) toml.Loader {
-	return toml.FromCommandLine(chrootIfRequired(hostRoot, "containerd", "config", "dump")...)
+func CommandLineSource(hostRoot string, executablePath string) toml.Loader {
+	if executablePath == "" {
+		executablePath = "containerd"
```
tariq1890, line 167: Probably tangential : So far, we've made the assumption that the `containerd` binary will be resolved via `$PATH`. Would it make sense to look up a set of known absolute paths of the containerd binary until we git a hit? 

I see that you already have a method to do exactly just this: https://github.com/NVIDIA/nvidia-container-toolkit/blob/1d2e1bd403102262beea909ffc99d26f294f084f/internal/lookup/path.go#L36

elezar: That could make sense. Do you have a list of known paths that you would like to check?

For reference the "executable" locator that uses `GetPaths` also checks the `PATH` variable to determine which locations to check (in addition to the default values). I didn't want to change that behavior as part of this change though.

cmd/nvidia-ctk-installer/container/runtime/runtime.go
elezar: Yes, I think this may be better. Let me take another pass.

elezar:
@tariq1890 do we need to backport this?

pkg/config/engine/containerd/containerd.go
tariq1890: Agreed

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -104,15 +111,20 @@ func Flags(opts *Options) []cli.Flag {
...
+// Validate checks whether the specified options are valid
+func (opts *Options) Validate(logger logger.Interface, c *cli.Context, runtime string, toolkitRoot string, to *toolkit.Options) error {
 	// We set this option here to ensure that it is available in future calls.
 	opts.RuntimeDir = toolkitRoot
 
 	if !c.IsSet("enable-cdi-in-runtime") {
 		opts.EnableCDI = to.CDI.Enabled
 	}
 
+	if opts.ExecutablePath != "" && opts.RuntimeName != containerd.Name {
```
tariq1890, line 123: I think we can add support for custom crio exec paths too

elezar: I thought so too at first. The issue here is that for crio we have:
```
// CommandLineSource returns the CLI-based crio config loader
func CommandLineSource(hostRoot string) toml.Loader {
	return toml.LoadFirst(
		toml.FromCommandLine(chrootIfRequired(hostRoot, "crio", "status", "config")...),
		toml.FromCommandLine(chrootIfRequired(hostRoot, "crio-status", "config")...),
	)
}
```
Do you mean that we ONLY replace `crio` above and leave `crio-status` as is?

cmd/nvidia-ctk/runtime/configure/configure.go
```diff
@@ -208,9 +212,9 @@ func (m command) validateFlags(c *cli.Context, config *config) error {
 		config.cdi.enabled = false
 	}
 
-	if config.runtimeConfigOverrideJSON != "" && config.runtime != "containerd" {
-		m.logger.Warningf("Ignoring runtime-config-override flag for %v", config.runtime)
-		config.runtimeConfigOverrideJSON = ""
+	if config.executablePath != "" && config.runtime != "containerd" && config.runtime != "crio" {
```
tariq1890, line 215: containerd, crio and docker are the only recognised runtimes by the code base. So wouldn't it just suffice to check if the runtime is docker instead?

elezar: Docker is also supported by the `nvidia-ctk runtime configure` command.

elezar: Updated.

PR #1030, opened 2025-04-09
Add envvar for libcuda.so parent dir to CDI spec

elezar:
This change adds a LIBCUDA_SO_PARENT_DIRECTORY_CONTAINER_PATH envvar to a generated CDI specification. This reports where the `libcuda.so.*` libraries will be injected into the container.


internal/discover/discover.go
```diff
@@ -22,6 +22,12 @@ type Device struct {
 	Path     string
 }
 
+// EnvVar represents a discovered environment variable.
+type EnvVar struct {
+	Name  string
```
ArangoGutierrez, line 27: I am more of a Key/Value mind, but not a blocker.

elezar: From https://wiki.debian.org/EnvironmentVariables#:~:text=Variables%20are%20defined,the%20terminal%20session.

> Variables are defined with name-value pairs: NAME=any string as value. The variable name is usually in capital letters. Anything that follows the equal-sign is considered the variable's value until the terminating line feed character. Variables can be defined ad hoc in a terminal by writing the appropriate command. In Bash this would be export MYVAL="Hello world". In this case the variable stays defined until the end of the terminal session.

jgehrcke: Ha. In math and computer science we indeed typically say "variable name", I think Eduardo would agree.

But I like keys, too. Especially the old ones :old_key:

ArangoGutierrez: Name is ok, marking this one as resolved

PR #1046, opened 2025-04-23
Fix resolution of libs in LDCache on ARM

### LINKED ISSUE #1045: Generate CDI specifications for architectures other than x86_64
The `nvidia-ctk` helper program fails to parse the `ldcache` file for `aarch64` because the list of flags in the code that parses this file is incomplete, compared to its [counterpart] in `libnvidia-container`

[counterpart]: https://github.com/NVIDIA/libnvidia-container/blob/a198166e1c1166f4847598438115ea97dacc7a92/src/ldcache.h#L21

I was able to fix this by updating [this part of the code] to use `flagArchAarch64` with value `0x0a00`, as `libnvidia-container` does.

[this part of the code]: https://github.com/NVIDIA/nvidia-container-toolkit/blob/de3d7366637c8b0646df66b2134884d3ceafaa9e/internal/ldcache/ldcache.go#L188

elezar:
Since we explicitly check for the architecture of the libraries in the ldcache, we need to also check the architecture flag against the ARM constants.

Fixes #1045

### DIFF
--- a/internal/ldcache/ldcache.go
+++ b/internal/ldcache/ldcache.go
@@ -47,6 +47,11 @@ const (
 	flagArchX8664   = 0x0300
 	flagArchX32     = 0x0800
 	flagArchPpc64le = 0x0500
+
+	// flagArch_ARM_LIBHF is the flag value for 32-bit ARM libs using hard-float.
+	flagArch_ARM_LIBHF = 0x0900
+	// flagArch_AARCH64_LIB64 is the flag value for 64-bit ARM libs.
+	flagArch_AARCH64_LIB64 = 0x0a00
 )
 
 var errInvalidCache = errors.New("invalid ld.so.cache file")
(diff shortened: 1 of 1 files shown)


internal/ldcache/ldcache.go
```diff
@@ -47,6 +47,11 @@ const (
 	flagArchX8664   = 0x0300
 	flagArchX32     = 0x0800
 	flagArchPpc64le = 0x0500
+
+	// flagArch_ARM_LIBHF is the flag value for 32-bit ARM libs using hard-float.
+	flagArch_ARM_LIBHF = 0x0900
+	// flagArch_AARCH64_LIB64 is the flag value for 64-bit ARM libs.
+	flagArch_AARCH64_LIB64 = 0x0a00
```
arnaldo2792, line 54: Yup, this is how I fixed it, should you all carry on all the flags that `libnvidia-container` supports?

elezar: Yes, we could do that. Although we don't actively test on some of the platforms called out there anymore.

I will audit the list once more.

elezar: @arnaldo2792 I think for the time being I'm not going to update this list and only stick to ARM. If you have experience with other platforms that we should address, please let me know.

jgehrcke (approved):
I don't understand much but it seems like the change has been made with care.

ArangoGutierrez (approved):

cdesiniotis (approved):

PR #1068, opened 2025-05-09
Fix resolution of libs in LDCache on ARM

elezar:
Since we explicitly check for the architecture of the libraries in the ldcache, we need to also check the architecture flag against the ARM constants.

Backport of #1046


ArangoGutierrez (approved):
Backport approved

PR #1069, opened 2025-05-09
[no-relnote] Skip nil discoverers in merge

elezar:
When constructing a list of discoverers using discover.Merge we explicitly skip `nil` discoverers to simplify usage as we don't have to explicitly check validity when processing the discoverers in the list.

Backport of #1065


ArangoGutierrez (approved):
Backport approved

Issue #1074, opened 2025-05-12
Add a flag to nvidia-ctk cdi generate to allow users to specificy which Hooks to be skipped


ArangoGutierrez:
When running the `nvidia-ctk cdi generate` command, a user should be able to opt out of specific hooks. 
We propose to add a flag `--skip-hooks` that will take a comma-separated list of hooks that will be skipped when creating the CDI spec.

PR #1077, opened 2025-05-12
Add --skip-hooks flag to cdi generate command

ArangoGutierrez:
When running the nvidia-ctk cdi generate command, a user should be able to opt out of specific hooks. We propose to add a flag --skip-hooks that will take a comma-separated list of hooks that will be skipped when creating the CDI spec.

Fixes: #1074


cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -176,6 +177,12 @@ func (m command) build() *cli.Command {
 			Usage:       "Specify a pattern the CSV mount specifications.",
 			Destination: &opts.csv.ignorePatterns,
 		},
+		&cli.StringSliceFlag{
+			Name:        "skip-hook",
+			Usage:       "Comma-separated list of hooks to skip when generating the CDI specification.",
```
elezar, line 182: Technically the flag can also be specified multiple times.

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -57,6 +57,7 @@ type options struct {
 
 	configSearchPaths  cli.StringSlice
 	librarySearchPaths cli.StringSlice
+	skipHook           cli.StringSlice
```
elezar, line 60: This should be:
```suggestion
	skipHooks           cli.StringSlice
```
Since it contains 0 or more entries.

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -276,7 +283,15 @@ func (m command) generateSpec(opts *options) (spec.Interface, error) {
...
 		// We set the following to allow for dependency injection:
 		nvcdi.WithNvmlLib(opts.nvmllib),
-	)
+	}
+
+	if len(opts.skipHook.Value()) > 0 {
+		for _, hook := range opts.skipHook.Value() {
+			initOpts = append(initOpts, nvcdi.WithDisabledHook(nvcdi.HookName(hook)))
+		}
+	}
```
elezar, line 292: Did we want to check for validity of the names, or do we want to add that as a follow-up?

ArangoGutierrez: let's hold this PR until:
- https://github.com/NVIDIA/nvidia-container-toolkit/pull/947
- https://github.com/NVIDIA/nvidia-container-toolkit/pull/927
get in, then I will think on how to add a name check.

elezar: > let's hold this PR until:

Should this PR not just work automatically with those PRs -- assuming those hooks can be opted out of using the same option?

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -176,6 +177,12 @@ func (m command) build() *cli.Command {
 			Usage:       "Specify a pattern the CSV mount specifications.",
 			Destination: &opts.csv.ignorePatterns,
 		},
+		&cli.StringSliceFlag{
+			Name:        "skip-hook",
```
elezar, line 181: What about `disable-hook` (we could even add an alias as `disable-hooks`) instead? Thinking ahead, it may be useful to allow users to do something like:
```
nvidia-ctk cdi generate \
    --disable-hook=update-ldcache \
    --disable-hook=enable-cuda-compat \
    --mode=nvml
```

Even further forward, more conservative users may want to consider:
```
nvidia-ctk cdi generate --disable-hooks=all
```
although that is out of scope of this PR.

elezar: @klueska @jgehrcke  what are your thoughts on the UX?

cmd/nvidia-ctk/cdi/generate/generate.go
ArangoGutierrez: It now allows for disabling all hooks

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -276,7 +283,23 @@ func (m command) generateSpec(opts *options) (spec.Interface, error) {
 		nvcdi.WithCSVIgnorePatterns(opts.csv.ignorePatterns.Value()),
 		// We set the following to allow for dependency injection:
 		nvcdi.WithNvmlLib(opts.nvmllib),
-	)
+	}
+
+	if len(opts.disableHooks.Value()) > 0 {
+		for _, hook := range opts.disableHooks.Value() {
+			if hook == "all" {
```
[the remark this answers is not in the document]
ArangoGutierrez: Thanks for the idea, PTAL

pkg/nvcdi/api.go
```diff
@@ -44,4 +44,33 @@ const (
...
 	HookEnableCudaCompat = HookName("enable-cuda-compat")
+	// HookCreateSymlinks refers to the hook used create symlinks inside the
+	// directory path to be mounted into a container.
+	HookCreateSymlinks = HookName("create-symlinks")
+	// HookUpdateLDCache refers to the hook used to  Update the dynamic linker
+	// cache inside the directory path to be mounted into a container.
+	HookUpdateLDCache = HookName("update-ldcache")
 )
+
+// NewHookName takes a string and returns a []HookName, empty if the HookName
```
elezar, line 55: This is not the right place to implement this. we have a `pkg/nvcdi/hooks.go` file where this can go instead.

Also, in terms of the implementation, does adding special handling to the `disabledHooks` type for the `all` hook not make this cleaner?

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -276,7 +283,17 @@ func (m command) generateSpec(opts *options) (spec.Interface, error) {
...
-	)
+	}
+
+	if len(opts.disableHooks.Value()) > 0 {
+		for _, hook := range opts.disableHooks.Value() {
+			for _, hookName := range nvcdi.NewHookName(hook) {
+				initOpts = append(initOpts, nvcdi.WithDisabledHook(hookName))
+			}
+		}
+	}
```
elezar, line 294: The `len` check is not required. One can always iterate over an empty slice.
```suggestion
	for _, hook := range opts.disableHooks.Value() {
		initOpts = append(initOpts, nvcdi.WithDisabledHook(hook))
	}
```

(not that I have updated `WithDisabledHook` to:
```
// WithDisabledHook allows specific hooks to the disabled.
// This option can be specified multiple times for each hook.
func WithDisabledHook[T string | HookName](hook T) Option {
	return func(o *nvcdilib) {
		if o.disabledHooks == nil {
			o.disabledHooks = make(map[HookName]bool)
		}
		o.disabledHooks[HookName(hook)] = true
	}
}
```
to allow it to accept both `string` and `HookName` arguments.

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -57,6 +57,7 @@ type options struct {
 
 	configSearchPaths  cli.StringSlice
 	librarySearchPaths cli.StringSlice
+	disableHooks       cli.StringSlice
```
elezar, line 60:
```suggestion
	disabledHooks       cli.StringSlice
```

pkg/nvcdi/driver-nvml.go
```diff
@@ -99,21 +99,34 @@ func (l *nvcdilib) NewDriverLibraryDiscoverer(version string) (discover.Discover
 
 	var discoverers []discover.Discover
 
-	driverDotSoSymlinksDiscoverer := discover.WithDriverDotSoSymlinks(
-		libraries,
-		version,
-		l.nvidiaCDIHookPath,
-	)
-	discoverers = append(discoverers, driverDotSoSymlinksDiscoverer)
+	if l.HookIsSupported(HookCreateSymlinks) {
```
elezar, line 102: For this hook specifically (and the update-ldcache hook too), there are many places in the code where we are injecting this hook. As such this is not going to be sufficient to prevent their injection. This was why I suggested that handling these hooks is out of scope for this PR.

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -176,6 +177,12 @@ func (m command) build() *cli.Command {
 			Usage:       "Specify a pattern the CSV mount specifications.",
 			Destination: &opts.csv.ignorePatterns,
 		},
+		&cli.StringSliceFlag{
+			Name:        "disable-hook",
```
klueska, line 181: Should we also add a disable-all-hooks flag so people don't have to enumerate all of them to disable them.

Also, should we add a list-hooks command so people can discover what hooks are available for disabling

elezar: @klueska my thoughts on how to handle the disabling of all hooks is to also accept `all` as a valid hook name. So that `--disable-hook(s)=all` can be specified.

elezar: > Also, should we add a list-hooks command so people can discover what hooks are available for disabling

Do you mean something like:
```
nvidia-ctk cdi list-hooks
```?

Comment (ArangoGutierrez) [reply cmd/nvidia-ctk/cdi/generate/generate.go:181]: I have added an alias of `disable-hooks` and a better Usage description.

File: cmd/nvidia-ctk/cdi/generate/generate.go (hunk 8 above)
Comment (ArangoGutierrez) [reply cmd/nvidia-ctk/cdi/generate/generate.go]: Done

File: cmd/nvidia-ctk/cdi/generate/generate.go (hunk 7 above)
Comment (ArangoGutierrez) [reply cmd/nvidia-ctk/cdi/generate/generate.go]: Suggestion taken

**Comment by ArangoGutierrez** [comment]
PR https://github.com/NVIDIA/nvidia-container-toolkit/pull/1090 will refactor how we handle the Hook Creation, and after it, I'll update this PR to build on that and provide a better handling of disabled hooks via flags

File: internal/discover/hooks.go [hunk 11]
```diff
@@ -44,26 +44,66 @@ func (h *Hook) Hooks() ([]Hook, error) {
...
+	disabledHooks     DisabledHooks
 }
 
 type HookCreator interface {
-	Create(string, ...string) *Hook
+	Create(HookName, ...string) *Hook
 }
 
-func NewHookCreator(nvidiaCDIHookPath string) HookCreator {
+func NewHookCreator(nvidiaCDIHookPath string, disabledHooks DisabledHooks) HookCreator {
```
elezar, line 93: What about using:
```
```suggestion
func NewHookCreator[T string|HookName](nvidiaCDIHookPath string, disabledHooks ...T) HookCreator {
```

The fact that we're using a map should be internal to this pacakge and not exposed through the interface.

internal/discover/hooks.go
```diff
@@ -44,26 +44,66 @@ func (h *Hook) Hooks() ([]Hook, error) {
...
+func (d DisabledHooks) Set(value HookName) {
+	if value == "all" {
+		for _, hook := range AllHooks {
+			d[hook] = true
+		}
+		return
+	}
+
+	d[value] = true
+}
```
elezar, line 61: Instead of this, we can add an `isDisabled(HookName)` function (something like we had before in `IsHookSupported` that has been removed). We can impliment it as follows:
```
// HookIsSupported checks whether a hook of the specified name is supported.
// Hooks must be explicitly disabled, meaning that if no disabled hooks are
// all hooks are supported.
func (d disabledHooks) isDisabled(h HookName) bool {
	if d["all"] {
		return true
	}
	return d[h]
} 
```

internal/discover/hooks.go
```diff
@@ -44,26 +44,66 @@ func (h *Hook) Hooks() ([]Hook, error) {
 	return []Hook{*h}, nil
 }
 
+type HookName string
+
+// DisabledHooks allows individual hooks to be disabled.
+type DisabledHooks map[HookName]bool
```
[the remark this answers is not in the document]
ArangoGutierrez: Set to private

internal/discover/hooks.go
ArangoGutierrez: func removed, it was not needed by moving to a options implementation

internal/discover/hooks.go
elezar: Looks good. I have added a commit on top that also sets the hook path as an option for consistency.

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -270,7 +270,7 @@ func (m command) generateSpec(opts *options) (spec.Interface, error) {
 		deviceNamers = append(deviceNamers, deviceNamer)
 	}
 
-	initOpts := []nvcdi.Option{
+	cdiOptions := []nvcdi.Option{
```
elezar, line 275: nit: `initOptions` was too generic.

cmd/nvidia-ctk/cdi/generate/generate_test.go
```diff
@@ -127,7 +124,7 @@ containerEdits:
 				vendor:        "example.com",
 				class:         "device",
 				driverRoot:    driverRoot,
-				disabledHooks: *disableHook1,
+				disabledHooks: valueOf(cli.NewStringSlice("enable-cuda-compat")),
```
elezar, line 133: The use of arbitrary names for the variables made the intent of the tests more difficult to understand. A `valueOf` helper allows us to return the value of a string slice and list the disabled hooks in the tests.

internal/discover/hooks.go
```diff
@@ -22,6 +22,29 @@ import (
 	"tags.cncf.io/container-device-interface/pkg/cdi"
 )
 
+// A HookName represents a supported CDI hooks.
+type HookName string
+
+const (
```
elezar, line 29: I made the following changes:
* moved the constants earlier in the file as this is common practice in golang
* renamed the types `*Hook` instead of `Hook*`. This allows the code at the callsite to read more naturally. It also seems more idiomatic (https://google.github.io/styleguide/go/decisions.html#constant-names).
* added an `AllHooks` constant.
* added a deprecated `ChmodHook` constant.

internal/discover/hooks.go
```diff
@@ -44,89 +67,128 @@ func (h *Hook) Hooks() ([]Hook, error) {
...
-	// HookEnableCudaCompat refers to the hook used to enable CUDA Forward Compatibility.
-	// This was added with v1.17.5 of the NVIDIA Container Toolkit.
-	HookEnableCudaCompat = HookName("enable-cuda-compat")
-	// directory path to be mounted into a container.
-	HookCreateSymlinks = HookName("create-symlinks")
-	// HookUpdateLDCache refers to the hook used to  Update the dynamic linker
-	// cache inside the directory path to be mounted into a container.
-	HookUpdateLDCache = HookName("update-ldcache")
-)
+type cdiHookCreator struct {
```
elezar, line 72: I made the type private and renamed it for clarity.

internal/discover/hooks.go
```diff
@@ -44,89 +67,128 @@ func (h *Hook) Hooks() ([]Hook, error) {
 	return []Hook{*h}, nil
 }
 
-type HookName string
-
-// disabledHooks allows individual hooks to be disabled.
-type disabledHooks map[HookName]bool
```
elezar: The type was not used since we were not attaching methods to it.

internal/discover/hooks.go
```diff
@@ -44,89 +67,128 @@ func (h *Hook) Hooks() ([]Hook, error) {
...
+type cdiHookCreator struct {
+	nvidiaCDIHookPath string
+	disabledHooks     map[HookName]bool
 
-// AllHooks maintains a future-proof list of all defined hooks.
-var AllHooks = []HookName{
-	HookEnableCudaCompat,
-	HookCreateSymlinks,
-	HookUpdateLDCache,
+	fixedArgs []string
```
elezar: Instead of checking the path on every `Create` invocation, we check it once at construction.

internal/discover/hooks.go
```diff
@@ -44,89 +67,128 @@ func (h *Hook) Hooks() ([]Hook, error) {
...
 }
 
-type Option func(*CDIHook)
+// An allDisabledHookCreator is a HookCreator that does not create any hooks.
```
elezar, line 80: Instead of checking `disabledHooks` on every call to `Create`, we instantiate a type that creates no hooks. (This is the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern)).

internal/discover/hooks.go
```diff
@@ -44,89 +67,128 @@ func (h *Hook) Hooks() ([]Hook, error) {
...
-	if name == "create-symlinks" {
+	return &Hook{
+		Lifecycle: cdi.CreateContainerHook,
+		Path:      c.nvidiaCDIHookPath,
+		Args:      append(c.requiredArgs(name), c.transformArgs(name, args...)...),
+	}
+}
+
+// isDisabled checks if the specified hook name is disabled.
+func (c cdiHookCreator) isDisabled(name HookName, args ...string) bool {
```
elezar, line 144: Moved the `isDisabled` check to a method and made this dependent on the `args` as well. This allows for the `create-symlinks` and `chmod` hooks to be handled more cleanly.

internal/discover/hooks.go
```diff
@@ -44,89 +67,128 @@ func (h *Hook) Hooks() ([]Hook, error) {
...
 
-	return &Hook{
-		Lifecycle: cdi.CreateContainerHook,
-		Path:      c.nvidiaCDIHookPath,
-		Args:      append(c.requiredArgs(string(name)), args...),
+func (c cdiHookCreator) requiredArgs(name HookName) []string {
+	return append(c.fixedArgs, string(name))
+}
+
+func (c cdiHookCreator) transformArgs(name HookName, args ...string) []string {
```
elezar, line 166: Added a function to handle the per-hook transforms.

internal/discover/hooks.go
elezar, line 144: Moved the `isDisabled` check to a method and made this dependent on the `args` as well. This allows for the `create-symlinks` and `chmod` hooks to be handled more cleanly.

internal/discover/symlinks_test.go
```diff
@@ -113,7 +113,7 @@ func TestWithWithDriverDotSoSymlinks(t *testing.T) {
 			expectedHooks: []Hook{
 				{
 					Lifecycle: "createContainer",
-					Path:      "/path/to/nvidia-cdi-hook",
+					Path:      "/usr/bin/nvidia-cdi-hook",
```
elezar, line 116: Using a specific path here wasn't adding value. Changed the tests to check the default.

elezar (approved):

PR #1078, opened 2025-05-13
Fix mode detection on Thor-based systems

elezar:
This change adds support for the detection of Thor-based systems as Tegra devices which use `csv` mode by default.

This was added to go-nvlib in https://github.com/NVIDIA/go-nvlib/pull/60


ArangoGutierrez (approved):
LGTM

PR #1097, opened 2025-05-21
Create an internal/hooks pkg to centralize hook management

ArangoGutierrez:
This is an intermediate step towards #1074 

The idea is to have a central place to handle Hooks, and HookName, current location is not the works but this PR proposes a long term solution to allow a single place to add Hook functionality and HookNames


elezar (commented):
This change has complicated the code significantly. We went from having a single well-defined concept of what a hook is from the point of view of discovery and being able to reuse this across the code by single calls to `hookCreator.Create` to a state where we have to implement the `Discover` interface wherever we are generating hooks. This means that if we were to add fields to the hook struct, for example, we would have to change all these occurrences.

ArangoGutierrez:
> This change has complicated the code significantly. We went from having a single well-defined concept of what a hook is from the point of view of discovery and being able to reuse this across the code by single calls to `hookCreator.Create` to a state where we have to implement the `Discover` interface wherever we are generating hooks. This means that if we were to add fields to the hook struct, for example, we would have to change all these occurrences.

Agree, I haven't thought on that.

Issue #1091, opened 2025-05-16
[rpm] libnvidia-container-tools should pin to nvidia-container-toolkit version


seemethere:
https://github.com/NVIDIA/nvidia-container-toolkit/blob/ac8f190c991a352733b5f94fc931f350b26756f3/packaging/rpm/SPECS/nvidia-container-toolkit.spec#L24-L25

The latest release of nvidia-container-toolkit bricked a lot of jobs on pytorch's CUDA CI (see [example log](https://github.com/pytorch/pytorch/actions/runs/15070809681/job/42373911963#step:17:1476)) because it mistakenly upgraded the following packages:

* nvidia-container-tools
* libnvidia-container1
* nvidia-container-toolkit-base

The issue manifested itself with containers being unable to access GPU resources and thus we silently stopped running CUDA CI altogether (this will be remedied by https://github.com/pytorch/test-infra/pull/6638).

I'm creating this issue more as a discussion point to check int to see if these dependencies can be pinned.

If they can be pinned I'll happily submit over a PR but wanted to get context on why they were not before in the past and if we should have a reasonable expectation that mismatched versions of these packages should work.

elezar:
I wanted to add a note for completeness regarding the user experience of the proposed change for our Debian packages. Since we already have a strictly equals dependency between the `nvidia-container-toolkit` and `nvidia-container-toolkit-base` package, we can use this as a demonstrator.

In the case of the rpm packages, the dependency is properly resolved and running:
```
dnf install nvidia-container-toolkit-1.16.2-1
```
also installs the `nvidia-container-toolkit-base-1.16.2-1` package. (it stands to reason that this can be done for the `libnvidia-container-tools` and `libnvidia-container1` packages as well to ensure that they need be explicitly pinned to the same version.

In the case of debian packages, however, running:
```
$ apt-get install nvidia-container-toolkit=1.16.2-1
```
shows the following output:
```
Reading package lists... Done
Building dependency tree
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 nvidia-container-toolkit : Depends: nvidia-container-toolkit-base (= 1.16.2-1) but 1.17.7-1 is to be installed
E: Unable to correct problems, you have held broken packages.
```
[...]

PR #1100, opened 2025-05-22
Require matching version of libnvidia-container-tools

elezar:
Since the nvidia-container-tools and libnvidia-container* packages are now released at the same time with the same version, a more restrictive version makes sense. Here we specifically require an equal version for the nvidia-container-toolkit* packages and the libnvidia-container* packages.

For this to function properly the changes from https://github.com/NVIDIA/libnvidia-container/pull/317 are also required and the following should also be noted:

* On rpm-based systems `dnf` is able to properly resolve the dependencies and install all packages with a matching version.
* On deb-based systems `apt` is *not* able to resolve the dependencies properly and that means that if newer versions of dependencies (i.e. the `libnvidia-container-tools` and `libnvidia-container1` packages) are available an error will be shown if these are not *explicitly pinned*.

See https://github.com/NVIDIA/nvidia-container-toolkit/issues/1091

- [x] https://github.com/NVIDIA/libnvidia-container/pull/317

### DIFF
--- a/packaging/debian/control
+++ b/packaging/debian/control
@@ -10,7 +10,7 @@ Build-Depends: debhelper (>= 9)
 
 Package: nvidia-container-toolkit
 Architecture: any
-Depends: ${misc:Depends}, nvidia-container-toolkit-base (= @VERSION@), libnvidia-container-tools (>= @LIBNVIDIA_CONTAINER_TOOLS_VERSION@), libnvidia-container-tools (<< 2.0.0)
+Depends: ${misc:Depends}, nvidia-container-toolkit-base (= @VERSION@), libnvidia-container-tools (= @VERSION@), libnvidia-container-tools (<< 2.0.0)
 Breaks: nvidia-container-runtime (<= 3.5.0-1), nvidia-container-runtime-hook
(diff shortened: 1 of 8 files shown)


ArangoGutierrez (commented):
LGTM - Draft

packaging/debian/control
```diff
@@ -10,7 +10,7 @@ Build-Depends: debhelper (>= 9)
 
 Package: nvidia-container-toolkit
 Architecture: any
-Depends: ${misc:Depends}, nvidia-container-toolkit-base (= @VERSION@), libnvidia-container-tools (>= @LIBNVIDIA_CONTAINER_TOOLS_VERSION@), libnvidia-container-tools (<< 2.0.0)
+Depends: ${misc:Depends}, nvidia-container-toolkit-base (= @VERSION@), libnvidia-container-tools (= @VERSION@), libnvidia-container-tools (<< 2.0.0)
```
ArangoGutierrez, line 13: if we now require versions to be the same, does the check for the version to be lower than 2.0 still holds value? (i get the idea of having it as a double check, but the `=` part kinds of makes it straight clear)

elezar: Yes, we could drop the less than. I didn't want to change more than required. It's also good to be explicit in case we relax the requirements again in the future.

PR #1102, opened 2025-05-22
Edit discover.mounts to have a deterministic output

### LINKED ISSUE #1094: nvidia-ctk cdi generate output is not deterministic
After multiple runs of `nvidia-ctk cdi generate` the mount list has different order

```bash
diff run2 run1
26,27d25
<     - libGLX_nvidia.so.575.51.03::/usr/lib/x86_64-linux-gnu/libGLX_indirect.so.0
<     - --link
28a27,28
>     - --link
>     - libGLX_nvidia.so.575.51.03::/usr/lib/x86_64-linux-gnu/libGLX_indirect.so.0
```

`nvidia-ctk cdi generate` _should_ return a deterministic output for consistency.

ArangoGutierrez:
FIxes #1094

### DIFF
--- a/internal/discover/mounts_test.go
+++ b/internal/discover/mounts_test.go
@@ -45,13 +45,14 @@ func TestMounts(t *testing.T) {
 		"rprivate",
 	}
 
-	logger, logHook := testlog.NewNullLogger()
+	logger, _ := testlog.NewNullLogger()
 
 	testCases := []struct {
 		description    string
 		expectedError  error
 		expectedMounts []Mount
 		input          *mounts
+		repeat         int
 	}{
 		{
 			description:   "nill lookup returns error",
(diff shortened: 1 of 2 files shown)


internal/discover/mounts_test.go
```diff
@@ -188,3 +188,131 @@ func TestMounts(t *testing.T) {
...
+	// Verify that both discoverers return the same mounts
+	require.Equal(t, len(mounts1), len(mounts2), "Both discoverers should return the same number of mounts")
+
+	// Verify that the mounts are in the same order
+	for i := range mounts1 {
+		require.Equal(t, mounts1[i].Path, mounts2[i].Path, "Mounts should be in the same order")
+		require.Equal(t, mounts1[i].HostPath, mounts2[i].HostPath, "Host paths should match")
+		require.Equal(t, mounts1[i].Options, mounts2[i].Options, "Mount options should match")
+		require.Equal(t, mountOptions, mounts1[i].Options, "Mount should have correct options")
+	}
```
elezar, line 253: Does `require.EqualValues` not work here?

internal/discover/mounts_test.go
```diff
@@ -188,3 +188,131 @@ func TestMounts(t *testing.T) {
 		})
 	}
 }
+
+func TestMountsDeterministicOrder(t *testing.T) {
```
elezar, line 192: Did you verify that these tests were failing before the changes?

ArangoGutierrez: No, I added them after the changes, actually.

ArangoGutierrez: let me reverse the changes on mounts.go file and run the tests.... 1 sec

internal/discover/mounts_test.go
```diff
@@ -188,3 +188,131 @@ func TestMounts(t *testing.T) {
...
+			if s == "lib2" {
+				return []string{
+					"/usr/lib/libnvidia.so",
+					"/usr/lib/libGLX_indirect.so",
+					"/usr/lib/libGLX_nvidia.so",
+				}, nil
+			}
+			return []string{s}, nil
+		},
+	}
```
elezar, line 222: The issue was not that the locator was returning things in a non-deterministic order, but that we were using a MAP to deduplicate the returned values. When we construct the slice from the map keys the order could change.

internal/discover/mounts_test.go
ArangoGutierrez: checked, the tests should catch a regression

internal/discover/mounts_test.go
ArangoGutierrez: edited to use EqualValues

internal/discover/mounts_test.go
ArangoGutierrez: Agree, I have simplified this to a single return.

internal/discover/mounts_test.go
elezar: I don't think that a separate test case is required here. We can add a new entry to the table test in `TestMounts` that returns more candidates and replace:
```
require.ElementsMatch(t, tc.expectedMounts, mounts)
```
with
```
require.EqualValues(t, tc.expectedMounts, mounts)
```
since `ElementsMatch` ignores ordering.

ArangoGutierrez: thanks for the suggestion, added!

ArangoGutierrez:
@elezar PTAL

elezar:
Thanks @ArangoGutierrez. This looks good. Please also backport to the `release-1.17` branch.

Issue #1049, opened 2025-04-25
Gated modifiers ignore device requests by volume mounts


elezar:
The gated modifiers used to add support for GDS, Mofed, and CUDA Forward Comatibility only check the `NVIDIA_VISIBLE_DEVICES` envvar to determine whether GPUs are requested and modifications should be made. This means that use cases where volume mounts are used to request devices (e.g. when using the GPU Device Plugin) are not supported.

We need to update the check [here](https://github.com/NVIDIA/nvidia-container-toolkit/blob/de3d7366637c8b0646df66b2134884d3ceafaa9e/internal/modifier/gated.go#L40-L43) to include other request mechanism:
```go
	if devices := image.VisibleDevicesFromEnvVar(); len(devices) == 0 {
		logger.Infof("No modification required; no devices requested")
		return nil, nil
	}
```

PR #1110, opened 2025-05-28
Refactor extracting requested devices from the container image

ArangoGutierrez:
This change consolidates the logic for determining requested devices
from the container image. The logic for this has been integrated into
the image.CUDA type so that multiple implementations are not required.

We will address #1049 in a follow-up.


cmd/nvidia-container-runtime-hook/container_config.go
```diff
@@ -251,7 +224,7 @@ func (hookConfig *hookConfig) getDriverCapabilities(cudaImage image.CUDA, legacy
 func (hookConfig *hookConfig) getNvidiaConfig(image image.CUDA, privileged bool) *nvidiaConfig {
 	legacyImage := image.IsLegacy()
 
-	devices := hookConfig.getDevices(image, privileged)
+	devices := image.GetDevices(hookConfig.AcceptDeviceListAsVolumeMounts, hookConfig.AcceptEnvvarUnprivileged)
```
elezar: How about also passing these two options to the image constructor. So that we only call: `image.VisibleDevices` here instead of having to pass through the config.

I already had a rough branch from some time ago which I have pushed here: https://github.com/elezar/nvidia-container-toolkit/tree/respect-volume-mounts-gated

internal/config/image/builder.go
```diff
@@ -24,9 +24,12 @@ import (
 )
 
 type builder struct {
+	spec           *specs.Spec
 	env            map[string]string
 	mounts         []specs.Mount
 	disableRequire bool
+
+	swarmResourceEnvvars []string
```
elezar: This should not be `swarmResourceEnvvars`, but `visibleDeviceEnvvars`.

internal/config/image/builder.go
```diff
@@ -24,9 +24,12 @@ import (
 )
 
 type builder struct {
+	spec           *specs.Spec
```
elezar: I don't think that we should include a reference to the spec here -- or if we do we should remove the `env` and `mounts` below since they contain the same information -- or if they don't it is not clear which should take precedence.

internal/config/image/builder.go
```diff
@@ -100,3 +105,57 @@ func WithMounts(mounts []specs.Mount) Option {
...
+
+		b.swarmResourceEnvvars = append(b.swarmResourceEnvvars, resource...)
+		return nil
+	}
+}
+
+// WithPrivileged sets the privileged option for the CUDA image.
+// This is to allow testing the privileged mode of the container.
+// DO NOT USE THIS IN PRODUCTION CODE. FOR TESTING PURPOSES ONLY.
+func WithPrivileged(privileged bool) Option {
```
elezar: Why is this not for "Production"? Even in production we need to determine whether a container is privileged.

internal/config/image/cuda_image.go
```diff
@@ -271,6 +282,28 @@ func (i CUDA) DevicesFromMounts() []string {
 	return devices
 }
 
+func (i CUDA) GetDevices(acceptDeviceListAsVolumeMounts, acceptEnvvarUnprivileged bool) []string {
```
elezar:
```suggestion
func (i CUDA) VisibleDevices() []string {
```

internal/config/image/cuda_image.go
```diff
@@ -271,6 +282,28 @@ func (i CUDA) DevicesFromMounts() []string {
...
+	// If enabled, try and get the device list from volume mounts first
+	if acceptDeviceListAsVolumeMounts {
+		devices := i.VisibleDevicesFromMounts()
```
elezar: We may want to make `VisibleDevicesFromMounts` private.

internal/config/image/cuda_image.go
```diff
@@ -219,6 +228,9 @@ func (i CUDA) OnlyFullyQualifiedCDIDevices() bool {
 // VisibleDevicesFromEnvVar returns the set of visible devices requested through
 // the NVIDIA_VISIBLE_DEVICES environment variable.
 func (i CUDA) VisibleDevicesFromEnvVar() []string {
+	if i.IsSwarmResource() {
+		return i.DevicesFromEnvvars(i.swarmResourceEnvvars...).List()
+	}
```
elezar: The current behaviour in `container_config` is sublty different. We return the devices for the first ENVVAR that is set in the list that is set. I don't think that's what `DevicesFromEnvvars` does.

internal/config/image/cuda_image.go
```diff
@@ -271,6 +282,28 @@ func (i CUDA) DevicesFromMounts() []string {
...
+	// If enabled, try and get the device list from volume mounts first
+	if acceptDeviceListAsVolumeMounts {
+		devices := i.VisibleDevicesFromMounts()
+		if len(devices) > 0 {
+			return devices
+		}
+	}
+
+	// Fallback to reading from the environment variable if privileges are correct
+	devices := i.VisibleDevicesFromEnvVar()
```
elezar: May want to make this private.

internal/config/image/cuda_image.go
elezar: Possibly out of scope for this change, but the same logic needs to be implemented for IMEX channels and CDI devices.

internal/config/image/cuda_image.go
elezar: Could you comment on this? Looking at the latest changes this was not addressed?

cmd/nvidia-container-runtime-hook/container_config_test.go
```diff
@@ -477,9 +477,19 @@ func TestGetNvidiaConfig(t *testing.T) {
 	}
 	for _, tc := range tests {
 		t.Run(tc.description, func(t *testing.T) {
-			image, _ := image.New(
+			opts := []image.Option{
```
elezar: Don't the results of this test ONLY depend on the CUDA image implementation? If this is the case, they should be moved to the `config/image` package.

elezar: OK. This specific one does depend on other code, but the one below could be moved.

cmd/nvidia-container-runtime-hook/container_config_test.go
```diff
@@ -622,12 +632,15 @@ func TestDeviceListSourcePriority(t *testing.T) {
 						},
 					),
 					image.WithMounts(tc.mountDevices),
+					image.WithPrivileged(tc.privileged),
+					image.WithAcceptDeviceListAsVolumeMounts(tc.acceptMounts),
+					image.WithAcceptEnvvarUnprivileged(tc.acceptUnprivileged),
 				)
 				defaultConfig, _ := config.GetDefault()
 				cfg := &hookConfig{defaultConfig}
```
elezar: `cfg` is unused.

cmd/nvidia-container-runtime-hook/hook_test.go
```diff
@@ -1,89 +0,0 @@
-package main
-
-import (
-	"encoding/json"
-	"testing"
-
-	"github.com/stretchr/testify/require"
-)
-
-func TestIsPrivileged(t *testing.T) {
```
elezar: I don't think these tests are covered in the new file `privileged_test.go`. We should readd them.

internal/modifier/gated.go
```diff
@@ -37,8 +37,9 @@ import (
 //
 // If not devices are selected, no changes are made.
 func NewFeatureGatedModifier(logger logger.Interface, cfg *config.Config, image image.CUDA, driver *root.Driver, hookCreator discover.HookCreator) (oci.SpecModifier, error) {
-	if devices := image.VisibleDevicesFromEnvVar(); len(devices) == 0 {
-		logger.Infof("No modification required; no devices requested")
+	devices := image.VisibleDevices()
+	if len(devices) == 0 {
+		logger.Debugf("No modification required; no devices requested")
```
elezar: This is a functional change and should at the very least be a separate commit from the refactoring which *should* only include non-functional changes.

internal/config/image/cuda_image.go
```diff
@@ -38,8 +38,17 @@ const (
 // a map of environment variable to values that can be used to perform lookups
 // such as requirements.
 type CUDA struct {
-	env    map[string]string
+	visibleDevicesEnvVars []string
+	annotationPrefixes    []string
```
elezar: As a nit: Let's move this towards the end of the struct so that we logically separate the input (env, annotations, mounts) and settings that control behaviour.

internal/config/image/cuda_image.go
```diff
@@ -78,9 +89,14 @@ func (i CUDA) HasEnvvar(key string) bool {
 // image is considered legacy if it has a CUDA_VERSION environment variable defined
 // and no NVIDIA_REQUIRE_CUDA environment variable defined.
 func (i CUDA) IsLegacy() bool {
-	legacyCudaVersion := i.env[EnvVarCudaVersion]
-	cudaRequire := i.env[EnvVarNvidiaRequireCuda]
-	return len(legacyCudaVersion) > 0 && len(cudaRequire) == 0
+	cudaVersion := i.Getenv(EnvVarCudaVersion)
```
elezar: What was the reason for the switch to `Getenv`? I understand that we may be using things inconsistently here, but let's rather address that as a follow-up to keep the refactoring diff smaller?

internal/config/image/cuda_image.go
```diff
@@ -78,9 +89,14 @@ func (i CUDA) HasEnvvar(key string) bool {
...
+	cudaRequire := i.Getenv(EnvVarNvidiaRequireCuda)
+	return len(cudaVersion) > 0 && len(cudaRequire) == 0
+}
+
+// IsSwarm returns whether the image is a Docker Swarm image.
+func (i CUDA) IsSwarmResource() bool {
+	return len(i.visibleDevicesEnvVars) > 0
```
elezar: This is not swarm specific. I don't think this function is useful.

internal/config/image/cuda_image.go
```diff
@@ -111,24 +127,25 @@ func (i CUDA) GetRequirements() ([]string, error) {
 // HasDisableRequire checks for the value of the NVIDIA_DISABLE_REQUIRE. If set
 // to a valid (true) boolean value this can be used to disable the requirement checks
 func (i CUDA) HasDisableRequire() bool {
-	if disable, exists := i.env[EnvVarNvidiaDisableRequire]; exists {
-		// i.logger.Debugf("NVIDIA_DISABLE_REQUIRE=%v; skipping requirement checks", disable)
+	disable := i.Getenv(EnvVarNvidiaDisableRequire)
+	if disable != "" {
```
elezar: Please revert.

internal/config/image/cuda_image.go
```diff
@@ -111,24 +127,25 @@ func (i CUDA) GetRequirements() ([]string, error) {
...
 		d, _ := strconv.ParseBool(disable)
 		return d
 	}
-
 	return false
 }
 
 // DevicesFromEnvvars returns the devices requested by the image through environment variables
 func (i CUDA) DevicesFromEnvvars(envVars ...string) VisibleDevices {
```
elezar: Why the changes to this function? If we're going to clean it up, let's make sure that this is a separate commit or even PR.

internal/config/image/cuda_image.go
```diff
@@ -219,11 +234,14 @@ func (i CUDA) OnlyFullyQualifiedCDIDevices() bool {
 // VisibleDevicesFromEnvVar returns the set of visible devices requested through
 // the NVIDIA_VISIBLE_DEVICES environment variable.
 func (i CUDA) VisibleDevicesFromEnvVar() []string {
+	if i.IsSwarmResource() {
+		return i.DevicesFromEnvvars(i.visibleDevicesEnvVars...).List()
+	}
```
elezar: Considering how this is implemented in `container_config`, I believe this should be:
```
	for _, envvar := range i.visibleDevicesEnvVars {
		if _, ok := i.env[envvar]; ok {
			return i.DevicesFromEnvvars(envvar).List()
		}
	}
```
(I think there was a bug in the original implementation, but I will have to confirm).

internal/config/image/cuda_image.go
elezar: This is not used, so remove this from the change set.
```suggestion
```

internal/config/image/builder.go
```diff
@@ -50,15 +53,36 @@ func (b builder) build() (CUDA, error) {
 		b.env[EnvVarNvidiaDisableRequire] = "true"
 	}
 
-	c := CUDA{
-		env:    b.env,
-		mounts: b.mounts,
+	return b.CUDA, nil
+}
+
+func WithAnnotationPrefixes(annotationPrefixes []string) Option {
```
elezar: remove this option from the changeset.

internal/config/image/cuda_image.go
```diff
@@ -38,8 +38,17 @@ const (
 // a map of environment variable to values that can be used to perform lookups
 // such as requirements.
 type CUDA struct {
-	env    map[string]string
+	visibleDevicesEnvVars []string
+	annotationPrefixes    []string
+
+	env         map[string]string
+	annotations map[string]string
```
elezar: Unless we're going to add the logic to get devices from annotations in this PR we should remove this field from the change set.

internal/config/image/builder.go
```diff
@@ -50,15 +53,36 @@ func (b builder) build() (CUDA, error) {
...
+	return func(b *builder) error {
+		b.annotationPrefixes = annotationPrefixes
+		return nil
 	}
-	return c, nil
 }
 
-// Option is a functional option for creating a CUDA image.
-type Option func(*builder) error
+func WithAnnotations(annotations map[string]string) Option {
```
elezar: Remove from this changeset.

internal/config/image/builder.go
```diff
@@ -100,3 +124,35 @@ func WithMounts(mounts []specs.Mount) Option {
...
+		return nil
+	}
+}
+
+// WithVisibleDevicesEnvVars sets the swarm resource for the CUDA image.
+func WithVisibleDevicesEnvVars(visibleDevicesEnvVars ...string) Option {
+	return func(b *builder) error {
+		if len(visibleDevicesEnvVars) == 0 {
+			return fmt.Errorf("visible devices env vars cannot be empty")
+		}
```
elezar: Why is it an error if the list of envvars is empty?

The implementation (that is still present in `hook_config.go`) is:
```
// getSwarmResourceEnvvars returns the swarm resource envvars for the config.
func (c *hookConfig) getSwarmResourceEnvvars() []string {
	if c.SwarmResource == "" {
		return nil
	}

	candidates := strings.Split(c.SwarmResource, ",")

	var envvars []string
	for _, c := range candidates {
		trimmed := strings.TrimSpace(c)
		if len(trimmed) > 0 {
			envvars = append(envvars, trimmed)
		}
	}

	return envvars
}
```

internal/config/image/builder.go
```diff
@@ -100,3 +124,35 @@ func WithMounts(mounts []specs.Mount) Option {
...
+			candidates := strings.Split(visibleDevicesEnvVars[0], ",")
+			for _, c := range candidates {
+				trimmed := strings.TrimSpace(c)
+				if len(trimmed) > 0 {
+					b.visibleDevicesEnvVars = append(b.visibleDevicesEnvVars, trimmed)
+				}
+			}
+			return nil
+		}
+
```
elezar: Could you explain this logic to me? This seems different from the previous implementation of `getSwarmResourceEnvvars()`.

internal/config/image/builder.go
ArangoGutierrez: I have removed the error return

internal/config/image/builder.go
ArangoGutierrez: is just a bit different, `getSwarmResourceEnvvars` works with a single string vs `WithVisibleDevicesEnvVars` can handle a slice of string, but the same concept is used.

internal/config/image/cuda_image.go
ArangoGutierrez: I have removed both annotationPrefixes and annotations from this PR

internal/config/image/cuda_image.go
ArangoGutierrez: I have removed both annotationPrefixes and annotations from this PR

internal/config/image/cuda_image.go
```diff
@@ -216,14 +224,17 @@ func (i CUDA) OnlyFullyQualifiedCDIDevices() bool {
...
 }
 
-// VisibleDevicesFromEnvVar returns the set of visible devices requested through
+// visibleDevicesFromEnvVar returns the set of visible devices requested through
 // the NVIDIA_VISIBLE_DEVICES environment variable.
-func (i CUDA) VisibleDevicesFromEnvVar() []string {
+func (i CUDA) visibleDevicesFromEnvVar() []string {
+	if len(i.visibleDevicesEnvVars) > 0 {
+		return i.DevicesFromEnvvars(i.visibleDevicesEnvVars...).List()
+	}
```
elezar: This is not the logic as it was implemented.

internal/modifier/cdi.go
```diff
@@ -92,7 +92,7 @@ func getDevicesFromSpec(logger logger.Interface, ociSpec oci.Spec, cfg *config.C
 
 	var devices []string
 	seen := make(map[string]bool)
-	for _, name := range container.VisibleDevicesFromEnvVar() {
+	for _, name := range container.VisibleDevices() {
```
elezar: Please note that we already handle `CDIDevicesFromMounts` above, meaning that we may be duplicating logic here. Let's rather not make this change as part of this changeset.

internal/modifier/csv.go
```diff
@@ -33,7 +33,7 @@ import (
 // NewCSVModifier creates a modifier that applies modications to an OCI spec if required by the runtime wrapper.
 // The modifications are defined by CSV MountSpecs.
 func NewCSVModifier(logger logger.Interface, cfg *config.Config, container image.CUDA) (oci.SpecModifier, error) {
-	if devices := container.VisibleDevicesFromEnvVar(); len(devices) == 0 {
+	if devices := container.VisibleDevices(); len(devices) == 0 {
```
elezar: Let use FIRST confirm that our refactoring didn't break anything before making this change.

internal/modifier/gated.go
```diff
@@ -37,7 +37,7 @@ import (
 //
 // If not devices are selected, no changes are made.
 func NewFeatureGatedModifier(logger logger.Interface, cfg *config.Config, image image.CUDA, driver *root.Driver, hookCreator discover.HookCreator) (oci.SpecModifier, error) {
-	if devices := image.VisibleDevicesFromEnvVar(); len(devices) == 0 {
+	if devices := image.VisibleDevices(); len(devices) == 0 {
```
elezar: Let use FIRST confirm that our refactoring didn't break anything before making this change.

internal/modifier/graphics.go
```diff
@@ -65,7 +65,7 @@ func NewGraphicsModifier(logger logger.Interface, cfg *config.Config, containerI
 
 // requiresGraphicsModifier determines whether a graphics modifier is required.
 func requiresGraphicsModifier(cudaImage image.CUDA) (bool, string) {
-	if devices := cudaImage.VisibleDevicesFromEnvVar(); len(devices) == 0 {
+	if devices := cudaImage.VisibleDevices(); len(devices) == 0 {
```
elezar: Let use FIRST confirm that our refactoring didn't break anything before making this change.

cmd/nvidia-container-runtime-hook/container_config.go
```diff
@@ -306,20 +256,30 @@ func (hookConfig *hookConfig) getContainerConfig() (config containerConfig) {
...
 		image.WithEnv(s.Process.Env),
 		image.WithMounts(s.Mounts),
+		image.WithPrivileged(privileged),
 		image.WithDisableRequire(hookConfig.DisableRequire),
-	)
+		image.WithAcceptDeviceListAsVolumeMounts(hookConfig.AcceptDeviceListAsVolumeMounts),
+		image.WithAcceptEnvvarUnprivileged(hookConfig.AcceptEnvvarUnprivileged),
+	}
+
+	if hookConfig.SwarmResource == "" {
```
elezar: This condition doesn't look correct.

internal/config/image/builder.go
elezar: Why not just split every string by commas?

cmd/nvidia-container-runtime-hook/container_config.go
```diff
@@ -306,20 +256,30 @@ func (hookConfig *hookConfig) getContainerConfig() (config containerConfig) {
...
+		opts = append(opts, image.WithVisibleDevicesEnvVars(hookConfig.SwarmResource))
```
elezar: This is the cause of the test failures. After calling this we have `image.visibleDeviceEnvVars = []string{""}`

internal/config/image/cuda_image.go
elezar: We only call `DevicesFromEnvvars` if the envvar exists in the image.

cmd/nvidia-container-runtime-hook/hook_config_test.go
```diff
@@ -107,52 +106,3 @@ func TestGetHookConfig(t *testing.T) {
 		})
 	}
 }
-
-func TestGetSwarmResourceEnvvars(t *testing.T) {
```
elezar: Was the test for this logic moved somewhere?

internal/modifier/cdi.go
ArangoGutierrez: That would mean not making VisibleDevicesFromEnvVar private for now

cmd/nvidia-container-runtime-hook/hook_config_test.go
elezar: (This should be what `WithVisibleDeviceEnvvars` does).

internal/config/image/builder.go
```diff
@@ -100,3 +111,31 @@ func WithMounts(mounts []specs.Mount) Option {
...
+// WithPrivileged sets whether an image is privileged or not.
+func WithPrivileged(isPrivileged bool) Option {
+	return func(b *builder) error {
+		b.isPrivileged = isPrivileged
+		return nil
+	}
+}
+
+// WithVisibleDevicesEnvVars sets the swarm resource for the CUDA image.
```
elezar: Update this comment.

internal/config/image/builder.go
elezar: Also. Please add unit tests for this logic. I would say that we want the following behaviour:

For each envvar passed:
1. Split on commas:
2. trim spaces and ignore empty strings
3. Concatenate all of this

cmd/nvidia-container-runtime-hook/hook_config_test.go
ArangoGutierrez: yes, the function `getSwarmResourceEnVars` is removed, replaced by `WithVisibleDeviceEnvvars` so this test is not needed. 

We now test that at `container_config_test.go` 

```go
	if len(tc.swarmResourceEnvvars) > 0 {
				opts = append(opts, image.WithVisibleDevicesEnvVars(tc.swarmResourceEnvvars...))
			}
```

internal/config/image/builder.go
ArangoGutierrez: updated

cmd/nvidia-container-runtime-hook/container_config.go
ArangoGutierrez: fixed

cmd/nvidia-container-runtime-hook/container_config.go
ArangoGutierrez: now the if check has been corrected

internal/modifier/graphics_test.go
```diff
@@ -92,6 +92,7 @@ func TestGraphicsModifier(t *testing.T) {
 		t.Run(tc.description, func(t *testing.T) {
 			image, _ := image.New(
 				image.WithEnvMap(tc.envmap),
+				image.WithAcceptEnvvarUnprivileged(true),
```
elezar: One thing that this calls out is that the *default* for `acceptEnvvarUnprivileged` is `true`, so maybe one should capture that in the image itself?

cmd/nvidia-container-runtime-hook/container_config.go
```diff
@@ -306,20 +256,30 @@ func (hookConfig *hookConfig) getContainerConfig() (config containerConfig) {
...
+		image.WithPrivileged(privileged),
 		image.WithDisableRequire(hookConfig.DisableRequire),
-	)
+		image.WithAcceptDeviceListAsVolumeMounts(hookConfig.AcceptDeviceListAsVolumeMounts),
+		image.WithAcceptEnvvarUnprivileged(hookConfig.AcceptEnvvarUnprivileged),
+	}
+
+	if hookConfig.SwarmResource != "" {
+		opts = append(opts, image.WithAdditionalVisibleDevicesEnvVars(hookConfig.SwarmResource))
+	}
```
elezar: Why not just?
```suggestion
		image.WithAcceptDeviceListAsVolumeMounts(hookConfig.AcceptDeviceListAsVolumeMounts),
		image.WithAcceptEnvvarUnprivileged(hookConfig.AcceptEnvvarUnprivileged),
		image.WithAdditionalVisibleDevicesEnvVars(getSwarmResourceEnvvars()),
	}
```

cmd/nvidia-container-runtime-hook/hook_config_test.go
elezar: I don't agree. The handling of whitespace and commas is no longer tested? If the logic was moved to `WithVisibleDeviceEnvvars`, then that should be tested.

internal/config/image/builder.go
```diff
@@ -100,3 +110,33 @@ func WithMounts(mounts []specs.Mount) Option {
...
+	return func(b *builder) error {
+		b.isPrivileged = isPrivileged
+		return nil
+	}
+}
+
+// WithAdditionalVisibleDevicesEnvVars sets the visible devices environment variables.
+// If the resource is a single string, it is split by comma and the resulting
+// strings are added to the visible devices environment variables.
+func WithAdditionalVisibleDevicesEnvVars(visibleDevicesEnvVars ...string) Option {
```
elezar: The more I think about it, the more I think we should make this a simple setter and handle the specifics for the SWARM envvars in the `hookConfig` type. That is to say, we undelete `getSwarmResourceEnvvars()` and then just pass the outpuit of that into `WithAdditionalVisibleDevicesEnvVars`.

internal/config/image/cuda_image.go
```diff
@@ -217,13 +225,17 @@ func (i CUDA) OnlyFullyQualifiedCDIDevices() bool {
...
 
 // VisibleDevicesFromEnvVar returns the set of visible devices requested through
-// the NVIDIA_VISIBLE_DEVICES environment variable.
+// the NVIDIA_VISIBLE_DEVICES environment variable or any variables specified
+// in visibleDevicesEnvVars.
 func (i CUDA) VisibleDevicesFromEnvVar() []string {
+	if len(i.additionalVisibleDevicesEnvVars) > 0 {
+		return i.DevicesFromEnvvars(i.additionalVisibleDevicesEnvVars...).List()
+	}
 	return i.DevicesFromEnvvars(EnvVarNvidiaVisibleDevices).List()
```
elezar: The implementation in `container_config.go` was:
```
	// We check if the image has at least one of the Swarm resource envvars defined and use this
	// if specified.
	for _, envvar := range swarmResourceEnvvars {
		if containerImage.HasEnvvar(envvar) {
			return containerImage.DevicesFromEnvvars(swarmResourceEnvvars...).List()
		}
	}

	return containerImage.VisibleDevicesFromEnvVar()
```
That is not the same as this. Note that the comment indicates the intent.

IF one of the swarm envvars is present in the image, we use all the specified envvars -- otherwise we fallback to `NVIDIA_VISIBLE_DEVICES`. If we don't do this, as soon as a user sets a swarm envvar in their config then all other non-swarm images will stop working as expected.

internal/modifier/graphics_test.go
ArangoGutierrez: not on this PR, I feel that could be an API change. don't you think? (I agree with the change)

cmd/nvidia-container-runtime-hook/container_config.go
ArangoGutierrez: Ummm yeah, no need to check if empty will be handled, good catch

internal/config/image/cuda_image.go
ArangoGutierrez: I have edited the func to be 
```go
func (i CUDA) VisibleDevicesFromEnvVar() []string {
	for _, envVar := range i.additionalVisibleDevicesEnvVars {
		if i.HasEnvvar(envVar) {
			return i.DevicesFromEnvvars(i.additionalVisibleDevicesEnvVars...).List()
		}
	}
	return i.DevicesFromEnvvars(EnvVarNvidiaVisibleDevices).List()
}
```

internal/config/image/builder.go
ArangoGutierrez: undeleted, done!

cmd/nvidia-container-runtime-hook/hook_config_test.go
ArangoGutierrez: test undeleted

internal/config/image/builder.go
ArangoGutierrez: func revisited and unit test added

internal/modifier/cdi.go
elezar: Sure. That comment was more of a todo

internal/modifier/graphics_test.go
elezar: That's not what I mean. I mean the default value for the member in the struct should be true.

elezar: Let me think on it some more

ArangoGutierrez: Yeah, I get that you mean setting the default value. I think having it default to true can be considered an API change, not a big one, but worth a commit on it's own

cmd/nvidia-container-runtime-hook/container_config.go
```diff
@@ -306,20 +256,27 @@ func (hookConfig *hookConfig) getContainerConfig() (config containerConfig) {
 
 	s := loadSpec(path.Join(b, "config.json"))
 
-	image, err := image.New(
+	privileged := isPrivileged(s)
+
+	opts := []image.Option{
```
elezar: Since we're no longer updating the options conditionally, we can pass them to the constructor directly.

cmd/nvidia-container-runtime-hook/container_config_test.go
```diff
@@ -843,10 +852,17 @@ func TestGetDevicesFromEnvvar(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.description, func(t *testing.T) {
-			image, _ := image.New(
+			opts := []image.Option{
```
elezar: The tests in `TestGetDevicesFromEnvvar` can be moved to `cuda_image_test.go` and renamed `TestGetVisibleDevices`. (note that strictly speaking this is only testing get visible devices from envvar unless we expand the test cases to also set the other options.

internal/config/image/builder.go
```diff
@@ -100,3 +110,32 @@ func WithMounts(mounts []specs.Mount) Option {
...
+			for _, c := range strings.Split(v, ",") {
+				trimmed := strings.TrimSpace(c)
+				if trimmed != "" {
+					result = append(result, trimmed)
+				}
+			}
+		}
+		b.additionalVisibleDevicesEnvVars = result
+		return nil
+	}
```
elezar:
```suggestion
	return func(b *builder) error {
		b.additionalVisibleDevicesEnvVars = visibleDevicesEnvVars
		return nil
	}
```
(also note that the argument name hasn't been changed to matche the member / function name).

internal/config/image/builder_test.go
```diff
@@ -0,0 +1,69 @@
+package image
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestWithAdditionalVisibleDevicesEnvVars(t *testing.T) {
```
elezar: Correct me if I'm wrong, but don't we now cover this in `getSwarmEnvvars`?

internal/config/image/cuda_image.go
```diff
@@ -38,8 +38,15 @@ const (
...
-	env    map[string]string
+	additionalVisibleDevicesEnvVars []string
+
+	env map[string]string
+
 	mounts []specs.Mount
+
+	acceptDeviceListAsVolumeMounts bool
+	acceptEnvvarUnprivileged       bool
+	isPrivileged                   bool
```
elezar:
```suggestion
	env map[string]string
	mounts []specs.Mount
 
	preferredVisibleDeviceEnvVars       []string
	acceptDeviceListAsVolumeMounts bool
	acceptEnvvarUnprivileged       bool
	isPrivileged                   bool
```
note that I think `preferredVisibleDevicesEnvVars` better indicates the intent.

internal/config/image/privileged.go
```diff
@@ -16,28 +16,45 @@
 
 package image
 
-import (
-	"github.com/opencontainers/runtime-spec/specs-go"
-)
+import "github.com/opencontainers/runtime-spec/specs-go"
```
elezar: Let's fix this formatting.

internal/modifier/graphics_test.go
elezar: An API change for what? The default for the config option is `true`. This mean that the DEFAULT behaviour for a CUDA image to allow the use of `NVIDIA_VISIBLE_DEVICES` -- this is what required this test change.

internal/config/image/cuda_image.go
```diff
@@ -155,7 +163,7 @@ func (i CUDA) DevicesFromEnvvars(envVars ...string) VisibleDevices {
 
 // GetDriverCapabilities returns the requested driver capabilities.
 func (i CUDA) GetDriverCapabilities() DriverCapabilities {
-	env := i.env[EnvVarNvidiaDriverCapabilities]
+	env := i.Getenv(EnvVarNvidiaDriverCapabilities)
```
elezar: Let's undo this change.

internal/config/image/cuda_image.go
```diff
@@ -166,7 +174,7 @@ func (i CUDA) GetDriverCapabilities() DriverCapabilities {
 }
 
 func (i CUDA) legacyVersion() (string, error) {
-	cudaVersion := i.env[EnvVarCudaVersion]
+	cudaVersion := i.Getenv(EnvVarCudaVersion)
```
elezar: Undo this change.

internal/config/image/cuda_image.go
```diff
@@ -217,13 +225,19 @@ func (i CUDA) OnlyFullyQualifiedCDIDevices() bool {
 }
 
 // VisibleDevicesFromEnvVar returns the set of visible devices requested through
-// the NVIDIA_VISIBLE_DEVICES environment variable.
+// the NVIDIA_VISIBLE_DEVICES environment variable or any variables specified
+// in visibleDevicesEnvVars.
```
elezar: This is still not accurate.

internal/config/image/cuda_image.go
```diff
@@ -271,6 +284,28 @@ func (i CUDA) DevicesFromMounts() []string {
 	return devices
 }
 
+func (i CUDA) VisibleDevices() []string {
```
elezar: Let's add a docstring with the intended behaviour.

internal/config/image/cuda_image.go
elezar: As called out, we can do this in a follow-up.

internal/config/image/cuda_image.go
```diff
@@ -271,6 +284,28 @@ func (i CUDA) DevicesFromMounts() []string {
...
+	// Fallback to reading from the environment variable if privileges are correct
+	devices := i.VisibleDevicesFromEnvVar()
+	if len(devices) == 0 {
+		return nil
+	}
+
+	if i.isPrivileged || i.acceptEnvvarUnprivileged {
+		return devices
+	}
+
```
elezar: The one thing that we're missing from the initial implementation is that we're not logging the fact that we're ignoring the devices requested as envvars.

Let's add a logger to the `CUDA` image as well and add the logging here:
```
	i.logger.Warningf("Ignoring devices specified in NVIDIA_VISIBLE_DEVICES in unprivileged container")
```

internal/config/image/builder.go
```diff
@@ -93,10 +114,38 @@ func WithEnvMap(env map[string]string) Option {
...
 		b.mounts = mounts
 		return nil
 	}
 }
+
+// WithPreferredVisibleDevicesEnvVars sets the visible devices environment variables.
+// For each envvar passed:
+//   - Split on commas
+//   - Trim spaces and ignore empty strings
+//   - Concatenate all of this
```
elezar: This comment is not accurate.

internal/config/image/cuda_image.go
```diff
@@ -38,27 +40,38 @@ const (
 // a map of environment variable to values that can be used to perform lookups
 // such as requirements.
 type CUDA struct {
+	logger logger.Interface
+
 	env    map[string]string
 	mounts []specs.Mount
+
```
elezar:
```suggestion
```

internal/config/image/cuda_image.go
```diff
@@ -38,27 +40,38 @@ const (
...
 
-	return New(
+	specOpts := []Option{
 		WithEnv(env),
 		WithMounts(spec.Mounts),
-	)
+		WithPrivileged(IsPrivileged((*OCISpec)(spec))),
+	}
+
+	return New(append(opts, specOpts...)...)
```
elezar, line 68: One thing about this behaviour is that it ASSUMES that the env and mounts from the spec are preferred. It would be good to add a test for this -- although that's not a blocker for this PR.

internal/config/image/cuda_image.go
```diff
@@ -216,14 +229,53 @@ func (i CUDA) OnlyFullyQualifiedCDIDevices() bool {
...
+	}
+
+	// If the container is privileged, or environment variable requests are
+	// allowed for unprivileged containers, these devices are returned.
+	if i.isPrivileged || i.acceptEnvvarUnprivileged {
+		return envVarDeviceRequests
+	}
+
+	// We log a warning if we are ignoring the environment variable requests.
+	i.logger.Warningf("Ignoring devices specified in NVIDIA_VISIBLE_DEVICES in unprivileged container")
```
elezar, line 258: technically this could be checking other envvars. I'm ok to leave this as is though.

internal/config/image/cuda_image.go
ArangoGutierrez: it does, but then the return appends the opts, so if the caller of `NewCUDAImageFromSpec` set's the `WithPreferred...` that will be present during the construction of the image

internal/config/image/cuda_image.go
ArangoGutierrez: You are right, because if the previous check is not fulfilled, we will land here, and if this is a SWARM use case for example we would also be ignoring `preferredVisibleDeviceEnvVars`

ArangoGutierrez: something like 

```go
	if len(i.preferredVisibleDeviceEnvVars) > 0 {
		i.logger.Warningf("Ignoring devices specified in %v in unprivileged container", i.preferredVisibleDeviceEnvVars)
	} else {
		i.logger.Warningf("Ignoring devices specified in NVIDIA_VISIBLE_DEVICES in unprivileged container")
	}
```

internal/config/image/cuda_image.go
elezar: I'm speaking specifically of `env` and `mounts`, not `preferredVisibleDevicesEnvVars`. The point of a test is to ensure that LATER modifications don't change behaviour.

internal/config/image/cuda_image.go
elezar: That does not match the logic that we use to determine which envvars to use.

We need:
```golang
	var ignoredEnvVars []string
	for _, envVar := range i.preferredVisibleDeviceEnvVars {
		if i.HasEnvvar(envVar) {
			ignoredEnvVars = append(ignoredEnvVars, envVar)
		}
	}
	if len(ignoredEnvVars) == 0 {
		ignoredEnvVars = append(ignoredEnvVars, EnvVarNvidiaVisibleDevices)
	}

	// We log a warning if we are ignoring the environment variable requests.
	i.logger.Warningf("Ignoring devices requested by environment variable(s) in unprivileged container: %v", strings.Join(ignoredEnvVars, ", "))
```

elezar: (strictly speaking we should factor this into a function and use that from `VisibleDevicesFromEnvVar`, but I don't want to waste cycles on this). Let's do this as a follow-up. Please create an issue to track.

internal/config/image/builder.go
```diff
@@ -93,10 +114,34 @@ func WithEnvMap(env map[string]string) Option {
...
+
 // WithMounts sets the mounts associated with the CUDA image.
 func WithMounts(mounts []specs.Mount) Option {
 	return func(b *builder) error {
 		b.mounts = mounts
 		return nil
 	}
 }
+
+// WithPreferredVisibleDevicesEnvVars sets the visible devices environment variables.
```
elezar:
```suggestion
// WithPreferredVisibleDevicesEnvVars sets the environment variables that should take precendence over the default NVIDIA_VISIBLE_DEVICES.
// If any of these are present in the image, they are used to determine which devices were requested.
```

internal/config/image/cuda_image.go
ArangoGutierrez: I'll create an issue once we merge the PR, so I can point to places in the code in the issue

internal/config/image/builder.go
ArangoGutierrez: done

internal/config/image/cuda_image.go
ArangoGutierrez: how about

```go
// visibleEnvVars returns the environment variables that are used to determine device visibility.
// It returns the preferred environment variables that are set, or NVIDIA_VISIBLE_DEVICES if none are set.
func (i CUDA) visibleEnvVars() []string {
	var envVars []string
	for _, envVar := range i.preferredVisibleDeviceEnvVars {
		if i.HasEnvvar(envVar) {
			envVars = append(envVars, envVar)
		}
	}
	if len(envVars) == 0 {
		envVars = append(envVars, EnvVarNvidiaVisibleDevices)
	}
	return envVars
}
```

(I have tested it and it works for both the logging use case and the `VisibleDevicesFromEnvVar` use case

elezar (approved):
Let's get this in.

internal/config/image/cuda_image.go
ArangoGutierrez: TODO: Add unit test for `NewCUDAImageFromSpec`

internal/config/image/cuda_image.go
ArangoGutierrez: TODO: add a patch with https://github.com/NVIDIA/nvidia-container-toolkit/pull/1110#discussion_r2128529835

PR #1118, opened 2025-05-30
[no-relnote] Add E2E for libnvidia-container

ArangoGutierrez:
This patch adds an E2E test for the nvidia-container-cli that will allow us to catch regressions on libnvidia-container

### DIFF
--- a/tests/e2e/Makefile
+++ b/tests/e2e/Makefile
@@ -20,8 +20,14 @@ LOG_ARTIFACTS_DIR ?= $(CURDIR)/e2e_logs
 
 GINKGO_BIN := $(CURDIR)/bin/ginkgo
 
+# If GINKGO_FOCUS is not set, run all tests
+# current available tests:
+# - nvidia-container-cli
+# - docker
+GINKGO_FOCUS ?=
+
 test: $(GINKGO_BIN)
-	$(GINKGO_BIN) $(GINKGO_ARGS) -v --json-report ginkgo.json ./tests/e2e/...
+	$(GINKGO_BIN) $(GINKGO_ARGS) -v --json-report ginkgo.json --focus="$(GINKGO_FOCUS)" ./tests/e2e/...
 
 # test-preinstalled runs the test cases against the version of the toolkit that
 # is already installed (and configured for docker) on the host.
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -64,9 +64,7 @@ func getTestEnv() {
 
 	if installCTK {
 		imageName = getRequiredEnvvar[string]("E2E_IMAGE_NAME")
-
 		imageTag = getRequiredEnvvar[string]("E2E_IMAGE_TAG")
-
 	}
 
 	sshHost = getEnvVarOrDefault("E2E_SSH_HOST", "")
--- a/tests/e2e/installer.go
+++ b/tests/e2e/installer.go
@@ -28,11 +28,18 @@ var dockerInstallTemplate = `
 #! /usr/bin/env bash
 set -xe
 
-: ${IMAGE:={{.Image}}}
+# if the TEMP_DIR is already set, use it
+if [ -f /tmp/ctk_e2e_temp_dir.txt ]; then
+    TEMP_DIR=$(cat /tmp/ctk_e2e_temp_dir.txt)
+else
+    TEMP_DIR="/tmp/ctk_e2e.$(date +%s)_$RANDOM"
+    echo "$TEMP_DIR" > /tmp/ctk_e2e_temp_dir.txt
+fi
 
-# Create a temporary directory
-TEMP_DIR="/tmp/ctk_e2e.$(date +%s)_$RANDOM"
-mkdir -p "$TEMP_DIR"
+# if TEMP_DIR does not exist, create it
+if [ ! -d "$TEMP_DIR" ]; then
+    mkdir -p "$TEMP_DIR"
+fi
 
 # Given that docker has an init function that checks for the existence of the
 # nvidia-container-toolkit, we need to create a symlink to the nvidia-container-runtime-hook
@@ -46,7 +53,7 @@ docker run --pid=host --rm -i --privileged	\
 	-v /var/run/docker.sock:/var/run/docker.sock	\
 	-v "$TEMP_DIR:$TEMP_DIR"	\
 	-v /etc/docker:/config-root	\
-	${IMAGE}	\
+	{{.Image}}	\
 	--root "$TEMP_DIR"	\
 	--runtime=docker	\
 	--config=/config-root/daemon.json	\
--- a/tests/e2e/nvidia-container-cli_test.go
+++ b/tests/e2e/nvidia-container-cli_test.go
@@ -114,17 +114,21 @@ unshare --mount --pid --fork --propagation private -- sh -eux <<'\''IN_NS'\''
 IN_NS
 `
 
-	dockerRunCmdTemplate = `docker run -d --name node-container-e2e --privileged --runtime=nvidia \
+	startTestContainerTemplate = `docker run -d --name {{.ContainerName}} --privileged --runtime=nvidia \
     -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=all \
     -e NVIDIA_DRIVER_CAPABILITIES=all \
+	{{ range $i, $a := .AdditionalArguments -}}
+	{{ $a }} \
+	{{ end -}}
 	ubuntu sleep infinity`
 )
 
 var _ = Describe("nvidia-container-cli", Ordered, ContinueOnFailure, Label("libnvidia-container"), func() {
 	var (
-		runner        Runner
-		containerName = "node-container-e2e"
-		hostOutput    string
+		runner                       Runner
+		containerName                = "node-container-e2e"
+		hostOutput                   string
+		additionalContainerArguments []string
 	)
 
 	BeforeAll(func(ctx context.Context) {
@@ -145,6 +149,21 @@ var _ = Describe("nvidia-container-cli", Ordered, ContinueOnFailure, Label("libn
 
 			err = installer.Install()
 			Expect(err).ToNot(HaveOccurred())
+		} else {
+			// If installCTK is false, we use the preinstalled toolkit.
+			// TODO: This should be updated for other distributions and other components of the toolkit.
+			output, _, err := runner.Run("ls /lib/**/libnvidia-container*.so.*.*")
+			Expect(err).ToNot(HaveOccurred())
+
+			output = strings.TrimSpace(output)
+			Expect(output).ToNot(BeEmpty())
+
+			for _, lib := range strings.Split(output, "\n") {
+				additionalContainerArguments = append(additionalContainerArguments, "-v "+lib+":"+lib)
+			}
+			additionalContainerArguments = append(additionalContainerArguments,
+				"-v /usr/bin/nvidia-container-cli:/usr/bin/nvidia-container-cli",
+			)
 		}
 
 		// Capture the host GPU list.
@@ -168,33 +187,46 @@ var _ = Describe("nvidia-container-cli", Ordered, ContinueOnFailure, Label("libn
 
 	It("should report the same GPUs inside the container as on the host", func(ctx context.Context) {
 		// Launch the container in detached mode.
-		_, _, err := runner.Run(dockerRunCmdTemplate)
+		var startContainerScriptBuilder strings.Builder
+		startContainerTemplate, err := template.New("startContainer").Parse(startTestContainerTemplate)
 		Expect(err).ToNot(HaveOccurred())
-
-		// Install docker and nvidia-container-toolkit in the container.
-		// Run as root and use bash for better compatibility
-		_, _, err = runner.Run(fmt.Sprintf("docker exec -u root %s bash -c '%s'", containerName, installDockerTemplate))
-		Expect(err).ToNot(HaveOccurred())
-
-		// Build the docker run command (detached mode) from the template so it
-		// stays readable while still resulting in a single-line invocation.
-		tmpl, err := template.New("toolkitInstall").Parse(installCTKTemplate)
-		Expect(err).ToNot(HaveOccurred())
-
-		var toolkitInstall strings.Builder
-		err = tmpl.Execute(&toolkitInstall, struct {
-			ToolkitImage string
+		err = startContainerTemplate.Execute(&startContainerScriptBuilder, struct {
+			ContainerName       string
+			AdditionalArguments []string
 		}{
-			ToolkitImage: imageName + ":" + imageTag,
+			ContainerName:       containerName,
+			AdditionalArguments: additionalContainerArguments,
 		})
 		Expect(err).ToNot(HaveOccurred())
 
-		_, _, err = runner.Run(fmt.Sprintf("docker exec -u root %s bash -c '%s'", containerName, toolkitInstall.String()))
+		startContainerScript := startContainerScriptBuilder.String()
+		GinkgoLogr.Info("Starting test container", "script", startContainerScript)
+		_, _, err = runner.Run(startContainerScript)
 		Expect(err).ToNot(HaveOccurred())
 
+		// Install docker in the container.
+		_, _, err = runner.Run(fmt.Sprintf("docker exec -u root "+containerName+" bash -c '%s'", installDockerTemplate))
+		Expect(err).ToNot(HaveOccurred())
+
+		if installCTK {
+			// Install nvidia-container-cli in the container.
+			tmpl, err := template.New("toolkitInstall").Parse(installCTKTemplate)
+			Expect(err).ToNot(HaveOccurred())
+
+			var toolkitInstall strings.Builder
+			err = tmpl.Execute(&toolkitInstall, struct {
+				ToolkitImage string
+			}{
+				ToolkitImage: imageName + ":" + imageTag,
+			})
+			Expect(err).ToNot(HaveOccurred())
+
+			_, _, err = runner.Run(fmt.Sprintf("docker exec -u root "+containerName+" bash -c '%s'", toolkitInstall.String()))
+			Expect(err).ToNot(HaveOccurred())
+		}
+
 		// Run the test script in the container.
-		// Capture but don't fail on errors - we'll check the results via container logs.
-		output, _, err := runner.Run(fmt.Sprintf("docker exec -u root %s bash -c '%s'", containerName, libnvidiaContainerCliTestTemplate))
+		output, _, err := runner.Run(fmt.Sprintf("docker exec -u root "+containerName+" bash -c '%s'", libnvidiaContainerCliTestTemplate))
 		Expect(err).ToNot(HaveOccurred())
 		Expect(strings.TrimSpace(output)).ToNot(BeEmpty())
 		Expect(hostOutput).To(ContainSubstring(strings.TrimSpace(output)))
--- a/tests/e2e/nvidia-container-cli_test.go
+++ b/tests/e2e/nvidia-container-cli_test.go
@@ -1,0 +1,202 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"text/template"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	installDockerTemplate = `
+export DEBIAN_FRONTEND=noninteractive
+
+# Add Docker official GPG key:
+apt-get update
+apt-get install -y ca-certificates curl apt-utils gnupg2
+install -m 0755 -d /etc/apt/keyrings
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
+chmod a+r /etc/apt/keyrings/docker.asc
+
+# Add the repository to Apt sources:
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu $(. /etc/os-release && echo \"${UBUNTU_CODENAME:-$VERSION_CODENAME}\") stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null
+apt-get update
+
+apt-get install -y docker-ce docker-ce-cli containerd.io
+
+# start dockerd in the background
+dockerd &
+
+# wait for dockerd to be ready with timeout
+timeout=30
+elapsed=0
+while ! docker info > /dev/null 2>&1 && [ $elapsed -lt $timeout ]; do
+    echo "Waiting for dockerd to be ready..."
+    sleep 1
+    elapsed=$((elapsed + 1))
+done
+if [ $elapsed -ge $timeout ]; then
+    echo "Docker failed to start within $timeout seconds"
+    exit 1
+fi
+`
+	installCTKTemplate = `
+# Create a temporary directory and rootfs path
+TMPDIR="$(mktemp -d)"
+
+# Expose TMPDIR for the child namespace
+export TMPDIR
+
+docker run --rm -v ${TMPDIR}:/host-tmpdir --entrypoint="sh" {{.ToolkitImage}}-packaging -c "cp -p -R /artifacts/* /host-tmpdir/"
+dpkg -i ${TMPDIR}/packages/ubuntu18.04/amd64/libnvidia-container1_*_amd64.deb ${TMPDIR}/packages/ubuntu18.04/amd64/nvidia-container-toolkit-base_*_amd64.deb ${TMPDIR}/packages/ubuntu18.04/amd64/libnvidia-container-tools_*_amd64.deb
+
+nvidia-container-cli --version
+`
+
+	libnvidiaContainerCliTestTemplate = `
+# Create a temporary directory and rootfs path
+TMPDIR="$(mktemp -d)"
+ROOTFS="${TMPDIR}/rootfs"
+mkdir -p "${ROOTFS}"
+
+# Expose ROOTFS for the child namespace
+export ROOTFS TMPDIR
+
+# Download Ubuntu base image with error handling
+curl -fsSL http://cdimage.ubuntu.com/ubuntu-base/releases/22.04/release/ubuntu-base-22.04-base-amd64.tar.gz | tar -C $ROOTFS -xz || {
+    echo "Failed to download or extract Ubuntu base image"
+    exit 1
+}
+
+# Enter a new mount + PID namespace so we can pivot_root without touching the
+# container'\''s original filesystem.
+unshare --mount --pid --fork --propagation private -- sh -eux <<'\''IN_NS'\''
+  : "${ROOTFS:?}"
+
+  # 1 Bind-mount the new root and make the mount private
+  mount --bind "$ROOTFS" "$ROOTFS"
+  mount --make-private "$ROOTFS"
+  cd "$ROOTFS"
+
+  # 2 Minimal virtual filesystems
+  mount -t proc  proc proc
+  mount -t sysfs sys  sys
+  mount -t tmpfs tmp  tmp
+  mount -t tmpfs run  run
+
+  # 3 Configure NVIDIA devices
+  nvidia-container-cli --load-kmods configure --ldconfig=@/sbin/ldconfig.real --no-cgroups --utility --device 0 $(pwd)
+
+  # 4 Switch root into the prepared filesystem
+  pivot_root . mnt
+  umount -l mnt
+  nvidia-smi -L
+
+IN_NS
+`
+
+	dockerRunCmdTemplate = `docker run -d --name node-container-e2e --privileged --runtime=nvidia \
+    -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=all \
+    -e NVIDIA_DRIVER_CAPABILITIES=all \
+	ubuntu sleep infinity`
+)
+
+var _ = Describe("nvidia-container-cli", Ordered, ContinueOnFailure, Label("libnvidia-container"), func() {
+	var (
+		runner        Runner
+		containerName = "node-container-e2e"
+		hostOutput    string
+	)
+
+	BeforeAll(func(ctx context.Context) {
+		runner = NewRunner(
+			WithHost(sshHost),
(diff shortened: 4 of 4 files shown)


tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,165 @@
...
+  -v /usr/bin/nvidia-container-runtime:/usr/bin/nvidia-container-runtime \
+  -v /usr/bin/nvidia-container-runtime-hook:/usr/bin/nvidia-container-runtime-hook \
+  -v /usr/bin/nvidia-container-toolkit:/usr/bin/nvidia-container-toolkit \
+  -v /usr/bin/nvidia-cdi-hook:/usr/bin/nvidia-cdi-hook \
+  -v /usr/bin/nvidia-container-runtime.cdi:/usr/bin/nvidia-container-runtime.cdi \
+  -v /usr/bin/nvidia-container-runtime.legacy:/usr/bin/nvidia-container-runtime.legacy \
+  -v /usr/local/nvidia/toolkit:/usr/local/nvidia/toolkit \
+  -v /etc/nvidia-container-runtime:/etc/nvidia-container-runtime \
+  -v /usr/lib/x86_64-linux-gnu/libnvidia-container.so.1:/usr/lib/x86_64-linux-gnu/libnvidia-container.so.1 \
+  -v /usr/lib/x86_64-linux-gnu/libnvidia-container-go.so.1:/usr/lib/x86_64-linux-gnu/libnvidia-container-go.so.1 \
```
elezar, line 46: THe `.so.1` files are symlinks and should not be mounted.

ArangoGutierrez:
Tests pass, PR ready for review @elezar

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	libnvidiaContainerCliDockerRunTemplate = `
+docker run -d --name test-nvidia-container-cli \
+  --privileged \
+  --runtime=nvidia \
```
elezar: It is not scalable to have to MOUNT everything into this container. Note that when we still had some simple integration tests in the toolkit we used https://github.com/NVIDIA/nvidia-container-toolkit/blob/27f5ec83debaa8a173e90c43b013d5acad4b27a0/tests/container/docker_test.sh#L20

Can we rather adapt this?

ArangoGutierrez: sure, will work on that. we need to make this test more robust and scalable

tests/e2e/installer.go
```diff
@@ -28,11 +28,20 @@ var dockerInstallTemplate = `
...
+
+# if TEMP_DIR does not exist, create it
+if [ ! -d "$TEMP_DIR" ]; then
+    mkdir -p "$TEMP_DIR"
+fi
 
-# Create a temporary directory
-TEMP_DIR="/tmp/ctk_e2e.$(date +%s)_$RANDOM"
-mkdir -p "$TEMP_DIR"
+: ${IMAGE:={{.Image}}}
```
elezar: Nit: Why did we swap the ordering of these?

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,167 @@
...
+    -e NVIDIA_DRIVER_CAPABILITIES=all \
+    -v %s:/libnvidia-container-cli.sh \
+    --entrypoint /libnvidia-container-cli.sh %s`
+)
+
+var _ = Describe("nvidia-container-cli", Ordered, ContinueOnFailure, func() {
+	var (
+		runner        Runner
+		testScript    = "/tmp/libnvidia-container-cli.sh"
+		dockerImage   = "ghcr.io/nvidia/container-toolkit:5e8c1411-ubuntu20.04"
```
elezar: Why are we hardcoding this?

elezar: Also, does this break once we switch to distroless?

tests/e2e/e2e_test.go
```diff
@@ -62,12 +62,8 @@ func getTestEnv() {
...
 	installCTK = getEnvVarOrDefault("E2E_INSTALL_CTK", false)
 
-	if installCTK {
-		imageName = getRequiredEnvvar[string]("E2E_IMAGE_NAME")
-
-		imageTag = getRequiredEnvvar[string]("E2E_IMAGE_TAG")
-
-	}
+	imageName = getRequiredEnvvar[string]("E2E_IMAGE_NAME")
+	imageTag = getRequiredEnvvar[string]("E2E_IMAGE_TAG")
```
elezar: Is there a reason we removed the conditional?

tests/e2e/installer.go
```diff
@@ -28,11 +28,20 @@ var dockerInstallTemplate = `
 #! /usr/bin/env bash
 set -xe
 
-: ${IMAGE:={{.Image}}}
+# if the TEMP_DIR is already set, use it
+if [ -f /tmp/ctk_e2e_temp_dir.txt ]; then
```
elezar, line 32: Why are we managing this state at a shell-script level? Should we not create the temp dir once in the go code and then use it here in our template?

tests/e2e/installer.go
elezar: As a general note to the scripts -- why are we using envvars at all and don't we just use `{{.Image}}` everywhere? Is there a case where `IMAGE` is already set to something else?

tests/e2e/e2e_test.go
ArangoGutierrez: Yes, regardless of if we want to install or not the toolkit on the host, I want to be able to get these 2 variables.

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,167 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	// libnvidiaContainerCliTestTemplate is an on-the-fly script that prepares a
+	// minimal rootfs with GPU support and finally prints the list of visible GPUs
+	// using `nvidia-smi -L`.  The exit code as well as the standard output of this
+	// script are therefore a good indicator of whether the NVIDIA Container
+	// Toolkit is functioning correctly inside the container.
```
elezar: This is not what we're testing. We're testing the `nvidia-container-cli` specifically.

tests/e2e/installer.go
ArangoGutierrez: switched back

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,167 @@
...
+	libnvidiaContainerCliTestTemplate = `#!/usr/bin/env bash
+set -euo pipefail
+
+apt-get update -y && apt-get install -y curl gnupg2
+
+WORKDIR="$(mktemp -d)"
+ROOTFS="${WORKDIR}/rootfs"
```
elezar: Why do we need two directories? What about:
```suggestion
ROOTFS="$(mktemp -d)/rootfs"
```

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,167 @@
...
+    -e NVIDIA_VISIBLE_DEVICES=all \
+    -e NVIDIA_DRIVER_CAPABILITIES=all \
+    -v %s:/libnvidia-container-cli.sh \
+    --entrypoint /libnvidia-container-cli.sh %s`
+)
+
+var _ = Describe("nvidia-container-cli", Ordered, ContinueOnFailure, func() {
+	var (
+		runner        Runner
+		testScript    = "/tmp/libnvidia-container-cli.sh"
```
elezar: Are we guaranteed a single script across all tests?

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,167 @@
...
+		dockerImage   = "ghcr.io/nvidia/container-toolkit:5e8c1411-ubuntu20.04"
+		containerName = "nvidia-cli-e2e"
+		dockerRunCmd  string
```
elezar: Is a variable at this scope required?

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	It("should report the same GPUs inside the container as on the host", func(ctx context.Context) {
+		// Write the script to the remote host and make it executable.
+		createScriptCmd := fmt.Sprintf(
+			"cat > %s <<'EOF'\n%s\nEOF\nchmod +x %s",
+			testScript, libnvidiaContainerCliTestTemplate, testScript,
+		)
+
+		_, _, err := runner.Run(createScriptCmd)
+		Expect(err).ToNot(HaveOccurred())
```
elezar: Should this not be in a `BeforeAll`?

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,167 @@
...
+		// If a container with the same name exists from a previous test run, remove it first.
+		runner.Run(fmt.Sprintf("docker rm -f %s", containerName))
+
+		// Build the docker run command (detached mode) from the template so it
+		// stays readable while still resulting in a single-line invocation.
+		dockerRunCmd = fmt.Sprintf(dockerRunCmdTemplate, containerName, testScript, dockerImage)
+
+		// Launch the container in detached mode.
+		_, _, err = runner.Run(dockerRunCmd)
+		Expect(err).ToNot(HaveOccurred())
```
elezar: Should this also be `BeforeAll`?

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+		// Capture the host GPU list.
+		hostOutput, _, err := runner.Run("nvidia-smi -L")
+		Expect(err).ToNot(HaveOccurred())
+
+		hostOutput = strings.TrimSpace(strings.ReplaceAll(hostOutput, "\r", ""))
```
elezar: Why are we replacing returns with `""` and not `" "`?

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+		// Poll the logs of the already running container until we observe
+		// the GPU list matching the host or until a 5-minute timeout elapses.
+		Eventually(func() string {
+			logs, _, err := runner.Run(fmt.Sprintf("docker logs %s", containerName))
```
elezar: If we're just looking for the last log line, does `docker logs --tail 1` work?

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: No, as this comment made me realize https://github.com/NVIDIA/nvidia-container-toolkit/pull/1118#discussion_r2212695758 and https://github.com/NVIDIA/nvidia-container-toolkit/pull/1118#discussion_r2212698797 This test script is for this test case only. Yes, right now is the only test under this `Describe`, but I prefer to keep it local to the `It` so we are clear to add further test cases

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: same reason as in https://github.com/NVIDIA/nvidia-container-toolkit/pull/1118#discussion_r2212771096

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: We replace \r with the empty string to obtain a whitespace-minimal representation of each line before comparison; inserting a space would alter the content rather than merely normalizing it.

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: I was not aware of the flag, thanks!

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: The hardcoded image was a mistake; it was an experimental feature introduced by me to iterate.

On distroless, it should work as we are setting the `--entrypoint /libnvidia-container-cli.sh` , so as long as the distroless supports `/usr/bin/env bash` scripts, this test should work

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: agreed, added

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: I have now edited to adjust for the distroless image

tests/e2e/e2e_test.go
elezar: Why? What do we need the image for if we're not installing the toolkit?

ArangoGutierrez: for the test cases added in this PR, we need to pass the image into the runner container, so we can get the artifacts to install the toolkit and the `nvidia-container-cli`

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+# install nvidia-container-toolkit from NVIDIA_TOOLKIT_IMAGE using docker
+TEMP_DIR="/tmp/nvidia-container-toolkit"
+mkdir -p "$TEMP_DIR"
+rm -f /usr/bin/nvidia-container-runtime-hook
+ln -s "$TEMP_DIR/toolkit/nvidia-container-runtime-hook" /usr/bin/nvidia-container-runtime-hook
+
+docker run --rm -d --name nvidia-container-toolkit-packaging --entrypoint "/usr/bin/sleep" ${NVIDIA_TOOLKIT_IMAGE}-packaging 120
+docker cp nvidia-container-toolkit-packaging:/artifacts/packages/ubuntu18.04/amd64 ${TEMP_DIR}
+docker rm -f nvidia-container-toolkit-packaging
```
elezar: What about something like: https://github.com/NVIDIA/nvidia-container-toolkit/blob/08b3a388e7b1d447e10d4c4d4a71dca29a98a964/hack/pull-packages.sh#L50-L58 instead? Then we don't have to worry about stopping the container again.

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+dpkg -i ${TEMP_DIR}/amd64/nvidia-container-toolkit-base_*_amd64.deb
+dpkg -i ${TEMP_DIR}/amd64/libnvidia-container1_*_amd64.deb
+dpkg -i ${TEMP_DIR}/amd64/libnvidia-container-tools_*_amd64.deb
```
elezar: I believe this can be a single line.
```suggestion
dpkg -i ${TEMP_DIR}/amd64/libnvidia-container1_*_amd64.deb \
    ${TEMP_DIR}/amd64/nvidia-container-toolkit-base_*_amd64.deb \
    ${TEMP_DIR}/amd64/libnvidia-container-tools_*_amd64.deb
```

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+while ! docker info > /dev/null 2>&1; do
+    echo "Waiting for dockerd to be ready..."
+    sleep 1
+done
+
+# install nvidia-container-toolkit from NVIDIA_TOOLKIT_IMAGE using docker
+TEMP_DIR="/tmp/nvidia-container-toolkit"
+mkdir -p "$TEMP_DIR"
+rm -f /usr/bin/nvidia-container-runtime-hook
+ln -s "$TEMP_DIR/toolkit/nvidia-container-runtime-hook" /usr/bin/nvidia-container-runtime-hook
```
elezar: Why is this required? We are installing all the required packages and as such we don't need to set up this symlink here.

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+docker cp nvidia-container-toolkit-packaging:/artifacts/packages/ubuntu18.04/amd64 ${TEMP_DIR}
+docker rm -f nvidia-container-toolkit-packaging
+
+dpkg -i ${TEMP_DIR}/amd64/nvidia-container-toolkit-base_*_amd64.deb
+dpkg -i ${TEMP_DIR}/amd64/libnvidia-container1_*_amd64.deb
+dpkg -i ${TEMP_DIR}/amd64/libnvidia-container-tools_*_amd64.deb
+
+nvidia-container-cli --version
+
+apt-get update -y && apt-get install -y curl gnupg2
```
elezar: Nit: We already install `curl` above. 

As a matter of interest, is there a reason that we don't separate the various parts of this script? Specifically setting up the Docker "node" and actually running the test?

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+  mount -t sysfs sys  sys
+  mount -t tmpfs tmp  tmp
+  mount -t tmpfs run  run
+
+  # 3 GPU setup via nvidia-container-cli
+  # Add potential install locations of nvidia-container-cli to PATH.
+  # /artifacts/{rpm,deb}/usr/bin are where the binary ends up after the
+  # packages are extracted in the application image.  /work is included for
+  # completeness since some images may copy the binary there.
+  export PATH="${PATH}:/artifacts/rpm/usr/bin:/artifacts/deb/usr/bin:/work"
```
elezar: Is this still the case? We now install the `deb` packages directly, meaning these should be installed to `/usr/bin` which SHOULD be in the path already.

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+  exec nvidia-smi -L
+IN_NS
+`
+
+	dockerRunCmdTemplate = `docker run --name %s --privileged --runtime=nvidia \
+    -e NVIDIA_VISIBLE_DEVICES=all \
+    -e NVIDIA_DRIVER_CAPABILITIES=all \
+	-e NVIDIA_TOOLKIT_IMAGE=%s \
+    -v %s:/libnvidia-container-cli.sh \
+    --entrypoint /libnvidia-container-cli.sh ubuntu`
```
elezar: For an `ubuntu` container `bash` is already the entrypoint, so we can replace this with:
```suggestion
    ubuntu /libnvidia-container-cli.sh`
```

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+  mkdir -p mnt
+  pivot_root . mnt
+  umount -l /mnt
+
+  exec nvidia-smi -L
+IN_NS
+`
+
+	dockerRunCmdTemplate = `docker run --name %s --privileged --runtime=nvidia \
+    -e NVIDIA_VISIBLE_DEVICES=all \
```
elezar: As a nit: it may be good to use something like:
```suggestion
    -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=all \
```
here instead to ensure that we're using CDI injection for the outer container.

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+    -e NVIDIA_DRIVER_CAPABILITIES=all \
+	-e NVIDIA_TOOLKIT_IMAGE=%s \
```
elezar: Why not use templates instead of format strings?

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+		// Write the script to the remote host and make it executable.
+		createScriptCmd := fmt.Sprintf(
+			"cat > %s <<'EOF'\n%s\nEOF\nchmod +x %s",
+			testScriptPath, libnvidiaContainerCliTestTemplate, testScriptPath,
+		)
+
+		_, _, err := runner.Run(createScriptCmd)
+		Expect(err).ToNot(HaveOccurred())
+
+		// If a container with the same name exists from a previous test run, remove it first.
```
elezar: Does it not make more sense to add this to a `BeforeEach` instead? Otherwise we have to remember to clean up in every test.

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+		// Build the docker run command (detached mode) from the template so it
+		// stays readable while still resulting in a single-line invocation.
+		dockerRunCmd := fmt.Sprintf(dockerRunCmdTemplate, containerName, imageName+":"+imageTag, testScriptPath)
+
+		// Launch the container in detached mode.
+		_, _, err = runner.Run(dockerRunCmd)
+		Expect(err).ToNot(HaveOccurred())
+
+		hostOutputNormalized := strings.TrimSpace(strings.ReplaceAll(hostOutput, "\r", ""))
```
elezar: Question: Is the `ReplaceAll` required? We're doing the same thing for the `Eventually` matcher below. Why can't we just set `hostOutput` to the trimmed output and compare it directly?

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,209 @@
...
+			if err != nil {
+				return fmt.Sprintf("error: %v", err)
+			}
+
+			// Docker already returns only the last line due to --tail 1, so just normalize.
+			return strings.TrimSpace(strings.ReplaceAll(logs, "\r", ""))
+		}, "5m", "10s").Should(Equal(hostOutputNormalized))
+
+		// Remove the script from the remote host.
+		runner.Run(fmt.Sprintf("rm -f %s", testScriptPath))
```
elezar: Should this not be an `After` or in a defer? If this test fails, we will NOT remove the script.

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: My bad on this section, previous to this iteration I was going for installing the toolkit from the container

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: Done

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: Done

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: done

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: moved to templating

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: done

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: done

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: done

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: done

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: done

tests/e2e/e2e_test.go
elezar: How do we handle the "test the locally installed toolkit" case?

tests/e2e/nvidia-container-cli_test.go
elezar: Question: Why do we include the `/artifacts` paths here?

ArangoGutierrez: Good catch, that's from the previous iteration when I was trying to run inside the toolkit image it self (Docker in docker) removed now

tests/e2e/nvidia-container-cli_test.go
elezar: > On distroless, it should work as we are setting the --entrypoint /libnvidia-container-cli.sh , so as long as the distroless supports /usr/bin/env bash scripts, this test should work

Distroless does not support bash. It includes `sh` from `busybox`.

tests/e2e/nvidia-container-cli_test.go
elezar: I think my confusion comes from considering the different "phases" of the container that we're using as a "node" as independent. We would want to set up a node in `BeforeAll` and then run a set of tests against it. Note that as discussed, we could actually run our full test set against a containerized node instead of against a remote docker deployment.

tests/e2e/nvidia-container-cli_test.go
elezar: As pointed out offline: Was there a specific reason that you didn't start a "node" running in a container and then run the test script against it explicitly? When looking at it, we could break the steps into:
1. Start a "node" with `docker run -d --name node --privileged --device=management.nvidia.com/gpu=all ubuntu sleep infinity` (or similar).
2. Run a script using docker exec against this node to install and start the Docker daemon.
3. Run a script using docker exec against this node to install the toolkit
4. Run a script using docker exec against this node to run the actual test.

This allows us enough flexibility to test various different aspects of our expected behaviour. For example, if we watnted to:
* test a different distribution we would update 1.
* test a different docker version we would update 2.
* test a different toolkit config we could update step 3.

tests/e2e/nvidia-container-cli_test.go
elezar: (in theory we could also generate a management spec on the node and then use NATIVE CDI support in docker).

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: Done

ArangoGutierrez:
@elezar PTAL RFR

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,242 @@
...
+	"fmt"
+	"strings"
+	"text/template"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	installDockerTemplate = `docker exec -u root {{.ContainerName}} bash -c '
```
elezar: Why not have the script be the "template" directly? i.e. not include the `docker exec` part? In most cases this would mean that the script is used as is and does not even require templating.

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+  # 4 Switch root into the prepared filesystem
+  pivot_root . mnt
+  umount -l mnt
+  nvidia-smi -L
+
+IN_NS
+'`
+
+	dockerRunCmdTemplate = `docker run -d --name {{.ContainerName}} --privileged --runtime=nvidia \
```
elezar: I also don't think this needs to be a template.

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: I tried that, but then the `fmt.SprintF("docker exec %s", template.String())` would run into issues, failing, complaining that line 1 didn't have EOL.

SO I decided to have self-contained templates that we can simply execute

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: Done

tests/e2e/nvidia-container-cli_test.go
ArangoGutierrez: Ok, I got what I was doing wrong, now is not a template

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"fmt"
+	"strings"
+	"text/template"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	installDockerTemplate = `
```
elezar, line 30: Nit: this is now a script and not a template.

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	AfterAll(func(ctx context.Context) {
+		// Cleanup: remove the container and the temporary script on the host.
+		// Use || true to ensure cleanup doesn't fail the test
+		runner.Run(fmt.Sprintf("docker rm -f %s 2>/dev/null || true", containerName)) //nolint:errcheck
+	})
+
+	It("should report the same GPUs inside the container as on the host", func(ctx context.Context) {
+		// Launch the container in detached mode.
+		_, _, err := runner.Run(dockerRunCmdTemplate)
```
elezar: I'm not really at the point where I want to block this, but what would be good would be if we could return a `Runner` when starting this container. We can then replace `runner.Run("docker exec ... ")` commands bellow with calls to this runner that JUST accept the script that we want to run. This could be done in a follow up though.

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,201 @@
...
+		// Build the docker run command (detached mode) from the template so it
+		// stays readable while still resulting in a single-line invocation.
+		tmpl, err := template.New("toolkitInstall").Parse(installCTKTemplate)
+		Expect(err).ToNot(HaveOccurred())
+
+		var toolkitInstall strings.Builder
+		err = tmpl.Execute(&toolkitInstall, struct {
+			ToolkitImage string
+		}{
+			ToolkitImage: imageName + ":" + imageTag,
```
elezar: As a note: This makes LOCAL tests difficult if one doesn't have the image built. Should we still make `imageName` and `imageTag` optional and skip this test if these are not set?

tests/e2e/nvidia-container-cli_test.go
```diff
@@ -0,0 +1,201 @@
...
+		})
+		Expect(err).ToNot(HaveOccurred())
+
+		_, _, err = runner.Run(fmt.Sprintf("docker exec -u root %s bash -c '%s'", containerName, toolkitInstall.String()))
+		Expect(err).ToNot(HaveOccurred())
+
+		// Run the test script in the container.
+		// Capture but don't fail on errors - we'll check the results via container logs.
```
elezar: What do you mean don't fail on errors? I think this comment might be out of date.

tests/e2e/Makefile
```diff
@@ -20,8 +20,14 @@ LOG_ARTIFACTS_DIR ?= $(CURDIR)/e2e_logs
 
 GINKGO_BIN := $(CURDIR)/bin/ginkgo
 
+# If GINKGO_FOCUS is not set, run all tests
+# current available tests:
+# - nvidia-container-cli
+# - docker
+GINKGO_FOCUS ?= nvidia-container-cli
```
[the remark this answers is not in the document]
elezar: This needs to be changed again.

tests/e2e/nvidia-container-cli_test.go
elezar: I have updated this PR with basic functionality to allow local tests. Here we mount the `nvidia-container-cli` and related libraries from the host instead of installing them from an image.

tests/e2e/nvidia-container-cli_test.go
elezar: I updated this.

tests/e2e/Makefile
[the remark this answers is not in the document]
elezar: I have removed this.

elezar (approved):

Issue #1124, opened 2025-06-02
vulkan failed in RHEL8(`vkCreateInstance failed with ERROR_INCOMPATIBLE_DRIVER`)


kimullaa:
When I tried vulkaninfo in RHEL8,  it failed.

```
$ docker run --runtime=nvidia --gpus all -it --rm  nvcr.io/nvidia/clara-holoscan/holoscan:v3.2.0-dgpu /bin/bash
root@a26235e4fd7f:/opt/nvidia/holoscan# apt update && apt install vulkan-tools

root@a26235e4fd7f:/opt/nvidia/holoscan# vulkaninfo
Cannot create Vulkan instance.
This problem is often caused by a faulty installation of the Vulkan driver or attempting to use a GPU that does not support Vulkan.
ERROR at ./vulkaninfo/vulkaninfo.h:649:vkCreateInstance failed with ERROR_INCOMPATIBLE_DRIVER
```

It seems that `/etc/vulkan/icd.d/` and libGLX.so is not mounted.

```
$ rpm -qa | grep nvidia-container-toolkit
nvidia-container-toolkit-1.17.6-1.x86_64
nvidia-container-toolkit-base-1.17.6-1.x86_64

$ rpm -qf /usr/share/vulkan/icd.d/nvidia_icd.x86_64.json
nvidia-driver-libs-575.51.03-1.el8.x86_64


$ docker run --runtime=nvidia --gpus all -it --rm  nvcr.io/nvidia/clara-holoscan/holoscan:v3.2.0-dgpu /bin/bash

=========================
== NVIDIA Holoscan SDK ==
=========================

NVIDIA Holoscan SDK Version: 3.2.0
Container image Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
Refer to /opt/nvidia/legal for inherited CUDA and TensorRT container licenses and copyrights.

This container includes the Holoscan libraries, GXF extensions, headers, example source
code, and sample datasets, as well as all Holoscan SDK dependencies.

Visit the User Guide to get started with the Holoscan SDK:
 https://docs.nvidia.com/holoscan/sdk-user-guide/getting_started.html

Python, C++, and GXF examples are installed in /opt/nvidia/holoscan/examples alongside their source
code, and run instructions:
 https://github.com/nvidia-holoscan/holoscan-sdk/tree/main/examples#readme.

See the HoloHub repository for a collection of Holoscan operators and applications:
[...]

Issue #811, opened 2024-11-25
nvidia-container-toolkit fails to create /etc/vulkan/icd.d/nvidia_icd.json


buzmeg:
When installing using toolbox on Fedora Silverblue 41, the nvidia-container-toolkit doesn't create the file /etc/vulkan/icd.d/nvidia_icd.json necessary for Vulkan to operate correctly--it doesn't find the nvidia card.  If you take a working nvidia_icd.json from elsewhere and copy it into place, the card is found.

Probably related to #767 

Installation transcript:
```
$ toolbox create -i registry.fedoraproject.org/fedora-toolbox:41 vk_new
Created container: vk_new
Enter with: toolbox enter vk_new
foo@fedora:~$ toolbox enter vk_new
⬢ [foo@toolbx ~]$ # Directions from https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html
⬢ [foo@toolbx ~]$ curl -s -L https://nvidia.github.io/libnvidia-container/stable/rpm/nvidia-container-toolkit.repo | \
  sudo tee /etc/yum.repos.d/nvidia-container-toolkit.repo
[nvidia-container-toolkit]
name=nvidia-container-toolkit
baseurl=https://nvidia.github.io/libnvidia-container/stable/rpm/$basearch
repo_gpgcheck=1
gpgcheck=0
enabled=1
gpgkey=https://nvidia.github.io/libnvidia-container/gpgkey
sslverify=1
sslcacert=/etc/pki/tls/certs/ca-bundle.crt

[nvidia-container-toolkit-experimental]
name=nvidia-container-toolkit-experimental
baseurl=https://nvidia.github.io/libnvidia-container/experimental/rpm/$basearch
repo_gpgcheck=1
gpgcheck=0
enabled=0
gpgkey=https://nvidia.github.io/libnvidia-container/gpgkey
sslverify=1
sslcacert=/etc/pki/tls/certs/ca-bundle.crt
⬢ [foo@toolbx ~]$ sudo dnf install nvidia-container-toolkit
Updating and loading repositories:
 nvidia-container-toolkit                                                                                                                                                      100% |  49.0 KiB/s |   2.9 KiB |  00m00s
>>> Librepo error: repomd.xml GPG signature verification error: Signing key not found
 Fedora 41 openh264 (From Cisco) - x86_64                                                                                                                                      100% |   8.7 KiB/s |   4.8 KiB |  00m01s
[...]

Issue #767, opened 2024-10-31
The NVIDIA ICD JSON occasionally goes missing from 'nvidia-ctk cdi generate'


debarshiray:
I have been playing with the NVIDIA Container Toolkit on Fedora 39 Workstation and the proprietary NVIDIA driver from RPM Fusion.  I have noticed that the NVIDIA installable client driver (or ICD) JSON for Vulkan occasionally goes missing from `nvidia-ctk cdi generate`:
```ShellSession
$ nvidia-ctk cdi generate --format yaml 2>/dev/null | grep vulkan
 - containerPath: /etc/vulkan/implicit_layer.d/nvidia_layers.json
   hostPath: /usr/share/vulkan/implicit_layer.d/nvidia_layers.json
```

... even though the file is present on the host operating system at `/usr/share/vulkan/icd.d/nvidia_icd.x86_64.json` and Vulkan support on the host is confirmed by:
```ShellSession
$ vulkaninfo --summary
...
...
Devices:
========
GPU0:
	apiVersion         = 1.3.280
	driverVersion      = 560.35.3.0
	vendorID           = 0x10de
	deviceID           = 0x1cbc
	deviceType         = PHYSICAL_DEVICE_TYPE_DISCRETE_GPU
	deviceName         = Quadro P600
	driverID           = DRIVER_ID_NVIDIA_PROPRIETARY
	driverName         = NVIDIA
	driverInfo         = 560.35.3.0
	conformanceVersion = 1.3.8.2
	deviceUUID         = 2efa4848-ba99-ccd3-0a19-f497b31331ca
	driverUUID         = c3ca0510-c7e6-5f1c-86a1-dc0ed4ea4e21
...
...
```

This means that Podman containers don't have Vulkan support through the proprietary NVIDIA driver, and can only use LLVMpipe.

Right now, I am observing this problem with:
```
$ uname --kernel-release
6.11.4-101.fc39.x86_64
$ rpm -q kernel
kernel-6.5.6-300.fc39.x86_64
kernel-6.11.4-101.fc39.x86_64
$ rpm -q kmod-nvidia
kmod-nvidia-560.35.03-1.fc39.x86_64
```

debarshiray:
I forgot to mention the NVIDIA Container Toolkit version:
```ShellSession
$ nvidia-ctk --version
NVIDIA Container Toolkit CLI version 1.16.1
$ rpm -qf $(which nvidia-ctk)
[...]

PR #1125, opened 2025-06-02
Add discovery of arch-specific vulkan ICD

elezar:
On some RPM-based platforms, the path of the Vulkan ICD file include an architecture-specific infix to distinguish it from other architectures. (Most notably x86_64 vs i686). This change attempts to discover the arch-specific ICD file in addition to the standard nvidia_icd.json.

See:
* https://github.com/NVIDIA/nvidia-container-toolkit/issues/767
* #1124
* https://github.com/NVIDIA/nvidia-container-toolkit/issues/811
* https://github.com/nvidia-holoscan/holohub/pull/884


internal/discover/graphics.go
```diff
@@ -81,11 +82,21 @@ func NewGraphicsMountsDiscoverer(logger logger.Interface, driver *root.Driver, h
...
 func newVulkanConfigsDiscover(logger logger.Interface, driver *root.Driver) Discover {
 	locator := lookup.First(driver.Configs(), driver.Files())
+
+	var archSpecificICD string
+	switch runtime.GOARCH {
+	case "amd64":
+		archSpecificICD = "vulkan/icd.d/nvidia_icd.x86_64.json"
+	case "arm64":
+		// TODO: Validate whether the aarch64 infix is correct.
+		archSpecificICD = "vulkan/icd.d/nvidia_icd.aarch64.json"
```
debarshiray: I think `aarch64` is correct.  The sources of the RPM Fusion build for the proprietary NVIDIA driver uses `%{_target_cpu}`, and I expect that to evaluate to `aarch64`.

See: https://bugzilla.rpmfusion.org/show_bug.cgi?id=7167

elezar: Thanks. Let's keep it as is and we can follow-up again before the final v1.18.0 release if we have a chance to test on an arm64-based system.

PR #1128, opened 2025-06-03
Sort links to get consistent and repeateable output

agustinhenze:
When you run sudo nvidia-ctk cdi generate --output=/etc/cdi/nvidia.yaml multiple times, you get small differences between each run, but only because the output of links is not sorted. Sorting the links solve the problem of reproducibility among runs.


elezar:
@agustinhenze since the symlinks rely on the ordering of the mounts and we were using a map to deduplicate mounts, this could have lead to the non-repeatable output. We did address this in #1102 and #1104 (which was included in the v1.17.8 release). Although I'm not against sorting the links, I would just like to be certain that we didn't miss some edge case in those changes.

agustinhenze:
> @agustinhenze since the symlinks rely on the ordering of the mounts and we were using a map to deduplicate mounts, this could have lead to the non-repeatable output. We did address this in #1102 and #1104 (which was included in the v1.17.8 release). Although I'm not against sorting the links, I would just like to be certain that we didn't miss some edge case in those changes.

mmm that's right, I thought I tried before the output was inconsistent in main. But it seems fixed now and I am using an old version. Thanks for reviewing!

PR #1130, opened 2025-06-05
BUGFIX: modifier: respect GPU volume-mount device requests

ArangoGutierrez:
The gated modifiers used to add support for GDS, Mofed, and CUDA Forward Comatibility only check the NVIDIA_VISIBLE_DEVICES envvar to determine whether GPUs are requested and modifications should be made. This means that use cases where volume mounts are used to request devices (e.g. when using the GPU Device Plugin) are not supported.

These changes make the extraction of devices for gated and graphics modifiers consistent with other methods.

Fixes: #1049

### DIFF
--- a/internal/config/image/cuda_image.go
+++ b/internal/config/image/cuda_image.go
@@ -144,8 +144,8 @@ func (i CUDA) HasDisableRequire() bool {
 	return false
 }
 
-// DevicesFromEnvvars returns the devices requested by the image through environment variables
-func (i CUDA) DevicesFromEnvvars(envVars ...string) VisibleDevices {
+// devicesFromEnvvars returns the devices requested by the image through environment variables
+func (i CUDA) devicesFromEnvvars(envVars ...string) []string {
 	// We concantenate all the devices from the specified env.
 	var isSet bool
 	var devices []string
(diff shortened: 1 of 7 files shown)


internal/modifier/cdi.go
```diff
@@ -95,7 +95,7 @@ func getDevicesFromSpec(logger logger.Interface, ociSpec oci.Spec, cfg *config.C
 
 	var devices []string
 	seen := make(map[string]bool)
-	for _, name := range container.VisibleDevicesFromEnvVar() {
+	for _, name := range container.VisibleDevices() {
```
elezar: I may have mentioned this in #1110, but we already process the mount devices from above, and although this MAY have the desired effect we may want to revisit how we handle CDI device requests and how we distinguish between these are "regular" requests. 

My suggestion would be to:
1. Keep `VisibleDevicesFromEnvVar` public for now.
2. Update the other usages as per this PR.
3. Handle CDI devices in a follow up.

elezar:
Please update the commit messages / PR heading and description to better indicate intent. We are making this change to FIX A BUG. That is not clear from the commit messages and as such this will be missed in our release notes.

elezar:
Please also add unit tests to capture the new behaviour in the affected code. (ideally first failing with the existing implementation).

This PR is not a blocker for the v1.18.0-rc.1 release.

PR #1131, opened 2025-06-05
Bump release v1.18.0 rc.1

elezar:


CHANGELOG.md
```diff
@@ -1,34 +1,11 @@
 # NVIDIA Container Toolkit Changelog
 
-## v1.17.4
```
ArangoGutierrez, line 3: why are we deleting the point releases from prev release?  this was not the case from 1.16 to 1.17.?

elezar: Because managing changelogs is hard and forward-porting changlog entries to `main` makes this even harder. My thinking was that: `main` will contain the changelog entries for the initial release of any point version (since we branch after that release). Subsequent changes for the next MINOR release will be in the changelog in `main`, but no point release changelogs will be present.

Maybe put another way ... the changelog describes what has changed between `v1.17.0` and `v1.18.0`. There is not clean way to show what has changed between `v1.17.x` (`x > 0`) and `v1.18.0`.

elezar: cc @jgehrcke

ArangoGutierrez: ok, that is something I can get on board with. Thanks for the context. And I agree, changelog should be from 1.x to 1.y, if someone wants fine-grain details, they can visit the release list on GitHub to see the specific changes from 1. x.y to 1.x.z

ArangoGutierrez (approved):
After the context provided on the comments I am ok with this PR

PR #1132, opened 2025-06-05
Make CDI device requests consistent with other methods

elezar:
Following the refactoring of device request extraction, we can now make CDI device requests consistent with other methods.

This change moves to using image.VisibleDevices instead of separate calls to CDIDevicesFromMounts and VisibleDevicesFromEnvVar. This also changes the way in which annotation requests are handled to be consistent with other mechanisms by including annotation devices in the list of VisibleDevices.

See:
* #1110 for the device request refactoring
* #1130 for making other requests consistent

### DIFF
--- a/internal/config/image/cuda_image.go
+++ b/internal/config/image/cuda_image.go
@@ -238,6 +238,12 @@ func (i CUDA) OnlyFullyQualifiedCDIDevices() bool {
 // In cases where environment variable requests required privileged containers,
 // such devices requests are ignored.
 func (i CUDA) VisibleDevices() []string {
+	// If annotation device requests are present, these are preferred.
+	annotationDeviceRequests := i.cdiDeviceRequestsFromAnnotations()
+	if len(annotationDeviceRequests) > 0 {
+		return annotationDeviceRequests
+	}
+
 	// If enabled, try and get the device list from volume mounts first
 	if i.acceptDeviceListAsVolumeMounts {
 		volumeMountDeviceRequests := i.visibleDevicesFromMounts()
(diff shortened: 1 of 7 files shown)


internal/config/image/cuda_image.go
```diff
@@ -276,20 +273,27 @@ func (i CUDA) VisibleDevicesFromEnvVar() []string {
...
 		}
-		devices = append(devices, device)
+
 	}
 	return devices
 }
 
-// DevicesFromMounts returns a list of device specified as mounts.
-func (i CUDA) DevicesFromMounts() []string {
+// requestsFromMounts returns a list of device specified as mounts.
```
elezar, line 333: I renamed this function since we're returning all requests including:
* device requests (legacy and CDI)
* imex channel requests

internal/modifier/cdi.go
```diff
@@ -66,55 +66,72 @@ func NewCDIModifier(logger logger.Interface, cfg *config.Config, ociSpec oci.Spe
...
 		return nil, nil
 	}
 
-	if cfg.AcceptEnvvarUnprivileged || image.IsPrivileged((*image.OCISpec)(rawSpec)) {
+	if c.acceptEnvvarUnprivileged || image.IsPrivileged((*image.OCISpec)(rawSpec)) {
 		return devices, nil
 	}
 
-	logger.Warningf("Ignoring devices specified in NVIDIA_VISIBLE_DEVICES: %v", devices)
+	c.logger.Warningf("Ignoring devices specified in NVIDIA_VISIBLE_DEVICES: %v", devices)
```
ArangoGutierrez: My question here is, `container.VisibleDevices()` does these 2 things as well, the privileged check and the logging. so I think is safe to assume that the devices returned if not nil are safe to return. Also a log with a similar string will be reported from `container.VisibleDevices()`

elezar: You're right. This is not required. I thought I changed it. Will update.

internal/config/image/cuda_image.go
ArangoGutierrez: Note: DevicesFromMounts has been renamed and turned private to the config pkg

elezar: Yes. It was not used outside this package.

internal/config/image/cuda_image.go
```diff
@@ -321,23 +363,30 @@ func (i CUDA) DevicesFromMounts() []string {
...
+		return "", fmt.Errorf("invalid mount CDI device request: %s", m)
 	}
-	return devices
+
+	requestedDevice := strings.TrimPrefix(string(m), volumeMountDevicePrefixCDI)
+	if parser.IsQualifiedName(requestedDevice) {
+		return requestedDevice, nil
+	}
+
+	parts := strings.SplitN(requestedDevice, "/", 3)
```
jgehrcke, line 385: is this trying to parse vendor, class, device from
```
/cdi/<vendor>/<class>/<device>
```
?

jgehrcke: ah, `	volumeMountDevicePrefixCDI  = "cdi/"`.

and the argument 3 probably means "at most two splits", i.e. max three tokens.

elezar: Yes, a mounted CDI device will have the path `/var/run/nvidia-container-devices/cdi/vendor.com/class/device` or `/var/run/nvidia-container-devices/cdi/vendor.com/class=device`. This code handles the former case where we split `vendor.com/class/device` into AT MOST three parts.

From the `SplitN` docs:
> // SplitN slices s into substrings separated by sep and returns a slice of
> // the substrings between those separators.
> //
> // The count determines the number of substrings to return:
> //   - n > 0: at most n substrings; the last substring will be the unsplit remainder;
> //   - n == 0: the result is nil (zero substrings);
> //   - n < 0: all substrings.

PR #1150, opened 2025-06-16
Fix discovery of nvidia-fs devices in non-privileged containers

elezar:
The /dev/nvidia-fs* device nodes for GDS are not greated at the driver root when running a containerized driver and are always created in /.

This change updates the search path for these device nodes so that non-privilged containers also have the device nodes injected.


elezar:
Closing until we figure out what is required.

PR #1152, opened 2025-06-17
Update image tags for single image

elezar:
These changes follow on from #602 to make the images that are published more consistent.

It also make some required changes to the internal CI.


.gitlab-ci.yml
cdesiniotis, line 1: Can we remove this file completely?

.nvidia-ci.yml
```diff
@@ -34,24 +34,62 @@ variables:
   SKIP_QEMU_SETUP: "1"
   # Define the public staging registry
   STAGING_REGISTRY: ghcr.io/nvidia
-  STAGING_VERSION: ${CI_COMMIT_SHORT_SHA}
+  STAGING_VERSION:  ${CI_COMMIT_SHORT_SHA}
```
cdesiniotis: nit: was this change intended?

tariq1890: why add an extra space here?

.nvidia-ci.yml
```diff
@@ -34,24 +34,62 @@ variables:
...
+    OUT_REGISTRY: "${CI_REGISTRY}"
     OUT_REGISTRY_USER: "${CI_REGISTRY_USER}"
     OUT_REGISTRY_TOKEN: "${CI_REGISTRY_PASSWORD}"
-    OUT_REGISTRY: "${CI_REGISTRY}"
     OUT_IMAGE_NAME: "${CI_REGISTRY_IMAGE}/container-toolkit"
-    PUSH_MULTIPLE_TAGS: "false"
+    OUT_IMAGE_TAG: "${CI_COMMIT_SHORT_SHA}"
   # We delay the job start to allow the public pipeline to generate the required images.
   rules:
+    - if: $CI_PIPELINE_SOURCE == "web"
```
cdesiniotis: Just curious, why is this required?

tariq1890 (approved):

.nvidia-ci.yml
```diff
@@ -199,34 +167,54 @@ scan-packaging:
...
-release:staging-ubuntu20.04:
-  extends:
-    - .release:staging
-    - .dist-ubuntu20.04
-  needs:
-    - image-ubuntu20.04
+    OUT_REGISTRY: "${NGC_REGISTRY}"
+    OUT_REGISTRY_USER: "${NGC_REGISTRY_USER}"
+    OUT_REGISTRY_TOKEN: "${NGC_REGISTRY_TOKEN}"
+    OUT_IMAGE_NAME: "${NGC_REGISTRY_STAGING_IMAGE_NAME}"
```
cdesiniotis: We have recently created a `NGC_STAGING_REGISTRY` CI/CD variable at the group level. Let's use that instead of a per-project variable.
```suggestion
    OUT_IMAGE_NAME: "${NGC_STAGING_REGISTRY}"/container-toolkit
```

.nvidia-ci.yml
elezar: Not on purpose. Will update.

.gitlab-ci.yml
elezar: Yes, we can.

.nvidia-ci.yml
elezar: Sounds good. Thanks!

.nvidia-ci.yml
elezar: Nope. Updated.

.nvidia-ci.yml
elezar: This means that when this pipeline is triggered from the web (through "Run Pipeline") there is no delay, otherwise is triggered with the delay below. (that is at least the idea).

elezar: I have added comments and updated the conditions.

.gitlab-ci.yml
elezar: I was wondering whether we should move to a single `.gitlab-ci.yml` here. This may simplify integration with Nspect for example where they require this for GitLab CI.

PR #1154, opened 2025-06-18
Switch to distroless Base image

elezar:
This change switches to the nvcr.io/nvidia/distroless/go:v3.1.9-dev distroless go image for both the application image and the packaging image.

### DIFF
--- a/deployments/container/Dockerfile
+++ b/deployments/container/Dockerfile
@@ -164,3 +164,8 @@ LABEL description="See summary"
 COPY LICENSE /licenses/
 
 ENTRYPOINT ["/work/nvidia-ctk-installer"]
+
+# The GPU Operator exec's nvidia-toolkit in its entrypoint.
+# We create a symlink here to ensure compatibility with older
+# GPU Operator versions.
+RUN ln -s /work/nvidia-ctk-installer /work/nvidia-toolkit


deployments/container/Dockerfile
```diff
@@ -116,13 +129,15 @@ RUN set -eux; \
...
+FROM scratch AS artifacts
 
 COPY --from=rpmpackages /artifacts/rpm /artifacts/rpm
 COPY --from=debpackages /artifacts/deb /artifacts/deb
 COPY --from=build /artifacts/bin /artifacts/build
 
-FROM nvcr.io/nvidia/cuda:12.9.0-base-ubi9
+# The application stage contains the application used as a GPU Operator
+# operand.
+FROM nvcr.io/nvidia/distroless/go:v3.1.9-dev AS application
```
cdesiniotis, line 138: Note, the shell included in the `-dev` tags is located at `/busybox/sh`. I would recommend creating a symlink at `/bin/sh` so that in the operator we can use `#! /bin/sh` as the shebang for the entrypoint script. By using `/bin/sh` we remain backwards compatible with older toolkit images that are not built on distroless. We have tested this with other operands, e.g. https://github.com/NVIDIA/k8s-kata-manager/blob/f58e4dad0695043a545b17e3e159e24828816a62/deployments/container/Dockerfile#L50-L51
```suggestion
FROM nvcr.io/nvidia/distroless/go:v3.1.9-dev AS application

SHELL ["/busybox/sh", "-c"]
RUN ln -s /busybox/sh /bin/sh
```

cdesiniotis: Also, should we explicitly set `USER 0:0` in the Dockerfile as the default user in distroless is uid 1000? I assume the toolkit requires running as root (for restarting containerd).

elezar: Thanks for the shell tip. Will update.

I'm not sure on the user preference. Does the GPU Operator not set the user in general? Would using the current user (`USER 1000:1000`) not be more "compliant"?

elezar: I've updated the user to `0:0` below.

cdesiniotis: >I'm not sure on the user preference. Does the GPU Operator not set the user in general? Would using the current user (USER 1000:1000) not be more "compliant"?

The GPU Operator does not explicitly set the `runAsUser` / `runAsGroup` fields when deploying Daemonsets, so we currently depend on the user / group defined in the image itself.

cdesiniotis: I have performed a quick sanity check of the toolkit image built in this PR. Looks good.

Besides having to change the entrypoint script to be a POSIX shell script (and not bash), I also had to change https://github.com/NVIDIA/gpu-operator/blob/6324d2aca562edf46d93cbf9d2a0837ab5c12e59/assets/state-container-toolkit/0400_configmap.yaml#L34 from
```
exec nvidia-toolkit
```
to
```
exec nvidia-ctk-installer
```

I see the name of the executable has changed. This is a breaking change that will need to be made when we bump the version of the toolkit to 1.18.0 in the GPU Operator.

cdesiniotis: I have raised https://github.com/NVIDIA/gpu-operator/pull/1496 which updates our entrypoint scripts in the operator to use `sh` instead of `bash`.

elezar: I think we can include an `nvidia-toolkit` symlink so that we maintain backward compatibility.

Also on:
> I've updated the user to 0:0 below.

I had to set the user before we create the `/bin/sh` symlink since the default user doesn't have permissions to write to `/bin`.

elezar: I have updated the image to include a `/work/nvidia-toolkit` -> `/work/nvidia-ctk-installer` symlink. This should allow compatibility with the GPU Operator.

cdesiniotis: Thanks!

PR #1164, opened 2025-06-26
fix(ldconfig): suppress /proc mount error for non-root users

### LINKED ISSUE #1163: nvidia-cdi-hook>=1.17.7: Failed to mount /proc in the rootless mode
In the rootless (podman) mode, the nvidia-cdi-hook update-ldconfig failed to execute: 1.

The program stucks in `mountProc`, in `internal/ldconfig/ldconfig_linux.go`

```go
// internal/ldconfig/ldconfig_linux.go

// mountProc mounts a clean proc filesystem in the new root.
func mountProc(newroot string) error {
        target := filepath.Join(newroot, "/proc")

        if err := os.MkdirAll(target, 0755); err != nil {
                return fmt.Errorf("error creating directory: %w", err)
        }
        return unix.Mount("proc", target, "proc", 0, "")
}
```

This changes has been created by https://github.com/NVIDIA/nvidia-container-toolkit/commit/ac9146832ba8e1ca11ed3661ff6c0ac48b3f4e9f (>=1.17.7)

## Solution

```go
// internal/ldconfig/ldconfig.go

func (l *Ldconfig) prepareRoot() (string, error) {
        // To prevent leaking the parent proc filesystem, we create a new proc mount
        // in the specified root.
        if err := mountProc(l.inRoot); err != nil {
                // BEGIN PATCH
                // return "", fmt.Errorf("error mounting /proc: %w", err)
                // Non-root users cannot mount /proc
                fmt.Fprintf(os.Stderr, "error mounting /proc: %w\n", err)
                // END PATCH
        }

        ...
}
```

HoKim98:
Non-root users are not permitted to mount /proc, which causes `mountProc` to fail and abort `prepareRoot`. To prevent this, the error is now suppressed with `rootless=true` argument at 3rd in `soname-symlinks` and `update-ldcache` instead of causing a failure, allowing continued execution.
## Fixes

Fixes #1163

### DIFF
--- a/internal/ldconfig/ldconfig.go
+++ b/internal/ldconfig/ldconfig.go
@@ -19,9 +19,11 @@ package ldconfig
 
 import (
 	"fmt"
+	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/config"
@@ -39,6 +41,7 @@ const (
 type Ldconfig struct {
 	ldconfigPath string
 	inRoot       string
+	isRootless   bool
 }
 
 // NewRunner creates an exec.Cmd that can be used to run ldconfig.
@@ -47,26 +50,59 @@ func NewRunner(id string, ldconfigPath string, containerRoot string, additionala
 		id,
 		strings.TrimPrefix(config.NormalizeLDConfigPath("@"+ldconfigPath), "@"),
 		containerRoot,
+		fmt.Sprintf("rootless=%v", os.Geteuid() != 0),
 	}
 	args = append(args, additionalargs...)
 
 	return createReexecCommand(args)
 }
 
-// New creates an Ldconfig struct that is used to perform operations on the
-// ldcache and libraries in a particular root (e.g. a container).
-func New(ldconfigPath string, inRoot string) (*Ldconfig, error) {
+// NewFromRunnerArgs creates an Ldconfig struct from the args passed to the Cmd
+// above.
+// This struct is used to perform operations on the ldcache and libraries in a
+// particular root (e.g. a container).
+//
+// args[0] is the reexec initializer function name
+// args[1] is the path of the ldconfig binary on the host
+// args[2] is the container root directory
+// args[3] is the optional flag indicating whether the container is being run rootless.
+// The remaining args are folders where soname symlinks need to be created.
+func NewFromArgs(args ...string) (*Ldconfig, []string, error) {
+	// Validate the number of arguments: [program, ldconfigPath, inRoot]
+	if len(args) < 4 {
+		return nil, args, fmt.Errorf("incorrect arguments: %v", args)
+	}
+
+	ldconfigPath := args[1]
+	if ldconfigPath == "" {
+		return nil, args, fmt.Errorf("an ldconfig path must be specified")
+	}
+
+	inRoot := args[2]
+	if inRoot == "" || inRoot == "/" {
+		return nil, args, fmt.Errorf("ldconfig must be run in the non-system root")
+	}
+
+	isRootless := false
+	if len(args) >= 4 && strings.HasPrefix(args[3], "rootless=") {
+		value, err := strconv.ParseBool(strings.TrimPrefix(args[3], "rootless="))
+		if err != nil {
+			return nil, args, err
+		}
+		isRootless = value
+	}
+
+	remainingArgs := args[3:]
+	if isRootless {
+		remainingArgs = args[4:]
+	}
+
 	l := &Ldconfig{
 		ldconfigPath: ldconfigPath,
 		inRoot:       inRoot,
+		isRootless:   isRootless,
 	}
-	if ldconfigPath == "" {
-		return nil, fmt.Errorf("an ldconfig path must be specified")
-	}
-	if inRoot == "" || inRoot == "/" {
-		return nil, fmt.Errorf("ldconfig must be run in the non-system root")
-	}
-	return l, nil
+	return l, remainingArgs, nil
 }
 
 // CreateSonameSymlinks uses ldconfig to create the soname symlinks in the
@@ -130,7 +166,10 @@ func (l *Ldconfig) prepareRoot() (string, error) {
 	// To prevent leaking the parent proc filesystem, we create a new proc mount
 	// in the specified root.
 	if err := mountProc(l.inRoot); err != nil {
-		return "", fmt.Errorf("error mounting /proc: %w", err)
+		if !l.isRootless {
+			return "", fmt.Errorf("error mounting /proc: %w", err)
+		}
+		log.Printf("Ignoring error for rootless container: %v", err)
 	}
 
 	// We mount the host ldconfig before we pivot root since host paths are not
--- a/cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
+++ b/cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
@@ -142,25 +142,11 @@ func createSonameSymlinksHandler() {
 // It is invoked from a reexec'd handler and provides namespace isolation for
 // the operations performed by this hook. At the point where this is invoked,
 // we are in a new mount namespace that is cloned from the parent.
-//
-// args[0] is the reexec initializer function name
-// args[1] is the path of the ldconfig binary on the host
-// args[2] is the container root directory
-// The remaining args are directories where soname symlinks need to be created.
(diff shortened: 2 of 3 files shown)


internal/ldconfig/ldconfig.go
```diff
@@ -130,7 +130,8 @@ func (l *Ldconfig) prepareRoot() (string, error) {
 	// To prevent leaking the parent proc filesystem, we create a new proc mount
 	// in the specified root.
 	if err := mountProc(l.inRoot); err != nil {
-		return "", fmt.Errorf("error mounting /proc: %w", err)
+		// Non-root users cannot mount /proc; print the error but continue execution
```
elezar: Is there a way to detect whether this is running as a non-root user instead?

HoKim98: I tried some ways.

At first, I thought getting uid and gid would help. But it outcomes "0" in the containers, caused by sandboxed namespaces.

AFAIK, the only way to detect whether this is running as a non-root user is to trying it.

elezar: Looking at `runc` the following check is [used to detect rooless](https://github.com/opencontainers/runc/blob/b1722d790214952e8a20d4ddd6a83451b9b665a1/utils_linux.go#L192):
```go
		RootlessEUID:     os.Geteuid() != 0,
```
In the hooks where this is used, we're already in a new userns here, so as you pointed out the check may not work here. What about checking this where the hook is actually invoked and then passing that through to the runner?

```
diff --git a/cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go b/cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
index 7f1da580..10d52c43 100644
--- a/cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
+++ b/cmd/nvidia-cdi-hook/create-soname-symlinks/soname-symlinks.go
@@ -142,22 +142,8 @@ func createSonameSymlinksHandler() {
 // It is invoked from a reexec'd handler and provides namespace isolation for
 // the operations performed by this hook. At the point where this is invoked,
 // we are in a new mount namespace that is cloned from the parent.
-//
-// args[0] is the reexec initializer function name
-// args[1] is the path of the ldconfig binary on the host
-// args[2] is the container root directory
-// The remaining args are directories where soname symlinks need to be created.
 func createSonameSymlinks(args []string) error {
-	if len(args) < 3 {
-		return fmt.Errorf("incorrect arguments: %v", args)
-	}
-	hostLdconfigPath := args[1]
-	containerRootDirPath := args[2]
-
-	ldconfig, err := ldconfig.New(
-		hostLdconfigPath,
-		containerRootDirPath,
-	)
+	ldconfig, err := ldconfig.NewFromArgs(args...)
 	if err != nil {
 		return fmt.Errorf("failed to construct ldconfig runner: %w", err)
 	}
diff --git a/cmd/nvidia-cdi-hook/update-ldcache/update-ldcache.go b/cmd/nvidia-cdi-hook/update-ldcache/update-ldcache.go
index ef614709..78219022 100644
--- a/cmd/nvidia-cdi-hook/update-ldcache/update-ldcache.go
+++ b/cmd/nvidia-cdi-hook/update-ldcache/update-ldcache.go
@@ -140,22 +140,8 @@ func updateLdCacheHandler() {
 // It is invoked from a reexec'd handler and provides namespace isolation for
 // the operations performed by this hook. At the point where this is invoked,
 // we are in a new mount namespace that is cloned from the parent.
-//
-// args[0] is the reexec initializer function name
-// args[1] is the path of the ldconfig binary on the host
-// args[2] is the container root directory
-// The remaining args are folders where soname symlinks need to be created.
 func updateLdCache(args []string) error {
-	if len(args) < 3 {
-		return fmt.Errorf("incorrect arguments: %v", args)
-	}
-	hostLdconfigPath := args[1]
-	containerRootDirPath := args[2]
-
-	ldconfig, err := ldconfig.New(
-		hostLdconfigPath,
-		containerRootDirPath,
-	)
+	ldconfig, err := ldconfig.NewFromArgs(args...)
 	if err != nil {
 		return fmt.Errorf("failed to construct ldconfig runner: %w", err)
 	}
diff --git a/internal/ldconfig/ldconfig.go b/internal/ldconfig/ldconfig.go
index f3db1a77..86324a82 100644
--- a/internal/ldconfig/ldconfig.go
+++ b/internal/ldconfig/ldconfig.go
@@ -19,9 +19,11 @@ package ldconfig
 
 import (
 	"fmt"
+	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/config"
@@ -39,6 +41,7 @@ const (
 type Ldconfig struct {
 	ldconfigPath string
 	inRoot       string
+	isRootless   bool
 }
 
 // NewRunner creates an exec.Cmd that can be used to run ldconfig.
@@ -47,25 +50,47 @@ func NewRunner(id string, ldconfigPath string, containerRoot string, additionala
 		id,
 		strings.TrimPrefix(config.NormalizeLDConfigPath("@"+ldconfigPath), "@"),
 		containerRoot,
+		fmt.Sprintf("rootless=%v", os.Geteuid() != 0),
 	}
 	args = append(args, additionalargs...)
 
 	return createReexecCommand(args)
 }
 
-// New creates an Ldconfig struct that is used to perform operations on the
-// ldcache and libraries in a particular root (e.g. a container).
-func New(ldconfigPath string, inRoot string) (*Ldconfig, error) {
-	l := &Ldconfig{
-		ldconfigPath: ldconfigPath,
-		inRoot:       inRoot,
-	}
+// NewFromRunnerArgs creates an Ldconfig struct from the args passed to the Cmd
+// above.
+// This struct is used to perform operations on the ldcache and libraries in a
+// particular root (e.g. a container).
+//
+// args[1] is the path of the ldconfig binary on the host
+// args[2] is the container root directory
+// args[3] is the flag indicating whether the container is being run rootless.
+func NewFromArgs(args ...string) (*Ldconfig, error) {
+	if len(args) < 4 {
+		return nil, fmt.Errorf("incorrect arguments: %v", args)
+	}
+	ldconfigPath := args[1]
 	if ldconfigPath == "" {
 		return nil, fmt.Errorf("an ldconfig path must be specified")
 	}
+
+	inRoot := args[2]
 	if inRoot == "" || inRoot == "/" {
 		return nil, fmt.Errorf("ldconfig must be run in the non-system root")
 	}
+
+	isRootless, err := strconv.ParseBool(strings.TrimPrefix(args[3], "rootless="))
+	if err != nil {
+		return nil, err
+	}
+
+	l := &Ldconfig{
+		ldconfigPath: ldconfigPath,
+		inRoot:       inRoot,
+		isRootless:   isRootless,
+	}
 	return l, nil
 }
 
@@ -130,7 +155,10 @@ func (l *Ldconfig) prepareRoot() (string, error) {
 	// To prevent leaking the parent proc filesystem, we create a new proc mount
 	// in the specified root.
 	if err := mountProc(l.inRoot); err != nil {
-		return "", fmt.Errorf("error mounting /proc: %w", err)
+		if !l.isRootless {
+			return "", fmt.Errorf("error mounting /proc: %w", err)
+		}
+		log.Printf("Ignoring error for rootless container: %v", err)
 	}
 
 	// We mount the host ldconfig before we pivot root since host paths are not
```

HoKim98: I agree. I believe we can avoid side effects (e.g., unexpected behavior for root users) by adding a third argument `rootless=true` to both `soname-symlinks` and `update-ldcache`. I've rebased the code based on your suggestion, with the following additional changes:

1. Optional argument: In your suggested code, I made the `rootless=` argument optional to ensure it does not affect existing `nvidia-ctk` behavior for end users. This helps avoid any potential side effects.
1. Added `remainingArgs`: The new `NewFromArgs` function now returns the remaining arguments it did not consume. This allows `soname-symlinks` and `update-ldcache` to forward any remaining arguments directly to `ldconfig`.

elezar: Thanks for confirming that this works. 

Note that I have created https://github.com/elezar/nvidia-container-toolkit/tree/fix/suppress-proc-mount-error-for-non-root-users that takes advantage of a refactor from #1174 to improve how the arguments are handled.

One question that I also have is whether the pivot_root operation works as expected in a rootless container, or whether we would have to fallback to a "standard" chroot?

elezar: @HoKim98 sorry. Does 👎 mean:
* that you don't agree with the changes
* that it doesn't work
* that pivot_root is not supported?

PR #1166, opened 2025-06-27
Refactor cdi api

elezar:
The original CDI spec generation API was focussed on NVML device specifically. Since then we have replaced the more specific functions (for GPU and MIG devices) in the API with more generally applicable functions based on mode and device IDs.

This organic growth of APIs also means that for the NVML case specifically we had multiple different implementations of CDI spec generation making keeping things consistent more difficult.

Thes changes remove the redundant functions in the `nvcdi.Interface` allowing devices to be requested by ID across all use cases. It also refactors the CDI spec generation for NVML devices to ensure that the same generation logic is used for all cases.

### DIFF
--- a/pkg/nvcdi/lib.go
+++ b/pkg/nvcdi/lib.go
@@ -118,39 +118,39 @@ func New(opts ...Option) (Interface, error) {
 		)
 	}
 
-	var lib wrapped
+	var factory deviceSpecGeneratorFactory
 	switch l.resolveMode() {
 	case ModeCSV:
 		if len(l.csvFiles) == 0 {
 			l.csvFiles = csv.DefaultFileList()
 		}
-		lib = (*csvlib)(l)
+		factory = (*csvlib)(l)
 	case ModeManagement:
 		if l.vendor == "" {
 			l.vendor = "management.nvidia.com"
 		}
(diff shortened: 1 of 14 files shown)


ArangoGutierrez (dismissed):
LGTM

pkg/nvcdi/lib.go
```diff
@@ -162,7 +162,7 @@ func New(opts ...Option) (Interface, error) {
 	)
 
 	w := wrapper{
-		Interface:           lib,
+		factory:             factory,
```
ArangoGutierrez, line 165:
```suggestion
		factory:           factory,
```

ArangoGutierrez: This probably means the file needs to be gofmt'ed

ArangoGutierrez (changes requested):

pkg/nvcdi/lib.go
elezar: I double-checked locally and I think this is a github formatting issue.

ArangoGutierrez (approved):
LGTM
Thanks for this Evan, this is a great refactor of the API

ArangoGutierrez (approved):
LGTM

PR #1168, opened 2025-06-30
Fix leaking of tmpfs mount in CDI mode

ArangoGutierrez:
This patch fixes the handling of temporary files and directories in the NVIDIA container runtime hook, ensuring we don't leak tmpfs mounts when handling the params file in the container when in CDI mode. This is done by ensuring that the tmpfs mount that we create for the modified params file is created in the container's mount namespace instead of on the host.

We also add end2end tests to check for leaking of mounts.

PR #1169, opened 2025-06-30
[no-relnote] Add e2e test for firmware path traversal

ArangoGutierrez:
A container image could be crafted with a symbolic link in
/lib/firmware/nvidia/ that points to a location outside of the
container's root filesystem. When running such a container with the
NVIDIA Container Toolkit, this could potentially lead to files being
created on the host filesystem.

This change adds an end-to-end test to ensure that the toolkit is not
vulnerable to this kind of path traversal attack.

The test case covers two scenarios:
- When using CDI, which is the default, the test ensures that the
  container runs without error and no files are created on the host.
- When using the legacy nvidia-container-runtime-hook, the test
  ensures that the operation fails with a specific path error,
  preventing any file creation on the host.

### DIFF
--- a/tests/e2e/nvidia-container-toolkit_test.go
+++ b/tests/e2e/nvidia-container-toolkit_test.go
@@ -18,6 +18,7 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"path/filepath"
 	"strings"
 
@@ -295,6 +296,59 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
 		It("should not leak mounts when using CDI mode", func(ctx context.Context) {
 			_, _, err := runner.Run("docker run --rm -i --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all -e NVIDIA_DRIVER_CAPABILITIES=runtime.nvidia.com/gpu=all --mount type=bind,source=" + tmpDirPath + ",target=/empty,bind-propagation=shared ubuntu true")
 			Expect(err).ToNot(HaveOccurred())
+		})
+	})
+
+	When("Running a container where the firmware folder resolves outside the container root", Ordered, func() {
+		var outputDir string
+		BeforeAll(func(ctx context.Context) {
+			output, _, err := runner.Run("mktemp -d -p $(pwd)")
+			Expect(err).ToNot(HaveOccurred())
+			outputDir = strings.TrimSpace(output)
+
+			_, _, err = runner.Run("docker pull ubuntu")
+			Expect(err).ToNot(HaveOccurred())
+
+			_, _, err = runner.Run(`docker build -t firmware-test \
+            --build-arg RM_VERSION="$(basename $(ls -d /lib/firmware/nvidia/*.*))" \
+            --build-arg CURRENT_DIR="` + outputDir + `" \
+            - <<EOF
+FROM ubuntu
+RUN mkdir -p /lib/firmware/nvidia/
+ARG RM_VERSION
+ARG CURRENT_DIR
+RUN ln -s /../../../../../../../../\$CURRENT_DIR /lib/firmware/nvidia/\$RM_VERSION
+EOF`)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		AfterEach(func(ctx context.Context) {
+			output, _, err := runner.Run("ls -A " + outputDir)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(BeEmpty())
+		})
+
+		AfterAll(func(ctx context.Context) {
+			if outputDir != "" {
+				runner.Run(fmt.Sprintf("rm -rf %s", outputDir))
+			}
+		})
+
+		It("should not fail when using CDI", func(ctx context.Context) {
+			output, _, err := runner.Run("docker run --rm --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=all firmware-test")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(BeEmpty())
+		})
+
+		It("should not fail when using the nvidia-container-runtime", func(ctx context.Context) {
+			_, _, err := runner.Run("docker run --rm --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all -e NVIDIA_DRIVER_CAPABILITIES=all firmware-test")
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should fail when using the nvidia-container-runtime-hook", Label("legacy"), func(ctx context.Context) {
+			_, stderr, err := runner.Run("docker run --rm --runtime=runc --gpus=all firmware-test")
+			Expect(err).To(HaveOccurred())
+			Expect(stderr).To(ContainSubstring("nvidia-container-cli.real: mount error: path error:"))
 		})
 	})
 })
--- a/tests/e2e/runner.go
+++ b/tests/e2e/runner.go
@@ -124,7 +124,7 @@ func (r remoteRunner) Run(script string) (string, string, error) {
 	// Run the script
 	err = session.Run(script)
 	if err != nil {
-		return "", "", fmt.Errorf("script execution failed: %v\nSTDOUT: %s\nSTDERR: %s", err, stdout.String(), stderr.String())
+		return "", stderr.String(), fmt.Errorf("script execution failed: %v\nSTDOUT: %s\nSTDERR: %s", err, stdout.String(), stderr.String())
 	}
 
 	// Return stdout as string if no errors


tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -240,6 +240,22 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
 		BeforeAll(func(ctx context.Context) {
 			_, _, err := runner.Run("docker pull ubuntu")
 			Expect(err).ToNot(HaveOccurred())
+
+			err = buildDockerImage(
```
elezar: I think this test should be in its own section. We can then test the different behaviours based on runtime mode. For example if we run `docker run --runtime=runc --gpus all` we would expect to see an error due to the path traversal.

tests/e2e/runner.go
```diff
@@ -169,3 +169,15 @@ func connectOrDie(sshKey, sshUser, host, port string) (*ssh.Client, error) {
...
+	var buildArgs string
+	for k, v := range args {
+		buildArgs += fmt.Sprintf("--build-arg %s=%q ", k, v)
+	}
+
+	command := fmt.Sprintf("docker build -t %s %s - <<EOF\n%s\nEOF", tag, buildArgs, dockerfile)
+
+	_, _, err := runner.Run(command)
+	return err
+}
```
elezar: I think the raw script is easier to follow.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -256,5 +272,14 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
 
 			Expect(libs).To(ContainElements([]string{"libcuda.so", "libcuda.so.1"}))
 		})
+
+		It("should not attempt to create empty firmware files", func(ctx context.Context) {
+			_, _, err := runner.Run("docker run --rm --runtime=nvidia --gpus=all firmware-test")
+			Expect(err).To(HaveOccurred())
+
+			containerOutput, _, err := runner.Run("docker run --rm --runtime=runc --gpus=all firmware-test")
```
elezar: This is EXPECTED to fail.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -256,5 +272,14 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
 
 			Expect(libs).To(ContainElements([]string{"libcuda.so", "libcuda.so.1"}))
 		})
+
+		It("should not attempt to create empty firmware files", func(ctx context.Context) {
+			_, _, err := runner.Run("docker run --rm --runtime=nvidia --gpus=all firmware-test")
+			Expect(err).To(HaveOccurred())
```
elezar: With the v1.18.0 release this uses CDI by default which means that there will be NO error. This should be a different test case.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,4 +257,32 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+FROM ubuntu
+RUN mkdir -p /lib/firmware/nvidia/
+ARG RM_VERSION
+ARG CURRENT_DIR
+RUN ln -s /../../../../../../../../\$CURRENT_DIR /lib/firmware/nvidia/\$RM_VERSION
+EOF`)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should not create empty firmware files on the host when using CDI", func(ctx context.Context) {
```
elezar: This doesn't actually check that the files are not created on this host.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,4 +257,32 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+			output, _, err := runner.Run("docker run --rm --runtime=nvidia --gpus=all firmware-test")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(BeEmpty())
+		})
+
+		It("should fail when using the pre-start hook", func(ctx context.Context) {
```
elezar: De we refer to this as "legacy" mode in other tests?

ArangoGutierrez: on another open PR we do add a "legacy" label https://github.com/NVIDIA/nvidia-container-toolkit/pull/1168/files#diff-d1fe210afb2df1ba6dd0ebbd1129033e4416de4d8146f52122811c41f081ae75R280

tests/e2e/nvidia-container-toolkit_test.go
ArangoGutierrez: I have now edited the test description to be in line with the context

tests/e2e/nvidia-container-toolkit_test.go
ArangoGutierrez: label added

elezar: That's not what I asked. In the other test that we explicitly want to trigger the legacy code path we use:
```
		It("should work with nvidia-container-runtime-hook", func(ctx context.Context) {
```
Let's update this one to be:
```
		It("should fail when using the nvidia-container-runtime-hook", func(ctx context.Context) {
```

ArangoGutierrez: test description edited

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,4 +257,32 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+ARG CURRENT_DIR
+RUN ln -s /../../../../../../../../\$CURRENT_DIR /lib/firmware/nvidia/\$RM_VERSION
+EOF`)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should not fail when using CDI", func(ctx context.Context) {
+			output, _, err := runner.Run("docker run --rm --runtime=nvidia --gpus=all firmware-test")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(BeEmpty())
```
elezar, line 340: Let's update the tests to actually check that no files are created. Using an affected version of the toolkit we see:
```
$ docker run --rm -ti --gpus=all --runtime=nvidia firmware-test true
$ ls
gsp_ga10x.bin  gsp_tu10x.bin
```
So we would ensure that `ls gsp_*.bin` is empty.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,4 +257,32 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
 			Expect(libs).To(ContainElements([]string{"libcuda.so", "libcuda.so.1"}))
 		})
 	})
+
+	When("A container tries to create firmware files on the host", Ordered, func() {
+		BeforeAll(func(ctx context.Context) {
+			_, _, err := runner.Run("docker pull ubuntu")
+			Expect(err).ToNot(HaveOccurred())
+
+			_, _, err = runner.Run(`docker build -t firmware-test --build-arg RM_VERSION="$(basename $(ls -d /lib/firmware/nvidia/*.*))" --build-arg CURRENT_DIR="$(pwd)" - <<EOF
```
elezar: If we set `CURRENT_DIR` to a tmp folder we can clean it up between each of the test cases to ensure that we're properly testing that we're NOT creating empty files on the host.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,4 +257,32 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+		It("should not fail when using CDI", func(ctx context.Context) {
+			output, _, err := runner.Run("docker run --rm --runtime=nvidia --gpus=all firmware-test")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(BeEmpty())
+		})
+
+		It("should fail when using the nvidia-container-runtime-hook", Label("legacy"), func(ctx context.Context) {
+			_, stderr, err := runner.Run("docker run --rm --runtime=runc --gpus=all firmware-test")
+			Expect(err).To(HaveOccurred())
+			Expect(stderr).To(ContainSubstring("nvidia-container-cli.real: mount error: path error:"))
```
elezar, line 351: In addition to this we should also check that files are not created.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,4 +258,59 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+			_, _, err := runner.Run("docker pull ubuntu")
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		BeforeEach(
+			func(ctx context.Context) {
+				var err error
+				tmpDir, _, err = runner.Run("mktemp -d")
+				Expect(err).ToNot(HaveOccurred())
+				tmpDir = strings.TrimSpace(tmpDir)
```
elezar: Is there no per-test tmp dir that we can use?

ArangoGutierrez: This is what the `BeforeEach` is doing: it creates a new tmp dir per test (per `If` block)

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,4 +258,59 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+
+				dockerBuildCmd := fmt.Sprintf(`docker build -t firmware-test --build-arg RM_VERSION="$(basename $(ls -d /lib/firmware/nvidia/*.*))" --build-arg CURRENT_DIR=%q - <<EOF`, tmpDir)
```
elezar: Does `%q` use produce double quotes?

ArangoGutierrez: yup -> https://gobyexample.com/string-formatting

"To double-quote strings as in Go source, use %q."

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,4 +258,54 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+			_, _, err = runner.Run("docker pull ubuntu")
+			Expect(err).ToNot(HaveOccurred())
+
+			dockerBuildCmd := fmt.Sprintf(`docker build -t firmware-test --build-arg RM_VERSION="$(basename $(ls -d /lib/firmware/nvidia/*.*))" --build-arg CURRENT_DIR=%q - <<EOF`, outputDir)
+			dockerBuildDockerfile := `
+FROM ubuntu
+RUN mkdir -p /lib/firmware/nvidia/
+ARG RM_VERSION
+ARG CURRENT_DIR
+RUN ln -s /../../../../../../../../\$CURRENT_DIR /lib/firmware/nvidia/\$RM_VERSION
```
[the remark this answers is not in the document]
elezar: We are explicitly creating this symlink in the container to attempt to create an empty file on the host when the contianer starts.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,4 +258,54 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
 			Expect(libs).To(ContainElements([]string{"libcuda.so", "libcuda.so.1"}))
 		})
 	})
+
+	When("A container tries to create firmware files on the host", Ordered, func() {
```
elezar: Maybe:
```suggestion
	When("Running a container where the firmware folder resolves outside the container root", Ordered, func() {
```

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -257,4 +258,54 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+		var outputDir string
+		BeforeAll(func(ctx context.Context) {
+			pwd, _, err := runner.Run("pwd")
```
elezar: Why not use the `os` package?

elezar:
```go
			pwd, err := os.Getwd()

			workingDir := filepath.Join(pwd, "test-output")

			err = os.MkdirAll(workingDir, 0755)
```

ArangoGutierrez: Remember, we are running this over SSH, since the `runner` interface is designed to be agnostic, we can not use Go packages for system-level actions, we need to use bash via the runner

tests/e2e/nvidia-container-toolkit_test.go
ArangoGutierrez: Adopted

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -297,4 +298,54 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+
+	When("Running a container where the firmware folder resolves outside the container root", Ordered, func() {
+		var outputDir string
+		BeforeAll(func(ctx context.Context) {
+			pwd, _, err := runner.Run("pwd")
+			Expect(err).ToNot(HaveOccurred())
+			outputDir = filepath.Join(strings.TrimSpace(pwd), "test-output")
+
+			_, _, err = runner.Run(fmt.Sprintf("mkdir -p %s", outputDir))
+			Expect(err).ToNot(HaveOccurred())
```
elezar: Can we replace this with:

```suggestion
			output, _, err := runner.Run("mktemp -d -p $(pwd)")
			Expect(err).ToNot(HaveOccurred())
			outputDir = strings.TrimSpace(output)
```

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -297,4 +298,51 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+FROM ubuntu
+RUN mkdir -p /lib/firmware/nvidia/
+ARG RM_VERSION
+ARG CURRENT_DIR
+RUN ln -s /../../../../../../../../\$CURRENT_DIR /lib/firmware/nvidia/\$RM_VERSION
+EOF`
+			_, _, err = runner.Run(dockerBuildCmd + dockerBuildDockerfile)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
```
elezar: Is the following easier to read (instead of combining two strings as the command)?
```suggestion
            _, _, err = runner.Run(`docker build -t firmware-test \
            --build-arg RM_VERSION="$(basename $(ls -d /lib/firmware/nvidia/*.*))" \
            --build-arg CURRENT_DIR="` + outputDir + `" \
            - <<EOF
FROM ubuntu
RUN mkdir -p /lib/firmware/nvidia/
ARG RM_VERSION
ARG CURRENT_DIR
RUN ln -s /../../../../../../../../\$CURRENT_DIR /lib/firmware/nvidia/\$RM_VERSION
EOF`)
```

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -297,4 +298,51 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+			if outputDir != "" {
+				runner.Run(fmt.Sprintf("rm -rf %s", outputDir))
+			}
+		})
+
+		It("should not fail when using CDI", func(ctx context.Context) {
+			output, _, err := runner.Run("docker run --rm --runtime=nvidia --gpus=all firmware-test")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(BeEmpty())
+
```
elezar, line 342: Our other CDI tests we use:
```suggestion
			output, _, err := runner.Run("docker run --rm --runtime=nvidia --gpus=runtime.nvidia.com/gpu=all firmware-test")
```

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -297,4 +298,51 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+			output, _, _ = runner.Run(fmt.Sprintf("ls -A %s", outputDir))
+			Expect(output).To(BeEmpty())
+		})
+
+		It("should fail when using the nvidia-container-runtime-hook", Label("legacy"), func(ctx context.Context) {
+			_, stderr, err := runner.Run("docker run --rm --runtime=runc --gpus=all firmware-test")
+			Expect(err).To(HaveOccurred())
+			Expect(stderr).To(ContainSubstring("nvidia-container-cli.real: mount error: path error:"))
```
elezar, line 351: Let's add a test for the `nvidia-container-runtime` too:

```suggestion
        It("should not fail when using the nvidia-container-runtime-hook", func(ctx context.Context) {
            _, _, err := runner.Run("docker run --rm --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all -e NVIDIA_DRIVER_CAPABILITIES=all firmware-test")
            Expect(err).ToNot(HaveOccurred())
            
            output, _, _ := runner.Run(fmt.Sprintf("ls -A %s", outputDir))
			Expect(output).To(BeEmpty())
        }

		It("should fail when using the nvidia-container-runtime-hook", Label("legacy"), func(ctx context.Context) {
```

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -297,4 +298,51 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+
+			output, _, _ := runner.Run(fmt.Sprintf("ls -A %s", outputDir))
+			Expect(output).To(BeEmpty())
+		})
+	})
 })
```
elezar, line 354: Does it make sense to move the check for empty output folders to an AfterEach so that we don't have to repeat it for every test (and also don't forget to add it to new tests).

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -297,4 +298,56 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+			}
+		})
+
+		It("should not fail when using CDI", func(ctx context.Context) {
+			output, _, err := runner.Run("docker run --rm --runtime=nvidia --gpus=runtime.nvidia.com/gpu=all firmware-test")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(BeEmpty())
+		})
+
+		It("should not fail when using the nvidia-container-runtime-hook", func(ctx context.Context) {
```
elezar: This might have been a typo in my suggestion:
```suggestion
		It("should not fail when using the nvidia-container-runtime", func(ctx context.Context) {
```

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -297,4 +298,56 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+ARG RM_VERSION
+ARG CURRENT_DIR
+RUN ln -s /../../../../../../../../\$CURRENT_DIR /lib/firmware/nvidia/\$RM_VERSION
+EOF`)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		AfterEach(func(ctx context.Context) {
+			output, _, _ := runner.Run(fmt.Sprintf("ls -A %s", outputDir))
+			Expect(output).To(BeEmpty())
```
elezar: (the string concatenation is a nit)
```suggestion
			output, _, err := runner.Run("ls -A " + outputDir)
			Expect(err).ToNot(HaveOccurred())
			Expect(output).To(BeEmpty())
```

PR #1174, opened 2025-07-02
Allow update-ldcache hook to work when pivot-root is not supported

elezar:
The changes made to run the update-ldcache hook in an isolated namespace were implemented using a pivot_root. This is, however, not supported in all use cases -- most notably in kata-containers where runc is typically invoked with the --no-pivot option.

This change attempts to detect when pivot_root is not supported and use the same move mount operation as implemented by runc in this case.


internal/ldconfig/ldconfig.go
```diff
@@ -39,40 +42,66 @@ const (
...
 	return createReexecCommand(args)
 }
 
-// New creates an Ldconfig struct that is used to perform operations on the
-// ldcache and libraries in a particular root (e.g. a container).
-func New(ldconfigPath string, inRoot string) (*Ldconfig, error) {
-	l := &Ldconfig{
-		ldconfigPath: ldconfigPath,
-		inRoot:       inRoot,
+// NewFromRunnerArgs creates an Ldconfig struct from the args passed to the Cmd
```
ArangoGutierrez:
```suggestion
// NewFromArgs creates an Ldconfig struct from the args passed to the cmd
```

Or edit func name

ArangoGutierrez (changes requested):

internal/ldconfig/ldconfig_linux.go
```diff
@@ -99,6 +101,86 @@ func pivotRoot(rootfs string) error {
 	return nil
 }
 
+// msMoveRoot is used
+// filesystem, and everything else is cleaned up.
```
ArangoGutierrez: Nit: Can we enhance this comment?

internal/ldconfig/ldconfig_linux.go
```diff
@@ -99,6 +101,86 @@ func pivotRoot(rootfs string) error {
...
+	// Users shouldn't be using --no-pivot except in exceptional circumstances,
+	// but to avoid such a trivial security flaw we apply a best-effort
+	// protection here. The kernel only allows a mount of a pseudo-filesystem
+	// like procfs or sysfs if there is a *full* mount (the root of the
+	// filesystem is mounted) without any other locked mount points covering a
+	// subtree of the mount.
+	//
+	// So we try to unmount (or mount tmpfs on top of) any mountpoint which is
+	// a full mount of either sysfs or procfs (since those are the most
+	// concerning filesystems to us).
```
ArangoGutierrez, line 128:
```suggestion
	//In order to avoid the above-mentioned risks, we attempt to 
	// unmount (or mount tmpfs on top of) any mountpoint which is
	// a full mount of either sysfs or procfs (since those are the most
	// concerning filesystems to us).
```

elezar: This comment was taken as-is from runc and I don't want to update it to introduce unnecessary diffs.

elezar:
Note that the following is related: https://github.com/moby/moby/issues/50755#issuecomment-3205083033

ArangoGutierrez (approved):

zvonkok (approved):
LGTM,  tested on latest Kata release.

PR #1178, opened 2025-07-08
[no-relnote] Setup envfile for systemd services

jcpunk:
This updates the packages to include the referenced environment file for the systemd unit files.  Additionally it switches the exec to then use a referenced variable for the cdi output.

### DIFF
--- a/deployments/systemd/nvidia-cdi-refresh.service
+++ b/deployments/systemd/nvidia-cdi-refresh.service
@@ -20,9 +20,11 @@ ConditionPathExists=/usr/bin/nvidia-ctk
 
 [Service]
 Type=oneshot
-EnvironmentFile=-/etc/nvidia-container-toolkit/cdi-refresh.env
+# Values from Environment will be replaced if defined in EnvironmentFile
+Environment=NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/var/run/cdi/nvidia.yaml
+EnvironmentFile=-/etc/nvidia-container-toolkit/nvidia-cdi-refresh.env
 ExecCondition=/usr/bin/grep -qE '/nvidia.ko' /lib/modules/%v/modules.dep
-ExecStart=/usr/bin/nvidia-ctk cdi generate --output=/var/run/cdi/nvidia.yaml
+ExecStart=/usr/bin/nvidia-ctk cdi generate
 CapabilityBoundingSet=CAP_SYS_MODULE CAP_SYS_ADMIN CAP_MKNOD
 
 [Install]
--- a/packaging/rpm/SPECS/nvidia-container-toolkit.spec
+++ b/packaging/rpm/SPECS/nvidia-container-toolkit.spec
@@ -19,6 +19,7 @@ Source5: nvidia-container-runtime.legacy
 Source6: nvidia-cdi-hook
 Source7: nvidia-cdi-refresh.service
 Source8: nvidia-cdi-refresh.path
+Source9: nvidia-cdi-refresh.env
 
 Obsoletes: nvidia-container-runtime <= 3.5.0-1, nvidia-container-runtime-hook <= 1.4.0-2
 Provides: nvidia-container-runtime
@@ -35,6 +36,7 @@ cp %{SOURCE0} %{SOURCE1} %{SOURCE2} %{SOURCE3} %{SOURCE4} %{SOURCE5} %{SOURCE6} 
 %install
 mkdir -p %{buildroot}%{_bindir}
 mkdir -p %{buildroot}%{_sysconfdir}/systemd/system/
+mkdir -p %{buildroot}%{_sysconfdir}/nvidia-container-toolkit
 
 install -m 755 -t %{buildroot}%{_bindir} nvidia-container-runtime-hook
 install -m 755 -t %{buildroot}%{_bindir} nvidia-container-runtime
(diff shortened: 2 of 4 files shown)


deployments/systemd/cdi-refresh.env
```diff
@@ -0,0 +1,2 @@
+# Where should the generated CDI file be deployed
+CDI_GENERATE_OUTFILE=/var/run/cdi/nvidia.yaml
```
elezar: Does it make sense to set this to `NVIDIA_CTK_CDI_OUTPUT_FILE_PATH` as per https://github.com/NVIDIA/nvidia-container-toolkit/blob/35babbe02db5cedd8759a56cbd3668fc0ab8533a/cmd/nvidia-ctk/cdi/generate/generate.go#L111C1-L112C1 and then remove the `--output` flag?

deployments/systemd/cdi-refresh.env
elezar, line 1: Should this file be `nvidia-cdi-refresh.env`?

packaging/rpm/SPECS/nvidia-container-toolkit.spec
```diff
@@ -44,6 +45,7 @@ install -m 755 -t %{buildroot}%{_bindir} nvidia-ctk
 install -m 755 -t %{buildroot}%{_bindir} nvidia-cdi-hook
 install -m 644 -t %{buildroot}%{_sysconfdir}/systemd/system nvidia-cdi-refresh.service
 install -m 644 -t %{buildroot}%{_sysconfdir}/systemd/system nvidia-cdi-refresh.path
+install -m 644 -t %{buildroot}%{_sysconfdir}/nvidia-container-toolkit cdi-refresh.env
```
elezar: Is this the correct destination for this file?

deployments/systemd/cdi-refresh.env
jcpunk: That is better, I change this up

deployments/systemd/cdi-refresh.env
jcpunk: The file name was taken from the existing unit file in the repo.

packaging/rpm/SPECS/nvidia-container-toolkit.spec
jcpunk: Unless I'm reading it wrongly, I thought that was the path from the existing unit file?

deployments/systemd/cdi-refresh.env
elezar: OK. I see. Sorry I missed that when taking a quick look yesterday.

packaging/rpm/SPECS/nvidia-container-toolkit.spec
elezar: Sorry. Once again, I misread in my initial pass.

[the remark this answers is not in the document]
elezar: @jcpunk I think this is valid, or does `install` ensure that the target directory is created?

deployments/systemd/cdi-refresh.env
```diff
@@ -0,0 +1,2 @@
+# Where should the generated CDI file be deployed
```
ArangoGutierrez: I like this change, not hardcoding the --output

packaging/rpm/SPECS/nvidia-container-toolkit.spec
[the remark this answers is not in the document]
ArangoGutierrez: ++

packaging/rpm/SPECS/nvidia-container-toolkit.spec
```diff
@@ -105,6 +108,7 @@ Provides tools such as the NVIDIA Container Runtime and NVIDIA Container Toolkit
 %{_bindir}/nvidia-cdi-hook
 %{_sysconfdir}/systemd/system/nvidia-cdi-refresh.service
 %{_sysconfdir}/systemd/system/nvidia-cdi-refresh.path
+%config(noreplace) %{_sysconfdir}/nvidia-container-toolkit/cdi-refresh.env
```
elezar: As a matter of interest, could we do something similar for debian packages?

packaging/rpm/SPECS/nvidia-container-toolkit.spec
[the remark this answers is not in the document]
elezar: I went ahead and made the change and updated your branch.

packaging/rpm/SPECS/nvidia-container-toolkit.spec
jcpunk: I believe the default for files in `/etc` for dpkg is equivalent behavior.

jcpunk:
I made a handful of minor changes to try and be more explicit/self-documenting.

packaging/rpm/SPECS/nvidia-container-toolkit.spec
ArangoGutierrez: something like 
```bash
nvidia-cdi-refresh.service   /etc/systemd/system/
nvidia-cdi-refresh.path      /etc/systemd/system/
nvidia-cdi-refresh.env       /etc/nvidia-container-toolkit/
```

on the  debian file `nvidia-container-toolkit-base.install` would do it

jcpunk: Is that not what I've got?

deployments/systemd/nvidia-cdi-refresh.service
```diff
@@ -20,9 +20,11 @@ ConditionPathExists=/usr/bin/nvidia-ctk
 
 [Service]
 Type=oneshot
-EnvironmentFile=-/etc/nvidia-container-toolkit/cdi-refresh.env
+# Values from Environment will be replaced if defined in EnvironmentFile
```
elezar, line 23: Given this change, I'm not certain that we need an environment file to be installed. Or if at all we can create an empty file?

packaging/rpm/SPECS/nvidia-container-toolkit.spec
ArangoGutierrez: yes

PR #1202, opened 2025-07-21
[no-relnote] Fix E2E matrix conditional

ArangoGutierrez:


.github/workflows/e2e.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
     strategy:
       matrix:
         is_pull_request:
-          - ${{!( startsWith(github.ref, 'refs/pull/') )}}
+          - ${{ startsWith(github.ref, 'refs/pull/') }}
```
ArangoGutierrez: 🤔  Now the conditional returns false, checking....

ArangoGutierrez:
Ok, I think this PR fixes the matrix execution 
@elezar Ready for review

.github/workflows/ci.yaml
```diff
@@ -29,10 +29,24 @@ jobs:
...
         run: echo "version=$(echo $GITHUB_SHA | cut -c1-8)" >> "$GITHUB_OUTPUT"
+      - name: Determine if push is to main or release branch
+        id: is_pull_request
+        run: |
+          # Extract the branch name from the ref (e.g., refs/heads/main -> main)
+          BRANCH_NAME="${GITHUB_REF#refs/heads/}"
+
+          # If the branch is 'main' or starts with 'release-', this is _not_ a pull request-equivalent push
+          # Otherwise, treat it as a pull-request branch (e.g., 'pull-request/123')
+          if [[ "$BRANCH_NAME" == "main" || "$BRANCH_NAME" == release-* ]]; then
```
elezar: Does the equality operator here actually expand the wildcard?

ArangoGutierrez: https://www.gnu.org/software/bash/manual/bash.html#index-_005b_005b 

When the ‘==’ and ‘!=’ operators are used, the string to the right of the operator is considered a pattern and matched according to the rules described below in [Pattern Matching](https://www.gnu.org/software/bash/manual/bash.html#Pattern-Matching), as if the extglob shell option were enabled. The ‘=’ operator is identical to ‘==’. If the nocasematch shell option (see the description of shopt in [The Shopt Builtin](https://www.gnu.org/software/bash/manual/bash.html#The-Shopt-Builtin)) is enabled, the match is performed without regard to the case of alphabetic characters. The return value is 0 if the string matches (‘==’) or does not match (‘!=’) the pattern, and 1 otherwise.

.github/workflows/ci.yaml
```diff
@@ -29,10 +29,24 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       version: ${{ steps.version.outputs.version }}
+      is_pull_request: ${{ steps.is_pull_request.outputs.is_pull_request }}
```
elezar: One quesiton: In [`image.yaml`](https://github.com/NVIDIA/nvidia-container-toolkit/blob/09200187e5bbe84b9662904337a24941dbe90607/.github/workflows/image.yaml#L40-L41) we already have
```
        ispr:
          - ${{ github.ref_name != 'main' && !startsWith( github.ref_name, 'release-' ) }}
```
Is there a reason that we don't just use that here again too?

ArangoGutierrez: will edit; I didn't saw that

elezar (approved):

Issue #1215, opened 2025-07-31
CDI Spec Generation is missing driver libraries


elezar:
See https://github.com/NVIDIA/k8s-dra-driver-gpu/issues/446 where certain driver libraries are reported as missing when generating CDI specifications. The issue is specifically focussed on a GKE driver installation which may mean that the libraries are in different locations, but these are also driver libraries that are not covered by our strategy of including everything that has a driver version suffix.

> libEGL.so.1.1.0, libGL.so.1.7.0, libGLESv1_CM.so.1.2.0, libGLESv2.so.2.1.0, libGLX.so.0, libGLdispatch.so.0, libOpenCL.so.1.0.0, libOpenGL.so.0, libnvidia-api.so.1, libnvidia-egl-xcb.so.1.0.0, libnvidia-egl-xlib.so.1.0.0

We should compare these libraries with a driver manifest and ensure that they are explicitly added to the list of discovered libraries if required.

For example, the `575.64.05` manifest shows:
```
libOpenCL.so.1.0.0 0755 OPENCL_WRAPPER_LIB NATIVE / MODULE:gpgpu
libOpenCL.so.1.0 0000 OPENCL_WRAPPER_SYMLINK NATIVE / libOpenCL.so.1.0.0 MODULE:gpgpu
libOpenCL.so.1 0000 OPENCL_WRAPPER_SYMLINK NATIVE / libOpenCL.so.1.0 MODULE:gpgpu
libOpenCL.so 0000 OPENCL_WRAPPER_SYMLINK NATIVE / libOpenCL.so.1 MODULE:gpgpu
```
indicating that the `libOpenCL.so.1.0.0` (and it's related symlinks) should be included in the container. (note that adding the `libOpenCL.so.1.0` symlink to the chain could be considered out of scope for this issue, but should be considered as a separate issue then).

PR #1216, opened 2025-07-31
Add explicitLibs list to libs discovery

ArangoGutierrez:
Fixes:  #1215

Add a list of explicit libraries that don't share the libcuda version suffix.

This can be considered an early iteration; a future iteration should read the libraries from nvsandboxutils to avoid hardcoding the library version suffix.

### Enhancements to library management:
* Added a list of explicit library names to be located, including `libEGL.so.1.1.0`, `libGL.so.1.7.0`, `libGLESv1_CM.so.1.2.0`, and others. These libraries are critical for the driver version being processed.
* Implemented a loop to locate each library in the list, appending found libraries to the main `libs` list and logging warnings for any missing libraries without causing errors. This ensures better visibility into missing dependencies while maintaining robustness.

### DIFF
--- a/pkg/nvcdi/driver-nvml.go
+++ b/pkg/nvcdi/driver-nvml.go
@@ -150,17 +150,17 @@ func (l *nvcdilib) getExplicitDriverLibraryMounts() (discover.Discover, error) {
 	// the sandboxutils-filelist or have a way to allow users to specify the
 	// libraries to mount from the config file.
 	explicitLibraries := []string{
-		"libEGL.so.1.1.0",
-		"libGL.so.1.7.0",
-		"libGLESv1_CM.so.1.2.0",
-		"libGLESv2.so.2.1.0",
-		"libGLX.so.0",
-		"libGLdispatch.so.0",
-		"libOpenCL.so.1.0.0",
-		"libOpenGL.so.0",
-		"libnvidia-api.so.1",
-		"libnvidia-egl-xcb.so.1.0.0",
-		"libnvidia-egl-xlib.so.1.0.0",
+		"libEGL.so",
+		"libGL.so",
+		"libGLESv1_CM.so",
+		"libGLESv2.so",
+		"libGLX.so",
+		"libGLdispatch.so",
+		"libOpenCL.so",
+		"libOpenGL.so",
+		"libnvidia-api.so",
+		"libnvidia-egl-xcb.so",
+		"libnvidia-egl-xlib.so",
 	}
 
 	driverLibraryLocator, err := l.driver.DriverLibraryLocator()
--- a/pkg/nvcdi/driver-nvml.go
+++ b/pkg/nvcdi/driver-nvml.go
@@ -76,19 +76,18 @@ func (l *nvcdilib) newDriverVersionDiscoverer() (discover.Discover, error) {
 
 // NewDriverLibraryDiscoverer creates a discoverer for the libraries associated with the specified driver version.
 func (l *nvcdilib) NewDriverLibraryDiscoverer(version string, libcudaSoParentDirPath string) (discover.Discover, error) {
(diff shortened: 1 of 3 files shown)


pkg/nvcdi/driver-nvml.go
```diff
@@ -235,6 +235,39 @@ func getVersionLibs(logger logger.Interface, driver *root.Driver, version string
...
+
+	// Locate each explicit library
+	var explicitLibs []string
+	for _, libName := range explicitLibraryNames {
+		found, _ := libraries.Locate(libName)
+		// Since we are using WithOptional, we don't expect an error, but we still
+		// log a warning if the library is not found.
+		if len(found) == 0 {
+			// Log warning when library is not found (but no error due to WithOptional)
+			logger.Warningf("library %s not found in search paths", libName)
```
elezar: This log line specifically makes me think that this is the incorrect place to do this. Note that when using the mount discoverer in the calling function we get logging similar to:
```
INFO[0000] Selecting /usr/lib/x86_64-linux-gnu/libnvidia-ptxjitcompiler.so.570.133.20 as /usr/lib/x86_64-linux-gnu/libnvidia-ptxjitcompiler.so.570.133.20
WARN[0000] Could not locate /nvidia-persistenced/socket: pattern /nvidia-persistenced/socket not found
```
Meaning that ideally we want to pass this list directly into a `Mounts` discoverer instead.

I would thus suggest updating the implementation to rather construct a second discoverer where call `getVersionLibs` and pass the list of libraries in there. Something like:
```
	versionSuffixLibraryPaths, libCudaDirectoryPath, err := getVersionLibs(l.logger, l.driver, version)
	if err != nil {
		return nil, fmt.Errorf("failed to get libraries for driver version: %v", err)
	}

	versionSuffixLibraryMounts := discover.NewMounts(
		l.logger,
		lookup.NewFileLocator(
			lookup.WithLogger(l.logger),
			lookup.WithRoot(l.driver.Root),
		),
		l.driver.Root,
		versionSuffixLibraryPaths,
	)

	explicitLibraries := []string{}
	explicitLibraryMounts := discover.NewMounts(
		l.logger,
		l.driver.Libraries(),
		l.driver.Root,
		explicitLibraries,
	)

	libraries := discover.Merge(
		versionSuffixLibraryMounts,
		explicitLibraryMounts,
	)
```
The one issue may be that we're not STRICTLY guaranteed that the same parent as `libcuda.so.VERSION` will be used, but this could be addressed by further refactoring in a follow-up.

ArangoGutierrez: Addressed

cmd/nvidia-ctk/cdi/generate/generate_test.go
```diff
@@ -315,6 +387,30 @@ containerEdits:
...
+            - rprivate
+        - hostPath: {{ .driverRoot }}/lib/x86_64-linux-gnu/libGLdispatch.so.0
+          containerPath: /lib/x86_64-linux-gnu/libGLdispatch.so.0
+          options:
+            - ro
+            - nosuid
+            - nodev
+            - rbind
+            - rprivate
+        - hostPath: {{ .driverRoot }}/lib/x86_64-linux-gnu/libOpenGL.so.0
```
elezar: I find it strange that these are listed before `libcuda.so` since we *should* be calling the version lib first. One other thing ... could it be that we're detecting these on the host where the tests are running? Do these tests pass on your macbook?

ArangoGutierrez: Yes, `make lint coverage` passes successfully on my local host

ArangoGutierrez: solved with the latest revision

elezar (approved):

PR #1221, opened 2025-08-04
Disable chmod-hook by default

### LINKED ISSUE #1218: Disable chmod hook by default
The chmod `createContainer` hook was added as a workaround to a specific `crun` issue for device nodes that are found in subdirectories of `/dev`. Since this has since been addressed in `crun`, we should disable this hook by default to reduce the number of supported hooks.

This means that by default the:
1. runtime will not add the `chmod` hook to the incomming OCI runtime spec.
2. CDI spec generation will not generate a `chmod` hook in the CDI spec.

We should:
1. Provide a `feature` flag that allows users to opt-in to hook injection.
2. Provide a mechanism of enabling this hook in the `nvcdi` API
3. Provide a command line option to the `nvidia-ctk cdi generate` command that enables the hook.

ArangoGutierrez:
Fixes #1218 

The chmod createContainer hook was added as a workaround to a specific `crun` issue for device nodes that are found in subdirectories of `/dev`. Since this has since been addressed in `crun`, this PR disables the chmod hook by default to reduce the number of supported hooks.

For users that *require* this hook, it is recommended that `cdi` be used and that the `--enable-hook=chmod` is passed to the `nvidia-ctk cdi generate` command.

See:
* https://gitlab.com/nvidia/container-toolkit/container-toolkit/-/merge_requests/232
* https://gitlab.com/nvidia/container-toolkit/container-toolkit/-/issues/8#note_1137929772
* https://github.com/containers/crun/issues/1047#issuecomment-1299934754
* https://github.com/containers/crun/pull/1051
* https://github.com/containers/crun/releases/tag/1.7

### DIFF
--- a/cmd/nvidia-ctk/cdi/generate/generate.go
+++ b/cmd/nvidia-ctk/cdi/generate/generate.go
@@ -62,6 +62,7 @@ type options struct {
 	configSearchPaths  []string
 	librarySearchPaths []string
 	disabledHooks      []string
+	enabledHooks       []string
 
 	csv struct {
 		files          []string
@@ -214,6 +215,13 @@ func (m command) build() *cli.Command {
 				Destination: &opts.disabledHooks,
 				Sources:     cli.EnvVars("NVIDIA_CTK_CDI_GENERATE_DISABLED_HOOKS"),
 			},
+			&cli.StringSliceFlag{
+				Name:        "enable-hook",
+				Aliases:     []string{"enable-hooks"},
+				Usage:       "Explicitly enable a hook in the generated CDI specification. This overrides disabled hooks. This can be specified multiple times.",
+				Destination: &opts.enabledHooks,
+				Sources:     cli.EnvVars("NVIDIA_CTK_CDI_GENERATE_ENABLED_HOOKS"),
+			},
 		},
 	}
 
@@ -257,6 +265,12 @@ func (m command) validateFlags(c *cli.Command, opts *options) error {
 	}
 	if err := cdi.ValidateClassName(opts.class); err != nil {
 		return fmt.Errorf("invalid CDI class name: %v", err)
+	}
+
+	for _, hook := range opts.enabledHooks {
+		if hook == "all" {
+			return fmt.Errorf("enabling all hooks is not supported")
+		}
 	}
 	return nil
 }
@@ -313,12 +327,10 @@ func (m command) generateSpec(opts *options) (spec.Interface, error) {
 		nvcdi.WithLibrarySearchPaths(opts.librarySearchPaths),
 		nvcdi.WithCSVFiles(opts.csv.files),
 		nvcdi.WithCSVIgnorePatterns(opts.csv.ignorePatterns),
+		nvcdi.WithDisabledHooks(opts.disabledHooks...),
+		nvcdi.WithEnabledHooks(opts.enabledHooks...),
 		// We set the following to allow for dependency injection:
 		nvcdi.WithNvmlLib(opts.nvmllib),
-	}
-
-	for _, hook := range opts.disabledHooks {
-		cdiOptions = append(cdiOptions, nvcdi.WithDisabledHook(hook))
 	}
 
 	cdilib, err := nvcdi.New(cdiOptions...)
--- a/internal/discover/hooks.go
+++ b/internal/discover/hooks.go
@@ -50,6 +50,14 @@ const (
 	defaultNvidiaCDIHookPath = "/usr/bin/nvidia-cdi-hook"
 )
 
+// defaultDisabledHooks defines hooks that are disabled by default.
+// These hooks can be explicitly enabled using the WithEnabledHooks option.
+var defaultDisabledHooks = []HookName{
+	// ChmodHook is disabled by default as it was a workaround for older
+	// versions of crun that has since been fixed.
+	ChmodHook,
+}
+
 var _ Discover = (*Hook)(nil)
 
 // Devices returns an empty list of devices for a Hook discoverer.
@@ -77,7 +85,14 @@ func (h *Hook) Hooks() ([]Hook, error) {
 	return []Hook{*h}, nil
 }
 
-type Option func(*cdiHookCreator)
+type hookCreatorOptions struct {
+	nvidiaCDIHookPath string
+	disabledHooks     []HookName
+	enabledHooks      []HookName
+	debugLogging      bool
+}
+
+type Option func(*hookCreatorOptions)
 
 type cdiHookCreator struct {
 	nvidiaCDIHookPath string
@@ -100,39 +115,66 @@ type HookCreator interface {
 	Create(HookName, ...string) *Hook
 }
 
-// WithDisabledHooks sets the set of hooks that are disabled for the CDI hook creator.
+func WithDebugLogging(debugLogging bool) Option {
+	return func(hco *hookCreatorOptions) {
+		hco.debugLogging = debugLogging
+	}
+}
+
+// WithDisabledHooks explicitly disables the specified hooks.
 // This can be specified multiple times.
 func WithDisabledHooks(hooks ...HookName) Option {
-	return func(c *cdiHookCreator) {
-		for _, hook := range hooks {
-			c.disabledHooks[hook] = true
-		}
+	return func(c *hookCreatorOptions) {
+		c.disabledHooks = append(c.disabledHooks, hooks...)
+	}
+}
+
+// WithEnabledHooks explicitly enables the specified hooks.
+// This is useful for enabling hooks that are disabled by default.
+func WithEnabledHooks(hooks ...HookName) Option {
+	return func(c *hookCreatorOptions) {
+		c.enabledHooks = append(c.enabledHooks, hooks...)
 	}
 }
 
 // WithNVIDIACDIHookPath sets the path to the nvidia-cdi-hook binary.
 func WithNVIDIACDIHookPath(nvidiaCDIHookPath string) Option {
-	return func(c *cdiHookCreator) {
+	return func(c *hookCreatorOptions) {
 		c.nvidiaCDIHookPath = nvidiaCDIHookPath
 	}
 }
 
 func NewHookCreator(opts ...Option) HookCreator {
-	cdiHookCreator := &cdiHookCreator{
+	o := &hookCreatorOptions{
 		nvidiaCDIHookPath: defaultNvidiaCDIHookPath,
-		disabledHooks:     make(map[HookName]bool),
 	}
 	for _, opt := range opts {
-		opt(cdiHookCreator)
+		opt(o)
 	}
 
-	if cdiHookCreator.disabledHooks[AllHooks] {
+	o.disabledHooks = append(o.disabledHooks, defaultDisabledHooks...)
+
+	disabledHooks := make(map[HookName]bool)
+	for _, h := range o.disabledHooks {
+		disabledHooks[h] = true
+	}
+
+	if disabledHooks[AllHooks] && len(o.enabledHooks) == 0 {
 		return &allDisabledHookCreator{}
 	}
 
-	cdiHookCreator.fixedArgs = getFixedArgsForCDIHookCLI(cdiHookCreator.nvidiaCDIHookPath)
+	for _, h := range o.enabledHooks {
+		disabledHooks[h] = false
+	}
 
-	return cdiHookCreator
+	c := &cdiHookCreator{
+		nvidiaCDIHookPath: o.nvidiaCDIHookPath,
+		disabledHooks:     disabledHooks,
+		fixedArgs:         getFixedArgsForCDIHookCLI(o.nvidiaCDIHookPath),
+		debugLogging:      o.debugLogging,
+	}
+
+	return c
 }
 
 // Create creates a new hook with the given name and arguments.
@@ -150,21 +192,19 @@ func (c cdiHookCreator) Create(name HookName, args ...string) *Hook {
 	}
 }
 
-// isDisabled checks if the specified hook name is disabled.
 func (c cdiHookCreator) isDisabled(name HookName, args ...string) bool {
-	if c.disabledHooks[name] {
+	disabled, ok := c.disabledHooks[name]
+	if ok {
+		return disabled
+	}
+	if c.disabledHooks[AllHooks] {
 		return true
 	}
 
+	// still reject hooks that require args if none were provided
 	switch name {
-	case CreateSymlinksHook:
-		if len(args) == 0 {
-			return true
-		}
-	case ChmodHook:
-		if len(args) == 0 {
-			return true
-		}
+	case CreateSymlinksHook, ChmodHook:
+		return len(args) == 0
 	}
 	return false
 }
--- a/internal/discover/hooks_test.go
+++ b/internal/discover/hooks_test.go
@@ -1,0 +1,469 @@
+/**
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package discover
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewHookCreator(t *testing.T) {
+	testCases := []struct {
+		name     string
+		opts     []Option
+		expected HookCreator
+	}{
+		{
+			name: "default creates cdiHookCreator with default path",
+			opts: nil,
+			expected: &cdiHookCreator{
+				nvidiaCDIHookPath: defaultNvidiaCDIHookPath,
+				fixedArgs:         []string{"nvidia-cdi-hook"},
+				disabledHooks: map[HookName]bool{
+					ChmodHook: true, // ChmodHook is disabled by default
+				},
+			},
+		},
+		{
+			name: "custom nvidia-cdi-hook path",
+			opts: []Option{
+				WithNVIDIACDIHookPath("/custom/path/nvidia-cdi-hook"),
+			},
+			expected: &cdiHookCreator{
+				nvidiaCDIHookPath: "/custom/path/nvidia-cdi-hook",
+				fixedArgs:         []string{"nvidia-cdi-hook"},
+				disabledHooks: map[HookName]bool{
+					ChmodHook: true,
+				},
+			},
+		},
+		{
+			name: "AllHooks disabled returns allDisabledHookCreator",
+			opts: []Option{
+				WithDisabledHooks(AllHooks),
+			},
+			expected: &allDisabledHookCreator{},
+		},
+		{
+			name: "AllHooks disabled but UpdateLDCacheHook enabled",
+			opts: []Option{
+				WithDisabledHooks(AllHooks),
+				WithEnabledHooks(UpdateLDCacheHook),
+			},
+			expected: &cdiHookCreator{
+				nvidiaCDIHookPath: defaultNvidiaCDIHookPath,
+				fixedArgs:         []string{"nvidia-cdi-hook"},
+				disabledHooks: map[HookName]bool{
+					AllHooks:          true,
+					UpdateLDCacheHook: false,
+					ChmodHook:         true,
+				},
+			},
+		},
+		{
+			name: "multiple hooks disabled and enabled",
+			opts: []Option{
+				WithDisabledHooks(UpdateLDCacheHook, CreateSymlinksHook, EnableCudaCompatHook, DisableDeviceNodeModificationHook),
+				WithEnabledHooks(ChmodHook, UpdateLDCacheHook),
+			},
+			expected: &cdiHookCreator{
+				nvidiaCDIHookPath: defaultNvidiaCDIHookPath,
+				fixedArgs:         []string{"nvidia-cdi-hook"},
+				disabledHooks: map[HookName]bool{
+					UpdateLDCacheHook:                 false,
+					CreateSymlinksHook:                true,
+					EnableCudaCompatHook:              true,
+					ChmodHook:                         false,
+					DisableDeviceNodeModificationHook: true,
+				},
+			},
+		},
+		{
+			name: "WithDisabledHooks can be called multiple times",
+			opts: []Option{
+				WithDisabledHooks(UpdateLDCacheHook),
+				WithDisabledHooks(CreateSymlinksHook),
+				WithDisabledHooks(EnableCudaCompatHook),
+			},
+			expected: &cdiHookCreator{
+				nvidiaCDIHookPath: defaultNvidiaCDIHookPath,
+				fixedArgs:         []string{"nvidia-cdi-hook"},
+				disabledHooks: map[HookName]bool{
+					UpdateLDCacheHook:    true,
+					CreateSymlinksHook:   true,
+					EnableCudaCompatHook: true,
+					ChmodHook:            true, // Default disabled
+				},
+			},
+		},
+		{
+			name: "WithEnabledHooks overrides defaultDisabledHooks",
+			opts: []Option{
+				WithEnabledHooks(ChmodHook),
+			},
+			expected: &cdiHookCreator{
+				nvidiaCDIHookPath: defaultNvidiaCDIHookPath,
+				fixedArgs:         []string{"nvidia-cdi-hook"},
+				disabledHooks: map[HookName]bool{
+					ChmodHook: false, // ChmodHook is enabled
+				},
+			},
+		},
+		{
+			name: "nvidia-ctk binary path affects fixed args",
+			opts: []Option{
+				WithNVIDIACDIHookPath("/usr/bin/nvidia-ctk"),
+			},
+			expected: &cdiHookCreator{
+				nvidiaCDIHookPath: "/usr/bin/nvidia-ctk",
+				fixedArgs:         []string{"nvidia-ctk", "hook"},
+				disabledHooks: map[HookName]bool{
+					ChmodHook: true,
+				},
+			},
+		},
+		{
+			name: "nvidia-ctk in custom NVIDIA toolkit path",
+			opts: []Option{
+				WithNVIDIACDIHookPath("/usr/local/nvidia/toolkit/nvidia-ctk"),
+			},
+			expected: &cdiHookCreator{
+				nvidiaCDIHookPath: "/usr/local/nvidia/toolkit/nvidia-ctk",
+				fixedArgs:         []string{"nvidia-ctk", "hook"},
+				disabledHooks: map[HookName]bool{
+					ChmodHook: true,
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			hookCreator := NewHookCreator(tc.opts...)
+
+			require.EqualValues(t, tc.expected, hookCreator)
+		})
+	}
+}
+func TestCDIHookCreator_Create(t *testing.T) {
+	testCases := []struct {
+		name         string
+		hookCreator  HookCreator
+		hookName     HookName
+		args         []string
+		expectedHook *Hook
+	}{
+		{
+			name:        "CreateSymlinksHook with args",
+			hookCreator: NewHookCreator(WithNVIDIACDIHookPath(defaultNvidiaCDIHookPath)),
+			hookName:    CreateSymlinksHook,
+			args:        []string{"/source::/target", "/source2::/target2"},
+			expectedHook: &Hook{
+				Lifecycle: "createContainer",
+				Path:      "/usr/bin/nvidia-cdi-hook",
+				Args:      []string{"nvidia-cdi-hook", "create-symlinks", "--link", "/source::/target", "--link", "/source2::/target2"},
+				Env:       []string{"NVIDIA_CTK_DEBUG=false"},
+			},
+		},
+		{
+			name:         "CreateSymlinksHook without args returns nil",
+			hookCreator:  NewHookCreator(WithNVIDIACDIHookPath(defaultNvidiaCDIHookPath)),
+			hookName:     CreateSymlinksHook,
+			args:         []string{},
+			expectedHook: nil,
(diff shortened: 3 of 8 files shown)


cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -214,6 +215,12 @@ func (m command) build() *cli.Command {
 				Destination: &opts.disabledHooks,
 				Sources:     cli.EnvVars("NVIDIA_CTK_CDI_GENERATE_DISABLED_HOOKS"),
 			},
+			&cli.BoolFlag{
+				Name:        "enable-chmod-hook",
+				Usage:       "Enable the chmod hook for device folder permissions. This hook is disabled by default.",
+				Destination: &opts.enableChmodHook,
+				Sources:     cli.EnvVars("NVIDIA_CTK_CDI_GENERATE_ENABLE_CHMOD_HOOK"),
+			},
```
elezar: Question: Why not have a `--force-hook[s]` option that overrides the `--disable-hook[s]` option above instead of implementing a single flag for each hook?

internal/config/features_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expected, tc.feature.IsEnabled())
+		})
+	}
+}
+
+func TestEnableChmodHookFeature(t *testing.T) {
```
elezar: Could you comment on whether this adds any specific value given the test cases above?

internal/discover/hooks.go
```diff
@@ -82,6 +82,7 @@ type Option func(*cdiHookCreator)
 type cdiHookCreator struct {
 	nvidiaCDIHookPath string
 	disabledHooks     map[HookName]bool
+	enableChmodHook   bool
```
elezar: As mentioned in the addition of the flag. Does maintaining a map of "forced" hooks make more sense here than handling each hook separately?

internal/runtime/runtime_factory.go
```diff
@@ -75,7 +75,10 @@ func newSpecModifier(logger logger.Interface, cfg *config.Config, ociSpec oci.Sp
 		return nil, err
 	}
 
-	hookCreator := discover.NewHookCreator(discover.WithNVIDIACDIHookPath(cfg.NVIDIACTKConfig.Path))
+	hookCreator := discover.NewHookCreator(
+		discover.WithNVIDIACDIHookPath(cfg.NVIDIACTKConfig.Path),
+		discover.WithEnableChmodHook(cfg.Features.EnableChmodHook.IsEnabled()),
```
elezar: Does this indicate that we may need something like the following in the runtime config?

```
[nvidia-container-runtime.hooks]

disabled = [...]
# Specify a list of explicitly enabled hooks. If a hook is specified as both enabled and disabled, it will be enabled.
enabled = [...]
```

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -214,6 +215,13 @@ func (m command) build() *cli.Command {
 				Destination: &opts.disabledHooks,
 				Sources:     cli.EnvVars("NVIDIA_CTK_CDI_GENERATE_DISABLED_HOOKS"),
 			},
+			&cli.StringSliceFlag{
+				Name:        "enable-hook",
+				Aliases:     []string{"enable-hooks"},
+				Usage:       "Explicitly enable a hook in the generated CDI specification. This can be specified multiple times.",
```
elezar:
```suggestion
				Usage:       "Explicitly enable a hook in the generated CDI specification. This overrides disabled hooks. This can be specified multiple times.",
```

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -292,6 +300,33 @@ func formatFromFilename(filename string) string {
...
+			if disabledHooks := m.config.Get("nvidia-container-runtime-hook.disabled"); disabledHooks != nil {
+				if hooks, ok := disabledHooks.([]interface{}); ok {
+					for _, hook := range hooks {
+						if h, ok := hook.(string); ok {
+							opts.disabledHooks = append(opts.disabledHooks, h)
+						}
+					}
+				}
+			}
+		}
```
elezar: Why is this required? I thought we added support to read settings from a config file using `altsrc`.

internal/config/hook.go
```diff
@@ -23,4 +23,12 @@ type RuntimeHookConfig struct {
...
 	// SkipModeDetection disables the mode check for the runtime hook.
 	SkipModeDetection bool `toml:"skip-mode-detection"`
+	// DisabledHooks specifies the hooks that are disabled for the NVIDIA
+	// Container Runtime hook.
+	DisabledHooks []string `toml:"disabled,omitempty"`
+	// EnabledHooks specifies the hooks that are enabled for the NVIDIA
+	// Container Runtime hook.
+	// Specify a list of explicitly enabled hooks. If a hook is specified as
+	// both enabled and disabled, it will be enabled.
+	EnabledHooks []string `toml:"enabled,omitempty"`
```
elezar: I don't think we should add these config options, and IF we decide to add them we should not add them in the `nvidia-container-runtime-hook` section. This section is specific for the `nvidia-container-runtime-hook` that is used to call out to the `nvidia-container-cli`.

internal/discover/hooks.go
```diff
@@ -100,12 +110,12 @@ type HookCreator interface {
...
 }
 
-// WithDisabledHooks sets the set of hooks that are disabled for the CDI hook creator.
+// WithDisabledHooks explicitly disables the specified hooks.
 // This can be specified multiple times.
 func WithDisabledHooks(hooks ...HookName) Option {
 	return func(c *cdiHookCreator) {
 		for _, hook := range hooks {
-			c.disabledHooks[hook] = true
+			c.hookStates[hook] = false
```
elezar: This implies ordering between the calls to `WithDisabledHooks` and `WithEnabledHooks`. I would rather have the order or precedence well-defined such that hooks that are specifically ENABLED take precedence over hooks that are specifically DISABLED.

elezar: Maybe this can be simplified by keeping the `cdiHookCreator` struct as is, and introducing an `options` struct that we can use to set up the correct values of `disabledHooks`.

cmd/nvidia-ctk/cdi/generate/generate.go
ArangoGutierrez: delete

cmd/nvidia-ctk/cdi/generate/generate.go
ArangoGutierrez: added

internal/config/hook.go
ArangoGutierrez: removed for now, maybe a discuss-todo item

internal/discover/hooks.go
ArangoGutierrez: you are right, let me think on a proper solution for this

ArangoGutierrez: Done

internal/config/hook.go
elezar: Did you forget to push?

ArangoGutierrez: yes...

ArangoGutierrez: Sorry, now pushed

internal/config/hook.go
```diff
@@ -23,4 +23,7 @@ type RuntimeHookConfig struct {
 	Path string `toml:"path"`
 	// SkipModeDetection disables the mode check for the runtime hook.
 	SkipModeDetection bool `toml:"skip-mode-detection"`
+	// DisabledHooks specifies the hooks that are disabled for the NVIDIA
+	// Container Runtime hook.
+	DisabledHooks []string `toml:"disabled,omitempty"`
```
elezar: I thought we said we're going to remove this?

internal/discover/hooks.go
```diff
@@ -50,6 +50,14 @@ const (
 	defaultNvidiaCDIHookPath = "/usr/bin/nvidia-cdi-hook"
 )
 
+// defaultDisabledHooks defines hooks that are disabled by default.
+// These hooks can be explicitly enabled using the WithEnabledHooks option.
+var defaultDisabledHooks = map[HookName]bool{
```
elezar: This need not be a map. Also does it make sense to just set this inline when constructing the hook creator?

internal/config/hook.go
ArangoGutierrez: Removed

internal/discover/hooks.go
ArangoGutierrez: turned into a list.
I do think it is better to have it here, so the list is easier to maintain. Over time, we might want to add other hooks to the list.

internal/discover/hooks.go
```diff
@@ -118,14 +137,27 @@ func WithNVIDIACDIHookPath(nvidiaCDIHookPath string) Option {
...
+		disabledHooks:     disabledHooks,
+		enabledHooks:      enabledHooks,
 	}
 	for _, opt := range opts {
 		opt(cdiHookCreator)
 	}
 
+	// Correct the disabledHooks map to ensure that explicitly enabled hooks
+	// are not disabled.
+	for hook := range enabledHooks {
```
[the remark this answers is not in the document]
elezar: Does this mean that we don't have a test for the case where enabled overrides disabled?

internal/runtime/runtime_factory.go
```diff
@@ -75,7 +75,11 @@ func newSpecModifier(logger logger.Interface, cfg *config.Config, ociSpec oci.Sp
 		return nil, err
 	}
 
-	hookCreator := discover.NewHookCreator(discover.WithNVIDIACDIHookPath(cfg.NVIDIACTKConfig.Path))
+	opts := []discover.Option{
```
elezar: Let's revert this change.

internal/discover/hooks.go
```diff
@@ -100,7 +109,7 @@ type HookCreator interface {
 	Create(HookName, ...string) *Hook
 }
 
-// WithDisabledHooks sets the set of hooks that are disabled for the CDI hook creator.
+// WithDisabledHooks explicitly disables the specified hooks.
 // This can be specified multiple times.
 func WithDisabledHooks(hooks ...HookName) Option {
```
elezar, line 126: Let's update these to also use generics so that we can pass `...string` or `...HookName`.

internal/discover/hooks.go
```diff
@@ -82,6 +90,7 @@ type Option func(*cdiHookCreator)
 type cdiHookCreator struct {
 	nvidiaCDIHookPath string
 	disabledHooks     map[HookName]bool
+	enabledHooks      map[HookName]bool
```
elezar: Since this is only actually used in the construction of the `cdiHookCreator` and doesn't affect behaviour, does it make sense to split out separate `options` instead of updating the struct directly? Something like:
```
diff --git a/internal/discover/hooks.go b/internal/discover/hooks.go
index acfaef86..ada69fb8 100644
--- a/internal/discover/hooks.go
+++ b/internal/discover/hooks.go
@@ -85,12 +85,17 @@ func (h *Hook) Hooks() ([]Hook, error) {
 	return []Hook{*h}, nil
 }
 
-type Option func(*cdiHookCreator)
+type hookCreatorOptions struct {
+	nvidiaCDIHookPath string
+	disabledHooks     []HookName
+	enabledHooks      []HookName
+}
+
+type Option func(*hookCreatorOptions)
 
 type cdiHookCreator struct {
 	nvidiaCDIHookPath string
 	disabledHooks     map[HookName]bool
-	enabledHooks      map[HookName]bool
 
 	fixedArgs    []string
 	debugLogging bool
@@ -111,60 +116,61 @@ type HookCreator interface {
 
 // WithDisabledHooks explicitly disables the specified hooks.
 // This can be specified multiple times.
-func WithDisabledHooks(hooks ...HookName) Option {
-	return func(c *cdiHookCreator) {
-		for _, hook := range hooks {
-			c.disabledHooks[hook] = true
+func WithDisabledHooks[T string | HookName](hooks ...T) Option {
+	return func(c *hookCreatorOptions) {
+		for _, h := range hooks {
+			c.disabledHooks = append(c.disabledHooks, HookName(h))
 		}
 	}
 }
 
 // WithEnabledHooks explicitly enables the specified hooks.
 // This is useful for enabling hooks that are disabled by default.
-func WithEnabledHooks(hooks ...HookName) Option {
-	return func(c *cdiHookCreator) {
-		for _, hook := range hooks {
-			c.enabledHooks[hook] = true
+func WithEnabledHooks[T string | HookName](hooks ...T) Option {
+	return func(c *hookCreatorOptions) {
+		for _, h := range hooks {
+			c.enabledHooks = append(c.disabledHooks, HookName(h))
 		}
 	}
 }
 
 // WithNVIDIACDIHookPath sets the path to the nvidia-cdi-hook binary.
 func WithNVIDIACDIHookPath(nvidiaCDIHookPath string) Option {
-	return func(c *cdiHookCreator) {
+	return func(c *hookCreatorOptions) {
 		c.nvidiaCDIHookPath = nvidiaCDIHookPath
 	}
 }
 
 func NewHookCreator(opts ...Option) HookCreator {
-	disabledHooks := make(map[HookName]bool)
-	enabledHooks := make(map[HookName]bool)
-	for _, hook := range defaultDisabledHooks {
-		disabledHooks[hook] = true
-	}
-
-	cdiHookCreator := &cdiHookCreator{
+	o := &hookCreatorOptions{
 		nvidiaCDIHookPath: defaultNvidiaCDIHookPath,
-		disabledHooks:     disabledHooks,
-		enabledHooks:      enabledHooks,
 	}
 	for _, opt := range opts {
-		opt(cdiHookCreator)
+		opt(o)
 	}
 
-	// Correct the disabledHooks map to ensure that explicitly enabled hooks
-	// are not disabled.
-	for hook := range enabledHooks {
-		cdiHookCreator.disabledHooks[hook] = false
+	o.disabledHooks = append(o.disabledHooks, defaultDisabledHooks...)
+
+	disabledHooks := make(map[HookName]bool)
+	for _, h := range o.disabledHooks {
+		disabledHooks[h] = true
 	}
 
-	if cdiHookCreator.disabledHooks[AllHooks] {
+	if disabledHooks[AllHooks] && len(o.enabledHooks) == 0 {
 		return &allDisabledHookCreator{}
 	}
 
-	cdiHookCreator.fixedArgs = getFixedArgsForCDIHookCLI(cdiHookCreator.nvidiaCDIHookPath)
+	for _, h := range o.enabledHooks {
+		disabledHooks[h] = false
+	}
+
+	c := &cdiHookCreator{
+		nvidiaCDIHookPath: o.nvidiaCDIHookPath,
+		disabledHooks:     make(map[HookName]bool),
+		fixedArgs:         getFixedArgsForCDIHookCLI(o.nvidiaCDIHookPath),
+	}
 
-	return cdiHookCreator
+	return c
 }
 
 // Create creates a new hook with the given name and arguments.
@@ -183,7 +189,11 @@ func (c cdiHookCreator) Create(name HookName, args ...string) *Hook {
 }
 
 func (c cdiHookCreator) isDisabled(name HookName, args ...string) bool {
-	if c.disabledHooks[name] {
+	disabled, ok := c.disabledHooks[name]
+	if ok {
+		return disabled
+	}
+	if c.disabledHooks[AllHooks] {
 		return true
 	}
 
```
It would also be good to add a unit test to test `cdiHookCreator.isDisabled` behaviour.

ArangoGutierrez: Unit test added

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,193 @@
+package discover
```
elezar, line 17: nit: license header

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,193 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			creator := &cdiHookCreator{
+				disabledHooks: tt.disabledHooks,
+			}
+
+			// Apply enabled hooks override
+			for _, enabledHook := range tt.enabledHooks {
+				creator.disabledHooks[enabledHook] = false
+			}
```
elezar: We should be calling `New()` to test the actual constructor. Here we are just testing that we disable the hooks in the test.

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+			result := creator.isDisabled(tt.hookName, tt.args...)
+			if result != tt.expectedResult {
+				t.Errorf("isDisabled() = %v, want %v", result, tt.expectedResult)
+			}
```
elezar: We use the `require` package for our tests in this project.

elezar:
One note. Users that REQUIRE this hook don't have the option to opt-in to it when using the runtime -- only CDI. Maybe that's sufficient though.

internal/discover/hooks_test.go
ArangoGutierrez: Done

internal/discover/hooks_test.go
ArangoGutierrez: done

internal/discover/hooks_test.go
ArangoGutierrez: done

ArangoGutierrez:
@elezar https://github.com/NVIDIA/nvidia-container-toolkit/pull/1221/commits/3ffdc7942f4e705a2d50f1bbd63a2048ac742c3d Unit tests as discussed offline

cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -214,6 +215,13 @@ func (m command) build() *cli.Command {
 				Destination: &opts.disabledHooks,
 				Sources:     cli.EnvVars("NVIDIA_CTK_CDI_GENERATE_DISABLED_HOOKS"),
 			},
+			&cli.StringSliceFlag{
+				Name:        "enable-hook",
+				Aliases:     []string{"enable-hooks"},
+				Usage:       "Explicitly enable a hook in the generated CDI specification. This overrides disabled hooks. This can be specified multiple times.",
```
elezar, line 221: One thing that we didn't talk about was how we should handle `enable-hooks=all`? The way I understand the implementation this will not have any effect. I would say that `disable-hooks=all` is valid, but that `enable-hooks=all` could be considered an invalid input.

ArangoGutierrez: ack, question, should we error if `enable-hooks=all` is used, or simply warn the user?

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,543 @@
...
+				WithEnabledHooks(ChmodHook),
+			},
+			expectedType: "cdiHookCreator",
+			expectedPath: defaultNvidiaCDIHookPath,
+			expectedDisabled: map[HookName]bool{
+				ChmodHook: false, // ChmodHook is enabled
+			},
+		},
+		{
+			name: "nvidia-ctk binary path affects fixed args",
```
elezar, line 128: Se don't seem to check the value of fixed args here.

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,543 @@
...
+		{
+			name: "nvidia-ctk in /usr/sbin (RPM-based systems)",
+			opts: []Option{
+				WithNVIDIACDIHookPath("/usr/sbin/nvidia-ctk"),
+			},
+			expectedType: "cdiHookCreator",
+			expectedPath: "/usr/sbin/nvidia-ctk",
+			expectedDisabled: map[HookName]bool{
+				ChmodHook: true,
+			},
```
elezar: While I see the point of this test in principle there is nothing in the constructor that depends on the `bin` vs `sbin` on various platforms. From the point of view of the contructor the path is just a string, and only the BASENAME of the executable affects the behaviour.

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,543 @@
...
+		},
+		{
+			name: "nvidia-ctk in custom NVIDIA toolkit path",
+			opts: []Option{
+				WithNVIDIACDIHookPath("/usr/local/nvidia/toolkit/nvidia-ctk"),
```
elezar, line 143: Same here.

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,543 @@
...
+			case "allDisabledHookCreator":
+				_, ok := hookCreator.(*allDisabledHookCreator)
+				require.True(t, ok, "Expected allDisabledHookCreator")
+			case "cdiHookCreator":
+				creator, ok := hookCreator.(*cdiHookCreator)
+				require.True(t, ok, "Expected cdiHookCreator")
+				require.Equal(t, tc.expectedPath, creator.nvidiaCDIHookPath)
+				for hook, disabled := range tc.expectedDisabled {
+					require.Equal(t, disabled, creator.disabledHooks[hook], "Hook %s disabled state mismatch", hook)
+				}
```
elezar: Can one use `require.ElementsMatch` or something similar here?

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,543 @@
...
+)
+
+func TestNewHookCreator(t *testing.T) {
+	testCases := []struct {
+		name             string
+		opts             []Option
+		expectedType     string // "cdiHookCreator" or "allDisabledHookCreator"
+		expectedPath     string
+		expectedDisabled map[HookName]bool
+	}{
```
elezar: I have pushed a commit with a proposal for testing the constructor. We can define `testCases` as:
```
testCases := []struct {
		name     string
		opts     []Option
		expected HookCreator
	}{
```
and clean up the implementation significantly.

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,534 @@
...
+				disabledHooks[h] = false
+			}
+
+			testCreator := &cdiHookCreator{
+				nvidiaCDIHookPath: o.nvidiaCDIHookPath,
+				disabledHooks:     disabledHooks,
+				fixedArgs:         getFixedArgsForCDIHookCLI(o.nvidiaCDIHookPath),
+			}
+
+			require.Equal(t, testCreator.isDisabled(tt.hookName, tt.args...), tt.expectedResult)
```
elezar: Why are we reimplementing the contructor here? What exactly are we testing?

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			hookCreator := NewHookCreator(tc.opts...)
+
+			require.EqualValues(t, tc.expected, hookCreator)
+		})
+	}
+}
+func TestCDIHookCreator_Create(t *testing.T) {
```
elezar, line 163: PLease see my proposal for improving this test.

cmd/nvidia-ctk/cdi/generate/generate.go
ArangoGutierrez: answered by https://github.com/NVIDIA/nvidia-container-toolkit/pull/1221/commits/57b103b914ad1000f8957f31c383598f4f43bd2b

internal/discover/hooks_test.go
ArangoGutierrez: The rationale behind the way this test is built is that we want to test the CDIHookCreator.isDisabled directly without testing the NewHookCreator func. 

I'll simplify this. I realized that we can trim down to focus on creating the cdiHookCreator

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,490 @@
...
+		hookCreator  HookCreator
+		hookName     HookName
+		args         []string
+		expectedHook *Hook
+	}{
+		{
+			name:        "CreateSymlinksHook with args",
+			hookCreator: NewHookCreator(WithNVIDIACDIHookPath(defaultNvidiaCDIHookPath)),
+			hookName:    CreateSymlinksHook,
+			args:        []string{"/source:/target", "/source2:/target2"},
```
elezar: nit: The format for the link is `{{ .Source }}::{{ .Target }}`: https://github.com/NVIDIA/nvidia-container-toolkit/blob/9d2ba796a0359367efb5acbf928d4da7dc969eaf/internal/discover/symlinks.go#L35

```suggestion
			args:        []string{"/source::/target", "/source2::/target2"},
```

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,490 @@
...
+			expectedHook: &Hook{
+				Lifecycle: "createContainer",
+				Path:      "/usr/bin/nvidia-cdi-hook",
+				Args:      []string{"nvidia-cdi-hook", "create-symlinks", "--link", "/source:/target", "--link", "/source2:/target2"},
```
elezar:
```suggestion
				Args:      []string{"nvidia-cdi-hook", "create-symlinks", "--link", "/source::/target", "--link", "/source2::/target2"},
```

ArangoGutierrez: fixed

internal/discover/hooks_test.go
ArangoGutierrez: fixed

internal/discover/hooks_test.go
ArangoGutierrez: discussed and added

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,490 @@
...
+				fixedArgs:         []string{"nvidia-cdi-hook"},
+				disabledHooks: map[HookName]bool{
+					AllHooks:          true,
+					UpdateLDCacheHook: false,
+					ChmodHook:         true,
+				},
+			},
+		},
+		{
+			name: "multiple hooks disabled and enabled",
```
elezar, line 79: nit: This doesn't test multiple enabled hooks.

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,490 @@
...
+				fixedArgs:         []string{"nvidia-cdi-hook"},
+				disabledHooks: map[HookName]bool{
+					UpdateLDCacheHook:  true,
+					CreateSymlinksHook: true,
+					ChmodHook:          false, // Explicitly enabled
+				},
+			},
+		},
+		{
+			name: "WithDisabledHooks can be called multiple times",
```
elezar, line 97: Do we need a similar test for `WithEnabledHooks`?

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,490 @@
...
+				fixedArgs:         []string{"nvidia-ctk", "hook"},
+				disabledHooks: map[HookName]bool{
+					ChmodHook: true,
+				},
+			},
+		},
+		{
+			name: "nvidia-ctk in custom NVIDIA toolkit path",
+			opts: []Option{
+				WithNVIDIACDIHookPath("/usr/local/nvidia/toolkit/nvidia-ctk"),
```
elezar, line 143: Nit: Let's use the same `custom` path as for the `nvidia-cdi-hook` tests.

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,490 @@
...
+				WithNVIDIACDIHookPath("/usr/local/bin/nvidia-cdi-hook"),
+			},
+			expected: &cdiHookCreator{
+				nvidiaCDIHookPath: "/usr/local/bin/nvidia-cdi-hook",
+				fixedArgs:         []string{"nvidia-cdi-hook"},
+				disabledHooks: map[HookName]bool{
+					ChmodHook: true,
+				},
+			},
+		},
```
elezar: This test doesn't test anything that isn't already covered by the `custom` path test already added.

internal/discover/hooks_test.go
```diff
@@ -0,0 +1,490 @@
...
+				disabledHooks[h] = false
+			}
+
+			testCreator := &cdiHookCreator{
+				nvidiaCDIHookPath: defaultNvidiaCDIHookPath,
+				disabledHooks:     disabledHooks,
+				fixedArgs:         getFixedArgsForCDIHookCLI(defaultNvidiaCDIHookPath),
+			}
+
+			require.Equal(t, testCreator.isDisabled(tt.hookName, tt.args...), tt.expectedResult)
```
elezar: I already proposed this change once:
```suggestion
			testCreator := NewHookCreator(
				WithDisabledHooks(tt.disabledHooks...),
				WithEnabledHooks(tt.enabledHooks...),
			)
			require.Equal(t, testCreator.(*cdiHookCreator).isDisabled(tt.hookName, tt.args...), tt.expectedResult)
		})
```
If you left it this way to test the "AllDisabled" cases, then let's rather add those to the `Create` tests above:
```
+               {
+                       name:         "allDisabled with unknown hook",
+                       hookCreator:  &allDisabledHookCreator{},
+                       hookName:     HookName("unknown-hook"),
+                       args:         []string{},
+                       expectedHook: nil,
+               },
+               {
+                       name:         "allDisabled with defined",
+                       hookCreator:  &allDisabledHookCreator{},
+                       hookName:     UpdateLDCacheHook,
+                       args:         []string{},
+                       expectedHook: nil,
+               },
```

internal/discover/hooks_test.go
elezar: As a note. If we wanted to test `isDisabled` directly then just have the `testCases` struct define a `cdiHookCreator` member directly. We can set that the way we want it WITHOUT having to reimplement the constructor.

internal/discover/hooks_test.go
ArangoGutierrez: Not really, as we are calling the `NewHookCreator` during the `create` test

internal/discover/hooks_test.go
ArangoGutierrez: ack

internal/discover/hooks_test.go
ArangoGutierrez: ack

internal/discover/hooks_test.go
ArangoGutierrez: ack

internal/discover/hooks_test.go
ArangoGutierrez: I acknowledge you proposed the change, but I ended up deciding for the current implementation, as the `testCreator` during the `AllDisabled` returns a `allDisabledHookCreator` which doesn't have an `isDisabled` method. 

This is the main reason I decided on the current implementation. 

With that said, let me add a change that might work.

ArangoGutierrez: Ok, I adopted your suggestion and added the AllHooks test to the `Create` section

internal/discover/hooks.go
elezar: I was too hasty in suggesting this. As discussed offline, I think a strongly-typed internal API is fine since we only every call this with `HookName` arguments. If we need more flexibility in the future we can add generics.

ArangoGutierrez:
@elezar commits squashed

PR #1229, opened 2025-08-07
[no-relnote] Auto-regenerate CDI spec on toolkit changes

ArangoGutierrez:
Monitor /usr/bin/nvidia-ctk for changes and trigger CDI refresh during package installation to ensure CDI specifications stay current with both driver and toolkit updates.

- Add nvidia-ctk binary monitoring to systemd path unit
- Trigger CDI refresh in DEB/RPM post-install scripts
- Reorganize RPM spec to handle base package services correctly


elezar (approved):
LGTM

PR #1235, opened 2025-08-08
[no-relnote] implement a nestedContainerRunner for E2E test suite

ArangoGutierrez:


tests/e2e/runner.go
```diff
@@ -21,11 +21,69 @@ import (
...
 )
 
+const (
+	startTestContainerTemplate = `docker run -d --name {{.ContainerName}} --privileged --runtime=nvidia \
+    -e NVIDIA_VISIBLE_DEVICES=runtime.nvidia.com/gpu=all \
+    -e NVIDIA_DRIVER_CAPABILITIES=all \
+	{{ range $i, $a := .AdditionalArguments -}}
+	{{ $a }} \
+	{{ end -}}
+	ubuntu sleep infinity`
```
elezar: Is this not the only thing that is required for the nested container runner? The other scripts that have been moved here are test-specific. (The one exception is the additional arguments required to mount the host installed libraries into the container that is started).

ArangoGutierrez: Not really.
The template `installDockerTemplate` is needed to install Docker inside Docker to enable nested runs. 
The template `installCTKTemplate` with the `installCTK` when calling `NewNestedContainerRunner` will allow us to create `nestedRunners` with or without the toolkit preinstalled inside the nested environment. 

That was my reasoning to move all the templates to the `runner.go`, leaving only `libnvidiaContainerCliTestTemplate` on the `nvidia-container-cli_test.go` file, as this is indeed test specific template

elezar: When we add logic to install `podman` in the container, would that also be implemented in the `runner.go`. Also, looking forward, how do we reuse the nested runner to also run other toolkit tests?

elezar:
As a general note. Our motiviation for adding this runner is to be able to test the generation of CDI specs by the systemd unit. Does it make sense to add that to this test as well so that the required functionality drives the types of changes we make?

tests/e2e/runner.go
ArangoGutierrez: >When we add logic to install podman in the container, would that also be implemented in the runner.go 

We could modify the `newNestedRunner` to select the `nested` runtime.

>looking forward, how do we reuse the nested runner to also run other toolkit tests?

You mean rewriting the current tests with the nestedRunner?

ArangoGutierrez:
> As a general note. Our motiviation for adding this runner is to be able to test the generation of CDI specs by the systemd unit. Does it make sense to add that to this test as well so that the required functionality drives the types of changes we make?

OK, I just wanted a small, focused PR, but I am ok with expanding the scope of the PR. I'll work on it

tests/e2e/runner.go
elezar: I don't mean that we should rewrite the current tests with the `nestedRunner`, but that the current tests should be able to run against a `nestedRunner` already.

As a question: from the perspective of the existing toolkit tests, would we expect a difference if these are running directly on an instance (i.e. the current runner), or in a containerized OS (the nested runner)? For all tests what we do is:

* Start a test environment (an AWS instance, an outer container) with the NVIDIA Driver available
* Install the prerequsites (currently Docker, but could be extended to Podman)

At this point we currently have a differences between the two classes of tests. For one of these we use toolkit container to install the toolkit to the host. For the second, we copy the packages out of the container and installs these on the "host" (actually the outer container). Note that for the systemd tests, this is what we would be "testing".

elezar:
As discussed offline: The intent of the "nested" Runner is to allow us to test other environments -- such as different target operating systems -- using our existin tests. With this in mind, we need to rework how we're constructing our "runner" and move it closer to being infrastructure instead of being so closely tied to each test.

Ideally the same tests should be able to be run against:
* The version of the toolkit already installed on the system
* The version under test in CI installed on an AWS instance
* The version under test installed in an "outer" container with a specific distribution

Note that if the last of these can be made reliable enough, then the middle one (i.e. installing directly on an AWS instance) should not even be required.

tests/e2e/nvidia-cdi-refresh_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/kind/pkg/build/nodeimage"
+)
+
+const (
+	nvidiaCdiRefreshServiceTestTemplate = `
```
elezar: Question: Is there an advantage to having this as a single test script instead of separate `It` contexts?

ArangoGutierrez: You are right, I have divided it into self-contained `IT` contexts, now it runs as follows:

```bash
nvidia-cdi-refresh when installing nvidia-container-toolkit should load the nvidia-cdi-refresh.path and nvidia-cdi-refresh.service units [systemd-unit]
/Users/eduardoa/src/github/nvidia/nvidia-container-toolkit/tests/e2e/nvidia-cdi-refresh_test.go:105
• [60.608 seconds]
------------------------------
nvidia-cdi-refresh when installing nvidia-container-toolkit should generate the nvidia.yaml file [systemd-unit]
/Users/eduardoa/src/github/nvidia/nvidia-container-toolkit/tests/e2e/nvidia-cdi-refresh_test.go:110
• [2.388 seconds]
------------------------------
nvidia-cdi-refresh when installing nvidia-container-toolkit should refresh the nvidia.yaml file after upgrading the nvidia-container-toolkit [systemd-unit]
/Users/eduardoa/src/github/nvidia/nvidia-container-toolkit/tests/e2e/nvidia-cdi-refresh_test.go:115
• [8.038 seconds]
```

packaging/debian/nvidia-container-toolkit-base.postinst
```diff
@@ -7,16 +7,20 @@ case "$1" in
...
-          systemctl enable --now nvidia-cdi-refresh.service || echo "Warning: Failed to enable nvidia-cdi-refresh.service" >&2
-          
-          # Trigger CDI spec regeneration immediately after install/upgrade
-          echo "Regenerating NVIDIA CDI specification..."
-          systemctl start nvidia-cdi-refresh.service || echo "Warning: Failed to trigger CDI refresh" >&2
+        # Support running, degraded, and maintenance systemd states
+        if command -v systemctl >/dev/null 2>&1; then
+          SYSTEMD_STATE=$(systemctl is-system-running 2>/dev/null || true)
+          case "$SYSTEMD_STATE" in
+            running|degraded|maintenance)
```
elezar: Question: What does `maintenance` mean in this context. SHOULD we be generating specs in this state?

packaging/rpm/SPECS/nvidia-container-toolkit.spec
elezar, line 1: Let's create a PR with just these changes so that we can cut an `rc.4` including them.

tests/e2e/nvidia-cdi-refresh_test.go
```diff
@@ -0,0 +1,169 @@
...
+[Unit]
+Description=Dummy systemd service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/sh -c "exit 0"
+EOF
+
+	# We know the dummy service will fail, so we can ignore the error
+	systemctl start dummy.service 2>/dev/null || true
```
elezar, line 62: Should this not only be a single configuration for the test case:
"The NVIDIA Container Toolkit installs and generates a CDI spec if the system is in a degraded state"?

tests/e2e/nvidia-cdi-refresh_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+const (
+	nvidiaCdiRefreshInstallTemplate = `
+	# Read the TMPDIR
+	TMPDIR=$(cat /tmp/ctk_e2e_temp_dir.txt)
+	export TMPDIR
+
+	# uninstall the nvidia-container-toolkit
+	apt-get remove -y nvidia-container-toolkit nvidia-container-toolkit-base libnvidia-container-tools libnvidia-container1
+	apt-get autoremove -y
```
elezar, line 37: Why is this not in an "AfterEach"?

tests/e2e/nvidia-cdi-refresh_test.go
```diff
@@ -0,0 +1,169 @@
...
+	# We know the dummy service will fail, so we can ignore the error
+	systemctl start dummy.service 2>/dev/null || true
+
+	# Install the nvidia-container-toolkit
+	dpkg -i ${TMPDIR}/packages/ubuntu18.04/amd64/libnvidia-container1_*_amd64.deb ${TMPDIR}/packages/ubuntu18.04/amd64/nvidia-container-toolkit-base_*_amd64.deb ${TMPDIR}/packages/ubuntu18.04/amd64/libnvidia-container-tools_*_amd64.deb
+
+	if ! systemctl status nvidia-cdi-refresh.path | grep "Active: active"; then
+		echo "nvidia-cdi-refresh.path is not Active"
+		exit 1
+	fi
```
elezar, line 71: Should this be an `It` clause?

tests/e2e/nvidia-cdi-refresh_test.go
```diff
@@ -0,0 +1,169 @@
...
+	if ! systemctl status nvidia-cdi-refresh.service | grep "Loaded: loaded"; then
+		echo "nvidia-cdi-refresh.service is not loaded"
+		exit 1
+	fi
```
elezar, line 77: Should this be an `It` clause?

tests/e2e/nvidia-cdi-refresh_test.go
```diff
@@ -0,0 +1,169 @@
...
+	`
+
+	nvidiaCdiRefreshUpgradeTemplate = `
+	# remove the generated files
+	rm /var/run/cdi/nvidia.yaml /tmp/nvidia.yaml
+
+	# Touch the nvidia-ctk binary to change the mtime
+	# This will trigger the nvidia-cdi-refresh.path unit to call the
+	# nvidia-cdi-refresh.service unit, simulating a change(update/downgrade) in the nvidia-ctk binary.
+	touch $(which nvidia-ctk)
```
elezar, line 104: We could also install an older version of the toolkit and test upgrades then?

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -360,7 +343,14 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
 
-			tmpDirPath = GinkgoT().TempDir()
+			// Make test runable from a MacOs hosts.
+			// On darwin, the temp dir is in /var/folders/.../T/
+			// We need to convert it to /tmp/...
+			if runtime.GOOS == "darwin" {
+				tmpDirPath = path.Join("/tmp", uuid.NewString())
+			} else {
+				tmpDirPath = GinkgoT().TempDir()
+			}
```
elezar: This is the wrong fix for this. What this means is that we probably want to run `mktemp -d` below instead.

tests/e2e/runner.go
```diff
@@ -79,6 +155,150 @@ func NewRunner(opts ...runnerOption) Runner {
...
+					additionalContainerArguments = append(additionalContainerArguments,
+						fmt.Sprintf("-v %s:%s", binaryPath, binaryPath))
+					break // Move to the next binary once found
+				}
+			}
+		}
+	}
+
+	// Mount the /lib/modules directory as a volume to enable the nvidia-cdi-refresh service
+	additionalContainerArguments = append(additionalContainerArguments, "-v /lib/modules:/lib/modules")
```
elezar, line 238: Quesiton: Why is this required?

tests/e2e/nvidia-cdi-refresh_test.go
```diff
@@ -0,0 +1,169 @@
...
+		echo "nvidia.yaml file is different"
+		exit 1
+	fi
+	`
+)
+
+var _ = Describe("nvidia-cdi-refresh", Ordered, ContinueOnFailure, Label("systemd-unit"), func() {
+	var (
+		nestedContainerRunner Runner
+		outerContainerImage   = nodeimage.DefaultBaseImage
```
elezar: Although I understand the intent of this, I feel as if importing the `nodeimage` package and all its transitive dependencies JUST for this constant is not ideal. Let's first just hardcode the value here and then as a follow-up consider alternatives.

One option would be to use dependabot, or to try to get this by checking the dockerhub registry, for example.

packaging/debian/nvidia-container-toolkit-base.postinst
ArangoGutierrez: From docs:

> System Maintenance Mode is a temporary status for a computer or software system where user access is restricted or blocked to allow administrators to perform essential updates, backups, or other maintenance tasks safely. During this mode, the system often displays a message informing users that it is unavailable and will be back online once the maintenance is complete.

So I think a user should not be installing things during a `maintenance` period

packaging/rpm/SPECS/nvidia-container-toolkit.spec
ArangoGutierrez: https://github.com/NVIDIA/nvidia-container-toolkit/pull/1273

packaging/debian/nvidia-container-toolkit-base.postinst
ArangoGutierrez: https://github.com/NVIDIA/nvidia-container-toolkit/pull/1273 drops maintenance

tests/e2e/nvidia-cdi-refresh_test.go
ArangoGutierrez: In this script, I am uninstalling the toolkit on purpose to then get `systemd` on a degraded state and then re-install the toolkit to test that we support the degraded state. Given that this test is running on a `nestedContainer` the `AfterAll` removes the `outher` container

tests/e2e/nvidia-cdi-refresh_test.go
ArangoGutierrez: Yes, you are totally right, I have a note for this, my idea is to wait for `v1.18.0` and then pin that as the `old` version. So for the scope of this PR `touch` is ok enough to fake/mock a new `nvidia-ctk` binary

tests/e2e/nvidia-container-toolkit_test.go
ArangoGutierrez: The issue here is that on Darwin `GinkgoT().TempDir()` returns 

```bash
❯ env |grep TMP
TMPDIR=/var/folders/nk/0fkkygw917n7797_jkxbdcjc0000gp/T/
```
resulting in `mktemp -d /var/folders/nk/0fkkygw917n7797_jkxbdcjc0000gp/T/`

tests/e2e/runner.go
ArangoGutierrez: On `nvidia-cdi-refresh.service`, we have `ExecCondition=/usr/bin/grep -qE '/nvidia.ko' /lib/modules/%v/modules.dep`, meaning that if that file doesn't exist, when the `nvidia-cdi-refresh.path` calls the `nvidia-cdi-refresh.service`, it will simply log the call and skip execution

tests/e2e/nvidia-cdi-refresh_test.go
ArangoGutierrez: ack, will hard code and note a TODO

ArangoGutierrez:
Depends on https://github.com/NVIDIA/nvidia-container-toolkit/pull/1273

tests/e2e/nvidia-cdi-refresh_test.go
ArangoGutierrez: Agree, I am now moving this to a stand alone template, and calling it from it's own `It` block

tests/e2e/nvidia-cdi-refresh_test.go
ArangoGutierrez: Agree, I am now moving this to a stand alone template, and calling it from it's own `It` block

tests/e2e/nvidia-cdi-refresh_test.go
ArangoGutierrez: I have turned this script just to create a degraded state and run it after testing on a healthy system, this way we test the 2 status we support `running|degraded`

ArangoGutierrez:
Closing in favor of https://github.com/NVIDIA/nvidia-container-toolkit/pull/1348

Issue #1222, opened 2025-08-05
Container Toolkit messed up config.toml in GKE cluster with containerd 2.0


quoctruong:
I've been using the GPU Operator to install Nvidia driver in my GKE cluster. This has worked fine until last week when the cluster version is upgraded from GKE 1.32 to 1.33. One of the big change is that the containerd version is now 2.0. Now, a lot of GPU operator Daemonset is failing with the following errors:
```
"failed to \"CreatePodSandbox\" for \"gpu-feature-discovery-nv7f9_gpu-operator(b07a2405-29b1-412e-8400-a1367a10e76d)\" with CreatePodSandboxError: \"Failed to create sandbox for pod \\\"gpu-feature-discovery-nv7f9_gpu-operator(b07a2405-29b1-412e-8400-a1367a10e76d)\\\": rpc error: code = Unknown desc = failed to setup network for sandbox \\\"2b62cb2980cd477cefbea1155871bcd2f50b2c6fb3e922039554d8b3fc14784f\\\": plugin type=\\\"gke\\\" failed (add): failed to find plugin \\\"gke\\\" in path [/opt/cni/bin]\"
```

So we went and looked at the `config.toml` file and it looks like it was modified and added an incorrect line of config:
```
    [plugins."io.containerd.cri.v1.runtime".cni]
      bin_dir = "/opt/cni/bin"
      conf_dir = "/etc/cni/net.d"
```

This is most likely because of the `nvidia-container-toolkit-daemonset` because it was the last pod to be able start successfully (and also it looks like the toolkit modifies this file?).

The toolkit version is from the container `nvcr.io/nvidia/k8s/container-toolkit:v1.17.8-ubuntu20.04` and the GPU operator is version 25.3.2 (the latest one).

Any insight on what is happening and why the `config.toml` was modified incorrectly for containerd 2.0?

carreter:
Seems like the root cause of this is https://github.com/containerd/containerd/issues/11747

`nvidia-container-toolkit` uses `containerd config dump` to get the containerd config by default, even if you specify the file path for the `config.toml`. This command only partially migrates the output from v2 to v3 and messes things up in the progress.

Easy fixes would be to either have `nvidia-container-toolkit` prefer reading config from file over `containerd config dump`, or using `containerd config migrate` instead which properly does the v2 -> v3 migration.

klueska:
[...]

PR #1251, opened 2025-08-19
Allow config sources to be specified for containerd and cri-o

elezar:
By default container extracts the current config by running the `containerd config dump` command and if that fails we read the existing config file.

This change allows ordering of these sources to be defined as we do for the nvidia-ctk runtime configure command.

See #1222 where setting `RUNTIME_CONFIG_SOURCES=file` or `RUNTIME_CONFIG_SOURCES=file,command` should allow the current config to be preferred over the `containerd config dump` command.

This further allows an explicit path to be specified for the `file` config source so that there is a separation between the config file being modified by the tooling and the config used as the source for drop-in files, for example.

### DIFF
--- a/cmd/nvidia-ctk-installer/container/runtime/containerd/containerd.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/containerd/containerd.go
@@ -172,12 +172,15 @@ func GetLowlevelRuntimePaths(o *container.Options, co *Options) ([]string, error
 }
 
 func getRuntimeConfig(o *container.Options, co *Options) (engine.Interface, error) {
+	loaders, err := o.GetConfigLoaders(containerd.CommandLineSource)
+	if err != nil {
+		return nil, err
+	}
 	options := []containerd.Option{
 		containerd.WithTopLevelConfigPath(o.TopLevelConfigPath),
 		containerd.WithConfigSource(
 			toml.LoadFirst(
-				containerd.CommandLineSource(o.HostRootMount, o.ExecutablePath),
-				toml.FromFile(o.TopLevelConfigPath),
+				loaders...,
 			),
 		),
 		containerd.WithRuntimeType(co.runtimeType),
--- a/cmd/nvidia-ctk-installer/container/runtime/runtime.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/runtime.go
@@ -118,9 +118,12 @@ func Flags(opts *Options) []cli.Flag {
 			Hidden:      true,
 		},
 		&cli.StringSliceFlag{
-			Name:        "config-source",
-			Aliases:     []string{"config-sources"},
-			Usage:       "Specify the config sources for the container runtime. Any combination of [command | file].",
+			Name:    "config-source",
+			Aliases: []string{"config-sources"},
+			Usage: "Specify the config sources for the container runtime. Any combination of " +
+				"[command | file]. " +
+				"When `file` is specified, the absolute path to the file to be used as a config source can " +
+				"be specified as `file=/path/to/source/config.toml",
 			Value:       []string{"command", "file"},
 			Destination: &opts.ConfigSources,
 			Sources:     cli.EnvVars("RUNTIME_CONFIG_SOURCES", "RUNTIME_CONFIG_SOURCE"),
--- a/cmd/nvidia-ctk-installer/container/runtime/runtime.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/runtime.go
@@ -152,7 +152,6 @@ func (opts *Options) Validate(logger logger.Interface, c *cli.Command, runtime s
 	}
 
 	// Apply the runtime-specific config changes.
-	// TODO: Add the runtime-specific DropInConfigs here.
 	switch runtime {
 	case containerd.Name:
 		if opts.TopLevelConfigPath == runtimeSpecificDefault {
--- a/cmd/nvidia-ctk-installer/container/runtime/runtime.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/runtime.go
@@ -117,6 +117,14 @@ func Flags(opts *Options) []cli.Flag {
 			Sources:     cli.EnvVars("NVIDIA_RUNTIME_SET_AS_DEFAULT", "CONTAINERD_SET_AS_DEFAULT", "DOCKER_SET_AS_DEFAULT"),
 			Hidden:      true,
 		},
+		&cli.StringSliceFlag{
+			Name:        "config-source",
+			Aliases:     []string{"config-sources"},
+			Usage:       "Specify the config sources for the container runtime. Any combination of [command | file].",
+			Value:       []string{"command", "file"},
+			Destination: &opts.ConfigSources,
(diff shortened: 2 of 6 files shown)


cmd/nvidia-ctk-installer/container/runtime/containerd/containerd.go
```diff
@@ -169,12 +169,23 @@ func GetLowlevelRuntimePaths(o *container.Options, co *Options) ([]string, error
 }
 
 func getRuntimeConfig(o *container.Options, co *Options) (engine.Interface, error) {
+	var loaders []toml.Loader
+	for _, source := range o.ConfigSources {
+		switch source {
+		case "file":
+			loaders = append(loaders, toml.FromFile(o.Config))
+		case "command":
```
elezar: Note to self: Since `command` here implies `containerd config dump` command we may want to make this more expressive. For now I have used the values that we support in the `nvidia-ctk runtime configure` command: https://github.com/NVIDIA/nvidia-container-toolkit/blob/7b1b20ba8ccebc4a8921f567ca88456785aee2e4/cmd/nvidia-ctk/runtime/configure/configure.go#L132-L137

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -103,6 +103,12 @@ func Flags(opts *Options) []cli.Flag {
 			Sources:     cli.EnvVars("NVIDIA_RUNTIME_SET_AS_DEFAULT", "CONTAINERD_SET_AS_DEFAULT", "DOCKER_SET_AS_DEFAULT"),
 			Hidden:      true,
 		},
+		&cli.StringSliceFlag{
+			Name:        "config-source",
+			Usage:       "specify the config sources",
+			Destination: &opts.ConfigSources,
+			Sources:     cli.EnvVars("RUNTIME_CONFIG_SOURCES"),
```
ArangoGutierrez: Flag to be called `config-source-order` (or similar)

`specify the order of config sources`

As the user can do `--config-source command,file` or `--config-source file,command`

same for the ENV VAR `RUNTIME_CONFIG_SOURCES_ORDER`

elezar: I don't agree that the flag name should be changed. The flag specifies the sources and the ordering is implied by the order in which the sources are specified. Valid values for this argument are currently:

* `""` 
* `"file"`
* `"command"`
* `"command,file"` (Default)
* `"file,command"`

ArangoGutierrez: Then let's expand the `Usage` to be more explicit, `"specify the config sources [file, command]"` as `nvidia-ctk runtime configure` has a flag with the same name but this one is to point to a path. We should provide guidance to the user via `Usage` with at least the valid options or the default behav of this flag if not set.

ArangoGutierrez:
Can we mark this PR as ready for review?

elezar:
> Can we mark this PR as ready for review?

I think we can move this to the next minor (if we decide to implement it at all). See the comment here https://github.com/NVIDIA/nvidia-container-toolkit/issues/1222#issuecomment-3301563236 for a workaround in the existing codebase.

cdesiniotis:
I am testing this with the operator and microk8s, by setting the following with the toolkit container:
```
env:
    - name: CONTAINERD_CONFIG
      value: /var/snap/microk8s/current/args/containerd-template.toml
    - name: CONTAINERD_SOCKET
      value: /var/snap/microk8s/common/run/containerd.sock
    - name: RUNTIME_CONFIG_SOURCE
      value: file=/var/snap/microk8s/current/args/containerd.toml
```

This does not work as is since the path provided to `RUNTIME_CONFIG_SOURCE` is the path to the file on the host and NOT in the container.

elezar:
> I am testing this with the operator and microk8s, by setting the following with the toolkit container:
> 
> ```
> env:
>     - name: CONTAINERD_CONFIG
>       value: /var/snap/microk8s/current/args/containerd-template.toml
>     - name: CONTAINERD_SOCKET
>       value: /var/snap/microk8s/common/run/containerd.sock
>     - name: RUNTIME_CONFIG_SOURCE
>       value: file=/var/snap/microk8s/current/args/containerd.toml
> ```
> 
> This does not work as is since the path provided to `RUNTIME_CONFIG_SOURCE` is the path to the file on the host and NOT in the container.

Yes, I missed this when trying to think this through. The user _could_ provide the container path here, but that's definitely not ideal.

elezar:
@cdesiniotis what if we also add the following to the GPU Operator:
```
	if runtimeConfigSources := getContainerEnv(container, "RUNTIME_CONFIG_SOURCE"); runtimeConfigSources != "" {
		var sources []string
		for _, runtimeConfigSource := range strings.Split(runtimeConfigSources, ",") {
			parts := strings.SplitN(runtimeConfigSource, "=", 2)
			if len(parts) == 1 || parts[0] != "file" {
				sources = append(sources, runtimeConfigSource)
				continue
			}
			sourceConfigFile := parts[1]
			containerConfigFile := sourceConfigFile
			if sourceConfigFile != "" {
				sourceConfigFileName := path.Base(sourceConfigFile)
				sourceConfigDir := path.Dir(sourceConfigFile)

				// TODO: If the source config file is not in the same folder as
				// the top-level config file, then we need to use a different
				// container path here.
				containerConfigDir := DefaultRuntimeConfigTargetDir
				containerConfigFile = containerConfigDir + sourceConfigFileName

				// If we've already mounted sourceConfigDir -> containerConfigDir
				// then we don't create a new volume mount.
				if hostToContainerDirs[sourceConfigDir] != "" {
					volMountConfigName := fmt.Sprintf("%s-source-config", runtime)
					volMountConfig := corev1.VolumeMount{Name: volMountConfigName, MountPath: containerConfigDir}
					container.VolumeMounts = append(container.VolumeMounts, volMountConfig)

					configVol := corev1.Volume{Name: volMountConfigName, VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: sourceConfigDir, Type: newHostPathType(corev1.HostPathDirectoryOrCreate)}}}
					obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, configVol)
				}
			}
			sources = append(sources, "file="+containerConfigFile)
		}
		setContainerEnv(container, "RUNTIME_CONFIG_SOURCE", strings.Join(sources, ","))
	}
```
(We can definitely factor out some logic here to make things more readable across the multiple instances where we are transforming host paths to container paths).

This would mean that a user would specify `file=/path/on/the/host/config.toml` as the source and the path will be updated in the same way as other paths are treated.

cdesiniotis:
A simpler solution would be to prepend `/host` to the file path -- as the toolkit container already has read-only access to the host's rootfs at `/host`. This would avoid us potentially having to add an additional volume. I have raised https://github.com/NVIDIA/gpu-operator/pull/1758 with this proposed change.

cdesiniotis (approved):

cdesiniotis:
I've tested this PR paired with https://github.com/NVIDIA/gpu-operator/pull/1758 and it provides a working solution for microk8s.

PR #1252, opened 2025-08-19
Allow config sources to be specified for containerd

elezar:
By default container extracts the current config by running the `containerd config dump` command and if that fails we read the existing config file.

This change allows ordering of these sources to be defined as we do for the nvidia-ctk runtime configure command.

Backport of #1251
See #1222

### DIFF
--- a/tools/container/container.go
+++ b/tools/container/container.go
@@ -50,6 +50,8 @@ type Options struct {
 	SetAsDefault  bool
 	RestartMode   string
 	HostRootMount string
+
+	ConfigSources cli.StringSlice
 }
 
 // ParseArgs parses the command line arguments to the CLI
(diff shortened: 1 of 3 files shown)


carreter:
Not a maintainer but LGTM. I think this is an adequate fix for https://github.com/NVIDIA/nvidia-container-toolkit/issues/1222 :)

elezar:
Closing this backport since a workaround exists that does not require any changes.

carreter:
Hey at @elezar , just saw that you closed this. Is there a plan to include this workaround or another more permanent fix in the next release? Also, any idea on when the release will happen?

elezar:
@carreter we included the fix in the v1.18.0 release. Please let us know if the workaround propopsed in #1222 is not sufficient.

PR #1256, opened 2025-08-21
Cleanup default runtime in runtime config when setAsDefault=false

cdesiniotis:
This commit updates the AddRuntime() implementation for containerd and cri-o to clear the default runtime field if the runtime we are adding is currently set as the default AND setAsDefault=false.

### DIFF
--- a/pkg/config/engine/containerd/config.go
+++ b/pkg/config/engine/containerd/config.go
@@ -64,6 +64,11 @@ func (c *Config) AddRuntime(name string, path string, setAsDefault bool) error {
 
 	if setAsDefault {
 		config.SetPath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"}, name)
+	} else {
+		defaultRuntime, ok := config.GetPath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"}).(string)
+		if ok && defaultRuntime == name {
+			config.DeletePath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"})
+		}
 	}
 
 	*c.Tree = config
--- a/pkg/config/engine/containerd/config_test.go
+++ b/pkg/config/engine/containerd/config_test.go
@@ -53,6 +53,26 @@ func TestAddRuntime(t *testing.T) {
 			expectedError: nil,
 		},
 		{
+			description:  "empty config, set as default runtime",
+			setAsDefault: true,
+			expectedConfig: `
+			version = 2
+			[plugins]
+			[plugins."io.containerd.grpc.v1.cri"]
+				[plugins."io.containerd.grpc.v1.cri".containerd]
+				default_runtime_name = "test"
+				[plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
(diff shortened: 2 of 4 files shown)


pkg/config/engine/containerd/config.go
```diff
@@ -64,6 +64,12 @@ func (c *Config) AddRuntime(name string, path string, setAsDefault bool) error {
 
 	if setAsDefault {
 		config.SetPath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"}, name)
+	} else {
+		if defaultRuntime, ok := config.GetPath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"}).(string); ok {
+			if defaultRuntime == name {
+				config.DeletePath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"})
+			}
+		}
```
elezar: What about the following to clean up intendation a bit:
```suggestion
		if defaultRuntime, ok := config.GetPath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"}).(string); ok && defaultRuntime == name {
			config.DeletePath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"})
		}
```
alternatively, it may be more readable if we do:
```suggestion
		defaultRuntime, ok := config.GetPath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"}).(string)
		if ok && defaultRuntime == name {
			config.DeletePath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "default_runtime_name"})
		}
```
which also decreases the nesting but makes the conditional easier to read.

pkg/config/engine/containerd/config_test.go
```diff
@@ -257,6 +277,75 @@ func TestAddRuntime(t *testing.T) {
...
+					runtime_engine = ""
+					runtime_root = ""
+					runtime_type = "io.containerd.runc.v2"
+					[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.test.options]
+						BinaryName = "/usr/bin/test"
+			`,
+			expectedError: nil,
+		},
+		{
+			description:  "runtime already exists in config, not default runtime",
```
elezar, line 316: I didn't check if it's covered explicilty, but what about the case where `default_runtime_name` exists but is not set to `test`?

pkg/config/engine/containerd/config.go
tariq1890: I prefer the latter suggestion; definitely more readable

pkg/config/engine/containerd/config_test.go
cdesiniotis: Yes, I believe this is covered in a prior test case on line 117 with description: `options from default runtime are imported`. For this test case, the original config has `default_runtime_name: default` and the `setAsDefault` setting is configured as `false`.

pkg/config/engine/containerd/config.go
cdesiniotis: I agree the latter suggestion is more readable. I have updated this now.

PR #1263, opened 2025-08-25
Remove redundant CDI annotations

### LINKED ISSUE #1262: Remove device annotations from generated specs
The annotations added to support the generation of CDI specs for coherent and noncoherent devices bump the required CDI spec version to v0.6.0. This is breaks compatibility with older podman versions.

The annotations are not strictly required in the final SPEC and this should be REMOVED once we have generated to different / separate specs.

elezar:
When generating CDI specs for coherent and noncoherent devices, we include CDI device annotations. This increases the minimum CDI spec version to v0.6.0 which is not as widely supported as v0.5.0 or v0.3.0.

This change removes these annotations after splitting the generated specs to produce coherent and/or non-coherent specs.

Closes #1262 

Note that since we're generating a CDI spec for `nvidia.com/gpu.coherent` devices, these are `v0.6.0` specs since `.` was only allowed in the spec from `v0.6.0`.

### DIFF
--- a/cmd/nvidia-ctk/cdi/generate/generate.go
+++ b/cmd/nvidia-ctk/cdi/generate/generate.go
@@ -447,15 +447,22 @@ type deviceSpecs []specs.Device
 func (d deviceSpecs) splitOnAnnotation(key string) map[string][]specs.Device {
 	splitSpecs := make(map[string][]specs.Device)
 
+	var specsToRemoveAnnotations []*specs.Device
 	for _, deviceSpec := range d {
-		if len(deviceSpec.Annotations) == 0 {
-			continue
-		}
 		value, ok := deviceSpec.Annotations[key]
 		if !ok {
 			continue
 		}
 		splitSpecs[key+"="+value] = append(splitSpecs[key+"="+value], deviceSpec)
+		specsToRemoveAnnotations = append(specsToRemoveAnnotations, &deviceSpec)
+	}
+
+	// We also remove the annotations that were used to split the devices:
+	for _, deviceSpec := range specsToRemoveAnnotations {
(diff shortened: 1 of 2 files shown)


cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -447,6 +447,7 @@ type deviceSpecs []specs.Device
 func (d deviceSpecs) splitOnAnnotation(key string) map[string][]specs.Device {
 	splitSpecs := make(map[string][]specs.Device)
 
+	var specsToRemoveAnnotations []*specs.Device
```
ArangoGutierrez, line 450:
```suggestion
for _, ds := range d {
    if v, ok := ds.Annotations[key]; ok {
        // split key
        k := key + "=" + v
        // mutate a copy before appending
        cp := ds
        if cp.Annotations != nil {
            delete(cp.Annotations, key)
            if len(cp.Annotations) == 0 {
                cp.Annotations = nil
            }
        }
        splitSpecs[k] = append(splitSpecs[k], cp)
    }
}
```

how about we merge both `for` loops into 1?

elezar: The problem is that the annotations in the copy are still the annotations in the iterator since these are copied by reference. This means that removing the annotation in the first pass will mean that subsequent device specs that reference the SAME annotation will no longer have the annotation. 

This is an issue because when we generate the specs, we duplicate these for the same device referenced using different names (e.g. index and uuid).

ArangoGutierrez (approved):
LGTM, thanks for the reply on my suggestion

PR #1270, opened 2025-08-27
Don't inject enable-cuda-compat hook in CSV mode

elezar:
Since the logic to determine the host driver version relies on the extension of `libcuda.so.<RM_VERSION>`, this is not supported on some Tegra-based systems where the file always has the name `libcuda.so.1.1`. This means that the `enable-cuda-compat` hook should not be injected in CSV mode in the `nvidia` runtime.

### DIFF
--- a/internal/modifier/gated.go
+++ b/internal/modifier/gated.go
@@ -92,6 +92,11 @@ func NewFeatureGatedModifier(logger logger.Interface, cfg *config.Config, image 
 }
 
 func getCudaCompatModeDiscoverer(logger logger.Interface, cfg *config.Config, driver *root.Driver, hookCreator discover.HookCreator) (discover.Discover, error) {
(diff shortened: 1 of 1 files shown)


agirault:
> Since the logic to determine the host driver version relies on the extension of libcuda.so.<RM_VERSION>, this is not supported on Tegra-based systems

Why? Thor is using openRM

ArangoGutierrez (approved):

elezar:
> > Since the logic to determine the host driver version relies on the extension of libcuda.so.<RM_VERSION>, this is not supported on Tegra-based systems
> 
> Why? Thor is using openRM

Yes, that might be the case, but I think there are still a number of situations where the driver installation is still different depending on the Jetpack release. (Or that is my understanding).

Issue #1260, opened 2025-08-25
Use drop-in files to configure containerd or cri-o


elezar:
Instead of producing a full config when configuring a runtime such as containerd or cri-o, we should produce minimal "drop-in" files to make the configurations of runtimes more robust. This also simplifies the removal of the nvidia runtimes as the relevant drop-in file(s) can be deleted to do so.

PR #1272, opened 2025-08-27
Implement Drop-In Lifecycle Support for Containerd and Crio

ArangoGutierrez:


cmd/nvidia-ctk/runtime/configure/configure.go
```diff
@@ -46,6 +46,10 @@ const (
 	defaultCrioConfigFilePath       = "/etc/crio/crio.conf"
 	defaultDockerConfigFilePath     = "/etc/docker/daemon.json"
 
+	defaultContainerdDropInDir = "/etc/containerd/conf.d"
+	defaultCrioDropInDir       = "/etc/crio/conf.d"
+	defaultDropInFileName      = "99.nvidia.toml"
```
elezar:
```suggestion
	defaultDropInFileName      = "99-nvidia.toml"
```

cmd/nvidia-ctk/runtime/configure/configure.go
elezar, line 1: What about the following:
```
nvidia-ctk runtime configure \
    --runtime=containerd \
    --config=/etc/containerd/config.toml \
    --nvidia-config=/etc/containerd/conf.d/99-nvidia.toml
```
and
```
nvidia-ctk runtime configure \
    --runtime=crio \
    --config=/etc/crio/crio.conf \
    --nvidia-config=/etc/crio/conf.d/99-nvidia.toml
```
the drop in dirs can be inferred as the PARENT dir of the `nvidia-config` file.

Note that IFF `--nvidia-config=""` we could trigger the same behaviour as `--dry-run` in which case no config modifications are made to the specified `config` files.

Does it make sense to set up tests for the various combinations here and then use that to drive development?

cmd/nvidia-ctk/runtime/configure/configure.go
```diff
@@ -170,6 +175,11 @@ func (m command) build() *cli.Command {
 				Usage:       "Enable CDI in the configured runtime",
 				Destination: &config.cdi.enabled,
 			},
+			&cli.StringFlag{
+				Name:        "drop-in-config-path",
+				Usage:       "the path to the drop-in configuration file for the containerd runtime",
+				Destination: &config.dropInConfigPath,
+			},
```
elezar: As a note: Here we use the same option and in the toolkit container logic we only expose a flag for containerd.

pkg/config/engine/containerd/config.go
```diff
@@ -96,49 +93,143 @@ func (c *Config) getRuntimeAnnotations(path []string) ([]string, error) {
...
+	if err := c.ensureMainConfigWithImports(); err != nil {
+		return 0, fmt.Errorf("failed to ensure main config with imports: %w", err)
+	}
+
+	// Create drop-in directory if it doesn't exist
+	if err := os.MkdirAll(c.dropInConfigPath, 0755); err != nil {
+		return 0, fmt.Errorf("failed to create drop-in directory %s: %w", c.dropInConfigPath, err)
+	}
+
+	return c.NVConfig.Save(path)
```
elezar: I believe this may be saving to the wrong path. Why is this not the drop in file?

pkg/config/engine/containerd/config.go
```diff
@@ -96,49 +93,143 @@ func (c *Config) getRuntimeAnnotations(path []string) ([]string, error) {
...
+	_, err := os.Stat(d.baseConfigPath)
+	if os.IsNotExist(err) {
+		// Create minimal config with imports
+		return d.createMinimalMainConfig()
+	} else if err != nil {
+		return fmt.Errorf("failed to check main config: %w", err)
+	}
+
+	// Main config exists, ensure it has imports directive
+	return d.ensureImportsDirective()
```
elezar: Should the only difference between these two cases not be whether the original config is empty or not. Updating the imports should not require different logic.

pkg/config/engine/containerd/containerd.go
```diff
@@ -100,8 +109,20 @@ func New(opts ...Option) (engine.Interface, error) {
 	}
 	b.logger.Infof("Using CRI runtime plugin name %q", criRuntimePluginName)
 
+	var dropInConfigPath string
+	if b.dropInDir != "" {
+		dropInConfigPath = b.dropInDir
```
elezar: There seems to be an inconsistency here. Is this a directory or a file path?

pkg/config/engine/crio/crio.go
```diff
@@ -67,39 +79,52 @@ func New(opts ...Option) (engine.Interface, error) {
 		return nil, err
 	}
 
+	var dropInConfigPath string
+	if b.dropInDir != "" {
```
elezar: I would argue that IF the dropIn config path is `""` then we should have the same behviour as we currently have.

elezar (commented):
As discussed offline, I think the approach here is differs from what my first instinct was to more clearly split SOURCE and TARGET / NVIDIA runtimes but to keep the logic pretty much as is. Let us assume that we make the changes as described in https://github.com/NVIDIA/nvidia-container-toolkit/pull/1279, then the switch to drop in files SHOULD reduce to implementing `GetDefaultRuntimeOptions` to pull these from the SOURCE config, whereas any functions that MODIFY the config should act on the TARGET config.

It may be useful to list the various combinations more clearly in the design doc or some kind of test here (please point it out if it is there).

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
elezar, line 1: I was looking at these tests, and I don't think they are actually too useful. What I think we should do here is replace these with tests that exercise the `Setup->Cleanup` cycle. We can use these to exercise the various scenarios that we're interested in.

For example, if we have:
```
func TestSetupCleanup(t *testing.T) {
	c := &cli.Command{
		Name: "test",
	}
	testCases := []struct {
		description                 string
		containerOptions            container.Options
		options                     Options
		prepareEnvironment          func(*testing.T, string) error
		expectedSetupError          error
		assertSetupPostConditions   func(*testing.T, string) error
		expectedCleanupError        error
		assertCleanupPostConditions func(*testing.T, string) error
	}{
		{
			description: "top-level config does not exist",
			containerOptions: container.Options{
				DropInConfig: "{{ .testRoot }}/conf.d/99-nvidia.toml",
				Config:       "{{ .testRoot }}/etc/containerd/config.toml",
			},
			assertSetupPostConditions: func(t *testing.T, testRoot string) error {
				require.FileExists(t, filepath.Join(testRoot, "/etc/containerd/config.toml"))
				topLevel, err := toml.FromFile(filepath.Join(testRoot, "/etc/containerd/config.toml")).Load()
				require.NoError(t, err)
				require.EqualValues(t, `imports = ["`+filepath.Join(testRoot, "/conf.d/*.toml")+`"]
version = 2
`, topLevel.String())
				require.FileExists(t, filepath.Join(testRoot, "/conf.d/99-nvidia.toml"))
				dropIn, err := toml.FromFile(filepath.Join(testRoot, "/conf.d/99-nvidia.toml")).Load()
				require.NoError(t, err)
				require.EqualValues(t, `version = 2

[plugins]

  [plugins."io.containerd.grpc.v1.cri"]

    [plugins."io.containerd.grpc.v1.cri".containerd]

      [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]

        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia]
          privileged_without_host_devices = false
          runtime_engine = ""
          runtime_root = ""
          runtime_type = ""

          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia.options]
            BinaryName = "nvidia-container-runtime"

        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia-cdi]
          privileged_without_host_devices = false
          runtime_engine = ""
          runtime_root = ""
          runtime_type = ""

          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia-cdi.options]
            BinaryName = "nvidia-container-runtime.cdi"

        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia-legacy]
          privileged_without_host_devices = false
          runtime_engine = ""
          runtime_root = ""
          runtime_type = ""

          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia-legacy.options]
            BinaryName = "nvidia-container-runtime.legacy"
`, dropIn.String())
				return nil
			},
			assertCleanupPostConditions: func(t *testing.T, testRoot string) error {
				require.NoFileExists(t, filepath.Join(testRoot, "/etc/containerd/config.toml"))
				require.NoFileExists(t, filepath.Join(testRoot, "/conf.d/99-nvidia.toml"))
				return nil
			},
		},
	}

		// TODO: Add tests for restart mode.
		tc.containerOptions.RestartMode = "none"
			// Update any paths as required:
			testRoot := t.TempDir()
			tc.containerOptions.Config = strings.ReplaceAll(tc.containerOptions.Config, "{{ .testRoot }}", testRoot)
			tc.containerOptions.DropInConfig = strings.ReplaceAll(tc.containerOptions.DropInConfig, "{{ .testRoot }}", testRoot)

			// Prepare the test environment
			if tc.prepareEnvironment != nil {
				require.NoError(t, tc.prepareEnvironment(t, testRoot))
			}

			err := Setup(c, &tc.containerOptions, &tc.options)
			require.EqualValues(t, tc.expectedSetupError, err)

			if tc.assertSetupPostConditions != nil {
				require.NoError(t, tc.assertSetupPostConditions(t, testRoot))
			}

			err = Cleanup(c, &tc.containerOptions, &tc.options)
			require.EqualValues(t, tc.expectedCleanupError, err)

			if tc.assertCleanupPostConditions != nil {
				require.NoError(t, tc.assertCleanupPostConditions(t, testRoot))
			}
		})
	}
}
```
this tests the scenario where we have no initial config file and expect it to be removed once we clean things up.

Other scenarios would be:
* A config.toml exsists
* A config.toml exists AND has the `nvidia` runtime defined.

With some work we could also test different config sources, but that need not be the focus of this PR.

ArangoGutierrez:
Fixes https://github.com/NVIDIA/nvidia-container-toolkit/issues/1260

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
ArangoGutierrez: This is now being discussed in a separate PR, see https://github.com/NVIDIA/nvidia-container-toolkit/pull/1293

cmd/nvidia-ctk/runtime/configure/configure.go
ArangoGutierrez: I have implemented this approach. PTAL

ArangoGutierrez:
> As discussed offline, I think the approach here is differs from what my first instinct was to more clearly split SOURCE and TARGET / NVIDIA runtimes but to keep the logic pretty much as is. Let us assume that we make the changes as described in #1279, then the switch to drop in files SHOULD reduce to implementing `GetDefaultRuntimeOptions` to pull these from the SOURCE config, whereas any functions that MODIFY the config should act on the TARGET config.
> 
> It may be useful to list the various combinations more clearly in the design doc or some kind of test here (please point it out if it is there).

This PR now rebases #1279 and builds on the concepts introduced there; now this PR is smaller and focused. 
The test bit's are also in an independent PR #1293

cmd/nvidia-ctk/runtime/configure/configure.go
```diff
@@ -118,6 +119,11 @@ func (m command) build() *cli.Command {
 				Usage:       "path to the config file for the target runtime",
 				Destination: &config.configFilePath,
 			},
+			&cli.StringFlag{
+				Name:        "nvidia-config",
```
elezar: Looking at this again, I don't think `nvidia-config` is specific enough. Something like `drop-in-config` may be better at communicating the intent here.

ArangoGutierrez: ack, will edit

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -54,6 +56,13 @@ func Flags(opts *Options) []cli.Flag {
 			Destination: &opts.Config,
 			Sources:     cli.EnvVars("RUNTIME_CONFIG", "CONTAINERD_CONFIG", "DOCKER_CONFIG"),
 		},
+		&cli.StringFlag{
+			Name:        "drop-in-config",
+			Usage:       "Path to the NVIDIA-specific config file to create. When specified, runtime configurations are saved to this file instead of modifying the main config file",
+			Destination: &opts.NvidiaConfig,
+			Value:       defaultNVIDIARuntimeConfigFilePath,
```
elezar, line 63: Question: How does this work for crio? Do we also have the ability to update the imports there? Is this something that we want to do?

ArangoGutierrez: It is possible to point to a diff location than the default one (see https://www.ibm.com/docs/en/zoscp/1.1.0?topic=options-crio-commands-flags#crio-commands__title__4) 
`--config-dir | Path to the configuration drop-in directory.`

So we could offer the same `drop-in-config` flag on the `crio` side. 
Let me see how that would look like

elezar: To be clear. I'm ok for runtime-specific defaults here. We already do this for other settings here: https://github.com/NVIDIA/nvidia-container-toolkit/blob/470b8411a0500abf48f562cb2b7cad97a09e3894/cmd/nvidia-ctk-installer/container/runtime/runtime.go#L128-L159

ArangoGutierrez: Ok, I think is ok to go with defaults for now

Issue #1271, opened 2025-08-27
The enable-cuda-compat hook should support Tegra-based systems


elezar:
The current injection of the `enable-cuda-compat` hook checks the version suffix of the `libcuda.so.RM_VERSION` file to determine the host driver version. On tegra-based systems this is not correct as the file has the name `libcuda.so.1.1` meaning that the compat libraries in a container are ALWAYS used.

The CDI spec generation for Tegra-based systems (mode=csv) must be updated to:
1. read the driver version from nvml if available
2. update the generation of the `enable-cuda-compat` hook to handle the orin case. 

This means that if an Orin system is detected, alternative CUDA compat libraries (rooted at `/usr/local/cuda/compat-orin` in the container). If these compat libraries are not present, then the host drivers will always be used effectively disabling CUDA forward compat on Orin systems.

The following PRs are relevant:
- [ ] https://github.com/NVIDIA/nvidia-container-toolkit/pull/1515

cdesiniotis:
Closing as this was included in v1.19.0. @elezar please re-open if I am missing something.

PR #1278, opened 2025-08-28
Don't inject enable-cuda-compat hook in CSV mode

elezar:
When running the nvidia runtime in csv mode, we should not inject the enable-cuda-compat hook until we make the host driver version detection more robust on these platforms. As it is, the host driver version detection returns a version of 1.1 (the suffix of libcuda.so) meaning that CUDA forward compatibility will always be used in a container regardless of whether this is required.

This will be properly added to the generated CDI spec for CSV-based platforms as a follow-up. See #1271 

This is a backport of #1270


elezar:
cc @agirault

PR #1279, opened 2025-08-28
[no-relnote] Refactor AddRuntime

elezar:
This change refactors the AddRuntime function into GetDefaultRuntimeOptions and AddRuntimeWithOptions. This allows for a clearer distinction between SOURCE and DESTINATION configs which should facilitate switching to drop-in files.

### DIFF
--- a/pkg/config/engine/containerd/config.go
+++ b/pkg/config/engine/containerd/config.go
@@ -28,29 +28,37 @@ func (c *Config) AddRuntime(name string, path string, setAsDefault bool) error {
 	if c == nil || c.Tree == nil {
 		return fmt.Errorf("config is nil")
 	}
+	defaultRuntimeOptions := c.GetDefaultRuntimeOptions()
+	return c.AddRuntimeWithOptions(name, path, setAsDefault, defaultRuntimeOptions)
+}
+
+func (c *Config) GetDefaultRuntimeOptions() interface{} {
+	runtimeNamesForConfig := engine.GetLowLevelRuntimes(c)
+	for _, r := range runtimeNamesForConfig {
+		options := c.GetSubtreeByPath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "runtimes", r})
+		if options != nil {
+			c.Logger.Debugf("Using options from runtime %v: %v", r, options)
+			return options.Copy()
+		}
+	}
+	c.Logger.Warningf("Could not infer options from runtimes %v", runtimeNamesForConfig)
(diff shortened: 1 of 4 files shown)


pkg/config/engine/containerd/config.go
```diff
@@ -28,29 +28,37 @@ func (c *Config) AddRuntime(name string, path string, setAsDefault bool) error {
...
-		c.Logger.Debugf("using options from runtime %v: %v", r, options)
-		config.SetPath([]string{"plugins", c.CRIRuntimePluginName, "containerd", "runtimes", name}, options.Copy())
-		break
 	}
+	c.Logger.Warningf("Could not infer options from runtimes %v", runtimeNamesForConfig)
+	options, _ := toml.TreeFromMap(map[string]interface{}{
+		"runtime_type":                    c.RuntimeType,
+		"runtime_root":                    "",
+		"runtime_engine":                  "",
+		"privileged_without_host_devices": false,
```
ArangoGutierrez, line 49: I wonder if, on this fallback, we should add the `SystemdCgroup = true|false` based on a function that looks if the system is running systemd.

pkg/config/engine/containerd/config_v1.go
```diff
@@ -33,59 +33,36 @@ func (c *ConfigV1) AddRuntime(name string, path string, setAsDefault bool) error
...
-			}
-			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "options", "BinaryName"}, path)
-			config.SetPath([]string{"plugins", "cri", "containerd", "default_runtime", "options", "Runtime"}, path)
-		}
+	if !c.UseLegacyConfig || !setAsDefault {
+		return nil
 	}
 
+	config = *c.Tree
+	// Note: This is deprecated in containerd 1.4.0 and will be removed in 1.5.0
```
ArangoGutierrez, line 57: Should we edit the note, now that `1.5.0` is released? and add a warning log.

pkg/config/engine/containerd/config.go
elezar: The intent of this change is not to add additional behaviour. It is a refactor.

pkg/config/engine/containerd/config_v1.go
elezar: The intent of this change is not to add additional behaviour. It is a refactor.

ArangoGutierrez (approved):
LGTM
I made a note to turn my 2 comments into TODO items

PR #1280, opened 2025-08-28
Add drop-in file support for containerd and crio

elezar:
This change adds drop-in file support for containerd and crio. Instead of writing the entire in-memory config to the specified config path, we write only the modified settings (nvidia runtimes, default runtime, enable_cdi) settings to a drop-in file and update the top-level configs to import this if required.

See also #1272

For GPU Operator integration see https://github.com/NVIDIA/gpu-operator/pull/1710

### DIFF
--- a/pkg/config/engine/config.go
+++ b/pkg/config/engine/config.go
@@ -1,0 +1,85 @@
+/**
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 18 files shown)


pkg/config/toml/source-empty.go
elezar, line 1: This could strictly speaking also be a separate commit.

ArangoGutierrez: https://github.com/NVIDIA/nvidia-container-toolkit/pull/1299

ArangoGutierrez:
/retitle "Implement Drop-In Lifecycle Support for Containerd and Crio"

pkg/config/engine/config.go
```diff
@@ -0,0 +1,85 @@
...
+	String() string
+}
+
+// AddRuntime adds a runtime to the destination config and optionally sets it as the default.
+// The options to apply to the added runtime are read from the source config
+// default runtime.
+func (c *Config) AddRuntime(name string, path string, setAsDefault bool) error {
+	options := c.Source.GetDefaultRuntimeOptions()
+	return c.Destination.AddRuntimeWithOptions(name, path, setAsDefault, options)
+}
```
cdesiniotis, line 56: I believe this means the changes I made in https://github.com/NVIDIA/nvidia-container-toolkit/pull/1256 will no longer take affect. If `default_runtime_name=nvidia` is set in the top-level config, and `setAsDefault=false`, we don't actually unset `nvidia` as the default runtime.

PR #1293, opened 2025-09-08
[no-relnote] Refactor cmd/installer/container/containerd unit test

ArangoGutierrez:


containerd.test
elezar, line 1: Should this file have been included?

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
```diff
@@ -30,51 +30,136 @@ import (
...
+			enableCDI:    true,
+			setAsDefault: true,
+			withRunc:     true,
 		},
 	}
+}
+
+// TestUpdateV2Config_NoConfigFile tests the scenario when there is no
+// containerd config file present
+func TestUpdateV2Config_NoConfigFile(t *testing.T) {
```
elezar: As discussed:
* offline in our call yesterday
* https://github.com/NVIDIA/nvidia-container-toolkit/pull/1272#discussion_r2309632391
* https://github.com/NVIDIA/nvidia-container-toolkit/pull/1280/files#diff-743eab44a71e6e333fef0be742b20bcdb7188341b5b8e6542bc94ec64704de35R35

I think these tests are testing the wrong thing. Although we may want to consider updating the tests for `Update` config, what we really want is tests to ensure the correct behaviour of the `Setup`->`Cleanup` cycle.

Would adapting the proposed implementation from the last two points above be feasible?

containerd.test
ArangoGutierrez: No, removing

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
ArangoGutierrez: Sure, thanks for the clarification. I now understand that I got the wrong idea from our conversation. I'll work on the proposed implementation

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
```diff
@@ -17,504 +17,385 @@
...
+				RestartMode:  "none",
+			},
+			options: Options{
+				runtimeType: defaultRuntmeType,
+			},
+			prepareEnvironment: func(t *testing.T, testRoot string) error {
+				configPath := filepath.Join(testRoot, "etc/containerd/config.toml")
+				require.NoError(t, os.MkdirAll(filepath.Dir(configPath), 0755))
+
+				cfg, err := containerd.New(
```
elezar: Question: Would this be cleaner to read if we were using a raw string?

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
ArangoGutierrez: I have implemented the tests following the structure of the comment https://github.com/NVIDIA/nvidia-container-toolkit/pull/1272#discussion_r2309632391 , PTAL

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
```diff
@@ -17,504 +17,385 @@
...
-			setAsDefault:               true,
-			runtimeName:                "nvidia",
-			expectedDefaultRuntimeName: "nvidia",
-		},
-		{
-			setAsDefault:               true,
-			runtimeName:                "NAME",
-			expectedDefaultRuntimeName: "NAME",
-		},
+// TestSetupCleanup tests the complete Setup->Cleanup lifecycle following Evan Lezar's pattern
```
elezar: Let's rather give the function a more expressive name than relying on a comment in this case. (This should only be required if we feel `TestSetupCleanup` is not sufficient).

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
```diff
@@ -17,504 +17,385 @@
...
+				verifyRuntimesPresent(t, filepath.Join(testRoot, "etc/containerd/config.toml"),
+					"nvidia", "nvidia-cdi", "nvidia-legacy")
+				return nil
+			},
+			assertCleanupPostConditions: func(t *testing.T, testRoot string) error {
+				// Verify that nvidia runtimes were removed
+				if _, err := os.Stat(filepath.Join(testRoot, "etc/containerd/config.toml")); err == nil {
+					verifyRuntimesAbsent(t, filepath.Join(testRoot, "etc/containerd/config.toml"),
+						"nvidia", "nvidia-cdi", "nvidia-legacy")
+				}
```
elezar: Should we not be checking for AN EMPTY CONFIG here? (actually, I believe the expected behaviour is that the config file is deleted).

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
```diff
@@ -17,504 +17,385 @@
...
-									"privileged_without_host_devices": false,
-									"container_annotations":           []string{"cdi.k8s.io/*"},
-									"options": map[string]interface{}{
-										"BinaryName": "/test/runtime/dir/nvidia-container-runtime.legacy",
+			description: "existing config without nvidia runtime",
+			containerOptions: container.Options{
+				Config:       "{{ .testRoot }}/etc/containerd/config.toml",
+				RuntimeName:  "nvidia",
+				RuntimeDir:   "/usr/bin",
+				EnableCDI:    true,
```
elezar: This ALSO enabled CDI.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
```diff
@@ -17,504 +17,385 @@
...
+			description: "top-level config does not exist",
+			containerOptions: container.Options{
+				Config:       "{{ .testRoot }}/etc/containerd/config.toml",
+				RuntimeName:  "nvidia",
+				RuntimeDir:   "/usr/bin",
+				SetAsDefault: false,
+				RestartMode:  "none",
+			},
+			options: Options{
+				runtimeType: defaultRuntimeType,
```
elezar: If we always have to set this, should we not set it in the test body?

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
```diff
@@ -17,504 +17,385 @@
...
+				require.NoError(t, err)
+				_, err = cfg.Save(configPath)
+				require.NoError(t, err)
+				return nil
+			},
+			assertSetupPostConditions: func(t *testing.T, testRoot string) error {
+				configPath := filepath.Join(testRoot, "etc/containerd/config.toml")
+				require.FileExists(t, configPath)
+				verifyRuntimesPresent(t, configPath, "nvidia", "nvidia-cdi", "nvidia-legacy")
+				verifyCDIEnabled(t, configPath, true)
```
elezar: I would that it's easier to reason about the full config. If we're concerned about raw strings, we could always use the toml package to adjust for formatting differences.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
```diff
@@ -17,504 +17,385 @@
...
+				require.FileExists(t, configPath)
+				verifyRuntimesAbsent(t, configPath, "nvidia", "nvidia-cdi", "nvidia-legacy")
+
+				// Default runtime should be cleared
+				cfg, err := containerd.New(
+					containerd.WithPath(configPath),
+					containerd.WithConfigSource(toml.FromFile(configPath)),
+				)
+				require.NoError(t, err)
+				c := cfg.(*containerd.Config)
```
elezar: THis leaks implementation details. As mentioned in our discussion, these tests were never properly updated when we migrated the logic to a central package for use from the `nvidia-ctk runtime configure` command. We should check the content of the files that we're interested in.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
```diff
@@ -17,504 +17,385 @@
...
-		t.Run(fmt.Sprintf("%v", tc.enableCDI), func(t *testing.T) {
-			o := &container.Options{
-				EnableCDI:    tc.enableCDI,
-				RuntimeName:  "nvidia",
-				RuntimeDir:   runtimeDir,
-				SetAsDefault: false,
+		t.Run(tc.description, func(t *testing.T) {
+			// Update any paths as required
+			testRoot := t.TempDir()
+			tc.containerOptions.Config = strings.ReplaceAll(tc.containerOptions.Config, "{{ .testRoot }}", testRoot)
```
elezar: Out of scope for this PR, but we could consider JSON marshalling, replacing the string, and then unmarshalling again. For a single field this is fine though.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v1_test.go
elezar, line 1: I'm happy to only focus on the v2 tests as part of this PR, but we could consider also also including v1 tests there. (If we move to testing the `Setup->Cleanup` cycle, then we don't need to distinguish between v1 and v2 tests.

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
```diff
@@ -0,0 +1,520 @@
...
+					crio.WithPath(configPath),
+					crio.WithConfigSource(toml.FromFile(configPath)),
+				)
+				require.NoError(t, err)
+				c := cfg.(*crio.Config)
+
+				// Check non-runtime settings
+				require.Equal(t, "/usr/libexec/crio/conmon", c.GetPath([]string{"crio", "runtime", "conmon"}))
+				require.Equal(t, true, c.GetPath([]string{"crio", "runtime", "selinux"}))
+				require.Equal(t, "/etc/cni/net.d/", c.GetPath([]string{"crio", "network", "network_dir"}))
```
elezar: Let's just compare raw strings.

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
```diff
@@ -0,0 +1,520 @@
...
+
+			if tc.assertCleanupPostConditions != nil {
+				require.NoError(t, tc.assertCleanupPostConditions(t, testRoot))
+			}
+		})
+	}
+}
+
+// TestSetupCleanupHookMode tests the Setup->Cleanup lifecycle for hook mode
+func TestSetupCleanupHookMode(t *testing.T) {
```
elezar: Is there a reason that this needs to be a separate function?

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
```diff
@@ -1,5 +1,5 @@
 /**
-# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
elezar: We don't change the year of existing files.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
ArangoGutierrez: Thanks for noticing, I have codified your suggestion

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
ArangoGutierrez: test description updated

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
ArangoGutierrez: Moved to raw string

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
ArangoGutierrez: test func renamed

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
ArangoGutierrez: ~We now use `NoFileExist` check~

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v1_test.go
ArangoGutierrez: Merged v1 and v2 now that we test focusing on the lifecycle

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
ArangoGutierrez: merged into the main test func

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
ArangoGutierrez: Moved to raw string

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
ArangoGutierrez: EDIT: the behaviour WILL be to delete the file, but currently it is not , see https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/pkg/config/engine/containerd/config.go#L127 , the logic to delete the file is introduced on the DropIns PR at https://github.com/NVIDIA/nvidia-container-toolkit/pull/1272/files#diff-59fededabdcd1a2abf2309e2d74e17b98c379747eb6fa2bdff7a24edffab8c14R129

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
ArangoGutierrez: TODO @ArangoGutierrez, let's add JSON marshaling

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
elezar: Did you perhaps forget to push this change? Where did you change to checking raw strings? To be clear, what I mean by "raw strings" is that the test will contain a raw (i.e. backtick) string containig the expected contents of the config file in the various steps. We already do this for the CDI specs we generated, for example.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
elezar: That is strange. We *should* be deleting the file if -- after removing `nvidia` runtimes -- there is only a `version` entry in the config. (We remove the version entry [here](https://github.com/NVIDIA/nvidia-container-toolkit/blob/8739e483f1800964adbb920335db7e7d32a88d02/pkg/config/engine/containerd/config.go#L150-L152) and then remove the output file when [writing the raw toml](https://github.com/NVIDIA/nvidia-container-toolkit/blob/8739e483f1800964adbb920335db7e7d32a88d02/pkg/config/raw.go#L39-L45)).

This is even more reason to *CHECK RAW STRINGS* so that we know the complete state of the config file we generate.

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
```diff
@@ -0,0 +1,451 @@
...
+	cli "github.com/urfave/cli/v3"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/container"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/config"
+	"github.com/NVIDIA/nvidia-container-toolkit/pkg/config/engine/crio"
+	"github.com/NVIDIA/nvidia-container-toolkit/pkg/config/toml"
+)
+
+// TestSetupCleanup tests the complete Setup->Cleanup lifecycle for both config and hook modes
+func TestSetupCleanup(t *testing.T) {
```
elezar: Let's at least be consistent between the two runtimes.

ArangoGutierrez: agree, done

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
elezar: OK. I have made the following changes to the test:
```
diff --git a/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go b/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
index 5847d86f..03bf60d9 100644
--- a/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
@@ -68,11 +68,7 @@ func TestContainerdConfigLifecycle(t *testing.T) {
 			assertCleanupPostConditions: func(t *testing.T, testRoot string) error {
 				configPath := filepath.Join(testRoot, "etc/containerd/config.toml")
 
-				if _, err := os.Stat(configPath); err == nil {
-					// TODO(@ArangoGutierrez): When moving to DropIn files, we should verify that the config file is deleted after cleanup.
-					// For now, we just verify that the runtimes are absent.
-					verifyRuntimesAbsent(t, configPath, "nvidia", "nvidia-cdi", "nvidia-legacy")
-				}
+				require.NoFileExists(t, configPath)
 				return nil
 			},
 		},
```
(Note that I check file not exists)

and the test still passes. Could you please elaborate on what you based the statement:
> EDIT: the behaviour WILL be to delete the file, but currently it is not , see https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/pkg/config/engine/containerd/config.go#L127 , the logic to delete the file is introduced on the DropIns PR at https://github.com/NVIDIA/nvidia-container-toolkit/pull/1272/files#diff-59fededabdcd1a2abf2309e2d74e17b98c379747eb6fa2bdff7a24edffab8c14R129

on?

ArangoGutierrez: Did you run those changes on a Mac or a Linux machine? 
I had it like you showed before, and it passed on my Mac, but then it failed on the GitHub Runner (Linux)

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
elezar: As was the case for checking the OUTPUT, I think there was a misunderstanding here. I mean that we define the CONTENT of the input file explicitly and no as any specific type. For example:
```
diff --git a/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go b/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
index 5847d86f..daa5d96f 100644
--- a/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
@@ -92,29 +92,24 @@ func TestContainerdConfigLifecycle(t *testing.T) {
 				configPath := filepath.Join(testRoot, "etc/containerd/config.toml")
 				require.NoError(t, os.MkdirAll(filepath.Dir(configPath), 0755))
 
-				cfg, err := containerd.New(
-					containerd.WithConfigSource(toml.FromMap(map[string]interface{}{
-						"version": int64(2),
-						"plugins": map[string]interface{}{
-							"io.containerd.grpc.v1.cri": map[string]interface{}{
-								"containerd": map[string]interface{}{
-									"default_runtime_name": "runc",
-									"runtimes": map[string]interface{}{
-										"runc": map[string]interface{}{
-											"runtime_type": "io.containerd.runc.v2",
-											"options": map[string]interface{}{
-												"BinaryName": "/usr/bin/runc",
-											},
-										},
-									},
-								},
-							},
-						},
-					})),
-				)
-				require.NoError(t, err)
-				_, err = cfg.Save(configPath)
-				require.NoError(t, err)
+				contents := `version = 2
+
+				[plugins]
+
+				  [plugins."io.containerd.grpc.v1.cri"]
+
+					[plugins."io.containerd.grpc.v1.cri".containerd]
+					   default_runtime_name = "runc"
+					  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
+
+						[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+						  runtime_type = "io.containerd.runc.v2"
+
+						  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia.options]
+							BinaryName = "/usr/bin/runc"
+				`
+
+				require.NoError(t, os.WriteFile(configPath, []byte(contents), 0755))
 				return nil
 			},
 			assertSetupPostConditions: func(t *testing.T, testRoot string) error {

```

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,577 @@
...
+				if _, err := os.Stat(configPath); err == nil {
+					// TODO(@ArangoGutierrez): When moving to DropIn files, we should verify that the config file is deleted after cleanup.
+					// For now, we just verify that the runtimes are absent.
+					verifyRuntimesAbsent(t, configPath, "nvidia", "nvidia-cdi", "nvidia-legacy")
+				}
+				return nil
+			},
+		},
+		{
+			description:   "v2: existing config without nvidia runtime and CDI enabled",
```
elezar: The case where `runc` or a default runtime is defined in the config is quite a different one to no default being defined. We may want to distinguish between these.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,577 @@
...
+				require.NoError(t, err)
+				return nil
+			},
+			assertSetupPostConditions: func(t *testing.T, testRoot string) error {
+				configPath := filepath.Join(testRoot, "etc/containerd/config.toml")
+				require.FileExists(t, configPath)
+				verifyRuntimesPresent(t, configPath, "nvidia", "nvidia-cdi", "nvidia-legacy")
+				verifyCDIEnabled(t, configPath, true)
+				return nil
+			},
```
elezar: Another motivation for using RAW STRINGS (i.e. full config strings) here is that we need to ensure that:
1. We do not REMOVE existing config options.
2. We do NOT override the default. (we don't check the default state here).
3. We correctly carry over the runtime options from the default, runc, or crun.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_v2_test.go
elezar: It could be that on the github runner we're generating the config from `containerd config dump`. This should be avoided unless we explicitly "mock" this command.

ArangoGutierrez: Agree, so given this, do you think we should have that as a follow-up, and on this PR assume the file is not empty?

elezar: No. We should check the file contents now. Note that if we make the following change:
```
diff --git a/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go b/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
index 5847d86f..62dc4bb4 100644
--- a/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
@@ -52,11 +52,12 @@ func TestContainerdConfigLifecycle(t *testing.T) {
 			description:   "v2: top-level config does not exist",
 			configVersion: 2,
 			containerOptions: container.Options{
-				Config:       "{{ .testRoot }}/etc/containerd/config.toml",
-				RuntimeName:  "nvidia",
-				RuntimeDir:   "/usr/bin",
-				SetAsDefault: false,
-				RestartMode:  "none",
+				Config:         "{{ .testRoot }}/etc/containerd/config.toml",
+				RuntimeName:    "nvidia",
+				RuntimeDir:     "/usr/bin",
+				SetAsDefault:   false,
+				RestartMode:    "none",
+				ExecutablePath: "not-containerd",
 			},
 			options: Options{},
 			assertSetupPostConditions: func(t *testing.T, testRoot string) error {
```
then we would force the `command-line` source to fail since `not-containerd` will not exist.

The alternative is to get #1251 in for more explicit control.

ArangoGutierrez: ok, I have added `ExecutablePath` and it works now

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
ArangoGutierrez: Added two extra cases to cover this

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
ArangoGutierrez: test now with raw strings

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
```diff
@@ -0,0 +1,451 @@
...
+				RestartMode:  "none",
+			},
+			options: Options{
+				configMode: "config",
+			},
+			prepareEnvironment: func(t *testing.T, testRoot string) error {
+				configPath := filepath.Join(testRoot, "etc/crio/crio.conf")
+				require.NoError(t, os.MkdirAll(filepath.Dir(configPath), 0755))
+
+				cfg, err := crio.New(
```
elezar: Please update the crio tests to use raw strings too.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+
+          [plugins.cri.containerd.runtimes.runc.options]
+            Runtime = "/usr/bin/runc"
+`
+				require.Equal(t, expected, string(actual))
+				return nil
+			},
+		},
+		{
+			description:     "v1 legacy: existing config with default runtime",
```
elezar: One of the things that we should test in this case is that the OPTIONS from the default runtime are inherrited. The test as implemented does not do this.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+				initialConfig := `version = 1
+
+[plugins]
+  [plugins.cri]
+    [plugins.cri.containerd]
+      [plugins.cri.containerd.default_runtime]
+        runtime_type = "io.containerd.runtime.v1.linux"
+
+        [plugins.cri.containerd.default_runtime.options]
+          Runtime = "/usr/bin/runc"
```
elezar: In the context of v1 configs, there are two ways to specify the default runtime: 
1. as an explicit `default_runtime` section (which is deprecated)
2. as a reference to a defined runtime under `plugins.cri.containerd.default_runtime_name`

Even in the v1 config we don't handle the first case consistently (or we do to update the spec, but don't when reading the spec). Let's focus on 2. since this also matches how v2 configs are handled and allows us to further simplify things.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+
+[plugins]
+
+  [plugins.cri]
+
+    [plugins.cri.containerd]
+
+      [plugins.cri.containerd.runtimes]
+
+        [plugins.cri.containerd.runtimes.runc]
```
elezar, line 561: Just a note: This is an example of where we don't handle the deprecated default runtime consistently since the pre and post configs are not the same -- or even equivalent.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+  [plugins."io.containerd.grpc.v1.cri"]
+    [plugins."io.containerd.grpc.v1.cri".containerd]
+      default_runtime_name = "runc"
+
+      [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
+        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+          runtime_type = "io.containerd.runc.v2"
+
+          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+            BinaryName = "/usr/bin/runc"
```
elezar, line 670: Once again it would be good to test that we're pulling in the options from `runc` (or the default runtime).

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+`
+				require.Equal(t, expected, string(actual))
+				return nil
+			},
+			assertCleanupPostConditions: func(t *testing.T, testRoot string) error {
+				configPath := filepath.Join(testRoot, "etc/containerd/config.toml")
+
+				if _, err := os.Stat(configPath); os.IsNotExist(err) {
+					return nil
+				}
```
elezar: The cleanup should NOT delete the file. We should always be able to read it.
```suggestion
				require.FileExists(t, configPath)
```

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+
+				// If file exists, verify content
```
elezar:
```suggestion
```

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+				SetAsDefault:   true,
+				RestartMode:    "none",
+				ExecutablePath: "not-containerd",
+			},
+			options: Options{},
+			prepareEnvironment: func(t *testing.T, testRoot string) error {
+				configPath := filepath.Join(testRoot, "etc/containerd/config.toml")
+				require.NoError(t, os.MkdirAll(filepath.Dir(configPath), 0755))
+
+				// Write initial config with nvidia runtime using old path
```
elezar: Let's rather pull this into the description somehow.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+`
+				require.Equal(t, expected, string(actual))
+				return nil
+			},
+			assertCleanupPostConditions: func(t *testing.T, testRoot string) error {
+				configPath := filepath.Join(testRoot, "etc/containerd/config.toml")
+
+				if _, err := os.Stat(configPath); os.IsNotExist(err) {
+					return nil
+				}
```
elezar: No: The file *must* exist:
```suggestion
				require.FileExists(t, configPath)
```

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+
+				// If file exists, verify nvidia runtimes were removed
+				actual, err := os.ReadFile(configPath)
+				require.NoError(t, err)
+				contentStr := string(actual)
+
+				require.NotContains(t, contentStr, "nvidia")
+				require.NotContains(t, contentStr, "nvidia-cdi")
+				require.NotContains(t, contentStr, "nvidia-legacy")
```
elezar, line 848: Why are we not using a full config string here consistently?

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+		},
+		{
+			description:   "v2: complex config with multiple plugins and settings",
+			configVersion: 2,
+			containerOptions: container.Options{
+				Config:         "{{ .testRoot }}/etc/containerd/config.toml",
+				RuntimeName:    "nvidia",
+				RuntimeDir:     "/usr/bin",
+				EnableCDI:      true,
+				SetAsDefault:   false,
```
elezar, line 859: We probably want the same test with `SetAsDefault = true` as well.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
elezar: This is done multiple times. Please update them all.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+  [plugins."io.containerd.grpc.v1.cri"]
+    [plugins."io.containerd.grpc.v1.cri".containerd]
+      # Note: No default_runtime_name specified
+
+      [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
+        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+          runtime_type = "io.containerd.runc.v2"
+
+          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+            BinaryName = "/usr/bin/runc"
```
elezar, line 1024: Once again, we're actually interested in the runtime options from the `default_runtime_name`, `runc`, and `crun`.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+				configPath := filepath.Join(testRoot, "etc/containerd/config.toml")
+				require.NoError(t, os.MkdirAll(filepath.Dir(configPath), 0755))
+
+				// Write initial config without any default runtime
+				initialConfig := `version = 2
+
+[plugins]
+  [plugins."io.containerd.grpc.v1.cri"]
+    [plugins."io.containerd.grpc.v1.cri".containerd]
+      # Note: No default_runtime_name specified
```
elezar, line 1130:
```suggestion
```

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+          runtime_type = "io.containerd.runc.v2"
+
+          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+            BinaryName = "/usr/bin/runc"
+`
+				require.Equal(t, expected, string(actual))
+
+				// Verify that default_runtime_name WAS added
+				require.Contains(t, string(actual), `default_runtime_name = "nvidia"`)
```
elezar: We are already doing a FULL STRING MATCH on the config. Why is this required?
```suggestion
```

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+          runtime_type = "io.containerd.runc.v2"
+
+          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+            BinaryName = "/usr/bin/runc"
+`
+				require.Equal(t, expected, string(actual))
+
+				// Verify that default_runtime_name was removed to restore original state
+				require.NotContains(t, string(actual), "default_runtime_name")
+				require.NotContains(t, string(actual), "nvidia")
```
elezar: We are already doing a full string match. Why is this required?
```suggestion
				// Verify that default_runtime_name was removed to restore original state
				require.NotContains(t, string(actual), "default_runtime_name")
				require.NotContains(t, string(actual), "nvidia")
```

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+
+// TestContainerdConfigLifecycle tests the complete Setup->Cleanup lifecycle for both v1 and v2 configs.
+func TestContainerdConfigLifecycle(t *testing.T) {
+	c := &cli.Command{
+		Name: "test",
+	}
+
+	testCases := []struct {
+		description                 string
+		configVersion               int  // 0 for auto-detect, 1 for v1, 2 for v2
```
elezar: What about v3?

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+		useLegacyConfig             bool // for v1 legacy behavior
```
elezar: Do we need an additional setting in this struct if we're just setting `options.useLegacyConfig` with this? We only need this for the deprecated v1 config structure, so having it a top-level test var seems out of place.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,978 @@
...
+				require.NoError(t, tc.prepareEnvironment(t, testRoot))
+			}
+
+			// Apply version-specific configuration
+			tc.options.useLegacyConfig = tc.useLegacyConfig
+
+			// Set default runtime type if not already set
+			if tc.options.runtimeType == "" {
+				tc.options.runtimeType = "io.containerd.runc.v2"
+			}
```
elezar: Question: Do we have any tests where we actually set this?

elezar: Also as a nit: move the setup to before we actually start running things.

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
```diff
@@ -0,0 +1,606 @@
...
+				configMode: "config",
+			},
+			assertSetupPostConditions: func(t *testing.T, testRoot string) error {
+				configPath := filepath.Join(testRoot, "etc/crio/crio.conf")
+				require.FileExists(t, configPath)
+
+				actual, err := os.ReadFile(configPath)
+				require.NoError(t, err)
+
+				expected := `
```
elezar, line 66: Do we really insert a newline at the start of the file?

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
```diff
@@ -0,0 +1,1496 @@
...
+		options                     Options
+		prepareEnvironment          func(*testing.T, string) error
+		expectedSetupError          error
+		assertSetupPostConditions   func(*testing.T, string) error
+		expectedCleanupError        error
+		assertCleanupPostConditions func(*testing.T, string) error
+	}{
+		// V1 test cases
+		// Note: We don't test "v1: top-level config does not exist" because new configs
+		// are always created as v2 configs by the containerd package
```
elezar, line 49: Is this true even if the version is explicitly set to 1?

OK. The issue is that the `configVersion` in the `testCases` struct is never used when creating the config object. 

I think this is OK given the current behaviour, so let's remove the `configVersion` instead.

cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
elezar: Nit: we can remove this comment in the file since we already have a code comment above.

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
```diff
@@ -0,0 +1,606 @@
...
+				hookFilename: "99-nvidia.json",
+			},
+			assertSetupPostConditions: func(t *testing.T, testRoot string) error {
+				hookPath := filepath.Join(testRoot, "etc/crio/hooks.d/99-nvidia.json")
+				require.FileExists(t, hookPath)
+
+				actual, err := os.ReadFile(hookPath)
+				require.NoError(t, err)
+
+				expectedBinary := filepath.Join("/usr/bin", config.NVIDIAContainerRuntimeHookExecutable)
```
elezar:
```suggestion
				expectedContents := filepath.Join("/usr/bin", config.NVIDIAContainerRuntimeHookExecutable)
```

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
```diff
@@ -0,0 +1,606 @@
...
+    "env": [
+      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+    ]
+  },
+  "when": {
+    "always": true,
+    "commands": [".*"]
+  },
+  "stages": ["prestart"]
+}`, expectedBinary)
```
elezar:
```suggestion
}`, expectedContents)
```

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
```diff
@@ -0,0 +1,606 @@
...
+
+				// Parse both JSONs to compare them semantically (ignoring formatting differences)
+				var expectedJSON, actualJSON interface{}
+				require.NoError(t, json.Unmarshal([]byte(expected), &expectedJSON))
+				require.NoError(t, json.Unmarshal(actual, &actualJSON))
+				require.Equal(t, expectedJSON, actualJSON)
```
elezar: This is NOT required. We're already checking for exactly matching contents. Is there a reason to expect that these would unmarshal differently?

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
```diff
@@ -0,0 +1,606 @@
...
```
elezar:
```suggestion
```

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
elezar:
```suggestion
```

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
```diff
@@ -0,0 +1,606 @@
...
+			tc.options.hooksDir = strings.ReplaceAll(tc.options.hooksDir, "{{ .testRoot }}", testRoot)
+
+			// Prepare the test environment
+			if tc.prepareEnvironment != nil {
+				require.NoError(t, tc.prepareEnvironment(t, testRoot))
+			} else if tc.options.configMode == "hook" {
+				// Ensure hooks directory exists for hook mode
+				hooksDir := tc.options.hooksDir
+				require.NoError(t, os.MkdirAll(hooksDir, 0755))
+			}
```
elezar: Why is this not just implemented in `prepareEnvironment`? Also, we explicitly create the hook dir as part of the installation, so this should not be reqiured in any case.

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
ArangoGutierrez: Yes, if not the test fails with ` Error:          Not equal: `

cmd/nvidia-ctk-installer/container/runtime/crio/config_test.go
elezar: OK. I see what happened here. I thought this was the contents, but it's just the binary path. Note that since we're refering to a constant here, I would rather avoid callig `fmt.Sprintf` to construct the expected string. I have updated.

PR #1302, opened 2025-09-19
[no-relnote] implement defaultRuntimesGetter interface

ArangoGutierrez:
This aims to further reduce the diff in https://github.com/NVIDIA/nvidia-container-toolkit/pull/1280


elezar:
Since #1280 has already been merged, I don't think this is required anymore.

PR #1308, opened 2025-09-22
Fix handling of unrecognised CDI hooks

### LINKED ISSUE #1225: CLI flags should be ignored when an unrecognised hook is detected
In #1000 we tried to avoid failures for unsupported hooks (to allow some version skew between toolkit versions). This does not work as expected beacaus the flags passed to the unexepected hook cause an error:

```
$ nvidia-cdi-hook create-soname-symlinks --folder foo
Incorrect Usage: flag provided but not defined: -folder

NAME:
   NVIDIA CDI Hook - Command to structure files for usage inside a container, called as hooks from a container runtime, defined in a CDI yaml file

USAGE:
   NVIDIA CDI Hook [global options] [command [command options]]

VERSION:
   1.18.0-rc.2
commit: 08b3a388e7b1d447e10d4c4d4a71dca29a98a964

COMMANDS:
   update-ldcache                    Update ldcache in a container by running ldconfig
   create-symlinks                   A hook to create symlinks in the container.
   chmod                             Set the permissions of folders in the container by running chmod. The container root is prefixed to the specified paths.
   enable-cuda-compat                This hook ensures that the folder containing the CUDA compat libraries is added to the ldconfig search path if required.
   disable-device-node-modification  Ensure that the /proc/driver/nvidia/params file present in the container does not allow device node modifications.
   help, h                           Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --debug, -d    Enable debug-level logging (default: false) [$NVIDIA_CTK_DEBUG, $NVIDIA_CDI_DEBUG]
   --quiet        Suppress all output except for errors; overrides --debug (default: false) [$NVIDIA_CTK_QUIET, $NVIDIA_CDI_QUIET]
   --help, -h     show help
   --version, -v  print the version
ERRO[0000] flag provided but not defined: -folder
```

Should give:
```
$ nvidia-cdi-hook create-soname-symlinks
WARN[0000] Unsupported CDI hook: create-soname-symlinks
```

ArangoGutierrez:
Fixes #1225

Previously, when an unrecognized CDI hook was invoked with command-line flags, the nvidia-cdi-hook command would fail with a flag parsing error instead of gracefully handling the unsupported hook. This could cause container launch failures when CDI specifications reference hooks that are not yet supported by the current NVIDIA Container Toolkit version.

This PR enhances the error handling in nvidia-cdi-hook to gracefully handle unrecognized CDI hooks that include command-line flags. The changes ensure that:
Unrecognized hooks with flags are handled gracefully - Instead of failing with flag parsing errors, the command now issues a warning and continues
Backwards compatibility is maintained - Existing behavior for supported hooks remains unchanged
Container launches don't fail - Unsupported hooks result in warnings rather than errors
Changes
Added OnUsageError handler in ConfigureCDIHookCommand to detect flag parsing errors for unrecognized commands
Added strings import for error message parsing to identify flag-related errors
Enhanced error detection to specifically catch "flag provided but not defined" errors for unrecognized hooks
Improved warning consistency by reusing the existing IssueUnsupportedHookWarning logic

# Test Cases
The following test cases demonstrate the improved behavior:

Before (failing):

```bash
$ ./nvidia-cdi-hook unknown-hook --some-flag value
Error: flag provided but not defined: -some-flag
```

After (graceful handling):

```bash
$ ./nvidia-cdi-hook unknown-hook --some-flag value
time="2025-09-22T11:25:26+02:00" level=warning msg="Unsupported CDI hook: unknown-hook"
```

Additional test cases:

```bash
# Unrecognized hook without flags
$ ./nvidia-cdi-hook unknown-hook
time="2025-09-22T11:25:26+02:00" level=warning msg="Unsupported CDI hook: unknown-hook"

# No hook specified
$ ./nvidia-cdi-hook
time="2025-09-22T11:25:26+02:00" level=warning msg="No CDI hook specified"

# Recognized hook with valid flags (unchanged behavior)
$ ./nvidia-cdi-hook update-ldcache --debug
# Works as expected, no warning

# Recognized hook with invalid flags (unchanged behavior)
$ ./nvidia-cdi-hook update-ldcache --invalid-flag
Error: flag provided but not defined: -invalid-flag
```

### DIFF
--- a/cmd/nvidia-cdi-hook/commands/commands.go
+++ b/cmd/nvidia-cdi-hook/commands/commands.go
@@ -17,6 +17,8 @@
 package commands
 
 import (
+	"context"
+
 	"github.com/urfave/cli/v3"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-cdi-hook/chmod"
@@ -27,27 +29,43 @@ import (
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
 )
 
-// New creates the commands associated with supported CDI hooks.
-// These are shared by the nvidia-cdi-hook and nvidia-ctk hook commands.
-func New(logger logger.Interface) []*cli.Command {
-	return []*cli.Command{
+// ConfigureCDIHookCommand configures a base command with supported CDI hooks
+// and error handling for unsupported hooks.
+// This allows the same command to be used for the nvidia-cdi-hook and
+// nvidia-ctk hook commands.
+func ConfigureCDIHookCommand(logger logger.Interface, cmd *cli.Command) *cli.Command {
+	// We set the default action for the command to issue a warning and exit
+	// with no error.
+	// This means that if an unsupported hook is run, a container will not fail
+	// to launch. An unsupported hook could be the result of a CDI specification
+	// referring to a new hook that is not yet supported by an older NVIDIA
+	// Container Toolkit version or a hook that has been removed in newer
+	// version.
+	cmd.Action = func(ctx context.Context, cmd *cli.Command) error {
+		return issueUnsupportedHookWarning(logger, cmd)
+	}
+	// Define the subcommands
+	cmd.Commands = []*cli.Command{
 		ldcache.NewCommand(logger),
 		symlinks.NewCommand(logger),
 		chmod.NewCommand(logger),
 		cudacompat.NewCommand(logger),
 		disabledevicenodemodification.NewCommand(logger),
 	}
+
+	return cmd
 }
 
-// IssueUnsupportedHookWarning logs a warning that no hook or an unsupported
+// issueUnsupportedHookWarning logs a warning that no hook or an unsupported
 // hook has been specified.
 // This happens if a subcommand is provided that does not match one of the
 // subcommands that has been explicitly specified.
-func IssueUnsupportedHookWarning(logger logger.Interface, c *cli.Command) {
+func issueUnsupportedHookWarning(logger logger.Interface, c *cli.Command) error {
 	args := c.Args().Slice()
 	if len(args) == 0 {
 		logger.Warningf("No CDI hook specified")
 	} else {
 		logger.Warningf("Unsupported CDI hook: %v", args[0])
 	}
+	return nil
 }
--- a/cmd/nvidia-cdi-hook/commands/commands.go
+++ b/cmd/nvidia-cdi-hook/commands/commands.go
@@ -18,6 +18,7 @@ package commands
 
 import (
 	"context"
+	"strings"
 
 	"github.com/urfave/cli/v3"
 
@@ -33,7 +34,7 @@ import (
 // and error handling for unsupported hooks.
 // This allows the same command to be used for the nvidia-cdi-hook and
 // nvidia-ctk hook commands.
-func ConfigureCDIHookCommand(logger logger.Interface, cmd *cli.Command) *cli.Command {
+func ConfigureCDIHookCommand(logger logger.Interface, base *cli.Command) *cli.Command {
 	// We set the default action for the command to issue a warning and exit
 	// with no error.
 	// This means that if an unsupported hook is run, a container will not fail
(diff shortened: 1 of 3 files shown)


ArangoGutierrez:
Tested 

```bash
./nvidia-cdi-hook unknown-hook --some-flag value 
time="2025-09-22T11:25:26+02:00" level=warning msg="Unsupported CDI hook: unknown-hook"
```

cmd/nvidia-cdi-hook/main.go
```diff
@@ -72,6 +73,26 @@ func main() {
 			commands.IssueUnsupportedHookWarning(logger, cmd)
 			return nil
 		},
+		// Handle usage errors to catch unrecognized hooks with flags
+		OnUsageError: func(ctx context.Context, cmd *cli.Command, err error, isSubcommand bool) error {
+			// Check if this is a "flag provided but not defined" error
+			errMsg := err.Error()
+			if strings.HasPrefix(errMsg, "flag provided but not defined: ") {
```
elezar: One nit: The err in `urfave` is defined as:
```
providedButNotDefinedErrMsg = "flag provided but not defined: -"
```
and we don't include the `-` here.

elezar (dismissed):
What about the case of a subcommand that does not accept arguments:
```
$ ./nvidia-cdi-hook foo
No help topic for 'foo'
$ echo $?
3
```

cmd/nvidia-cdi-hook/main.go
ArangoGutierrez: ah yeah, let me edit!

ArangoGutierrez: edited

cmd/nvidia-cdi-hook/main.go
```diff
@@ -72,6 +73,35 @@ func main() {
 			commands.IssueUnsupportedHookWarning(logger, cmd)
 			return nil
 		},
+		// Handle usage errors to catch unrecognized hooks
+		OnUsageError: func(ctx context.Context, cmd *cli.Command, err error, isSubcommand bool) error {
+			errMsg := err.Error()
+
+			// Check if this is a "No help topic for" error (unrecognized
```
elezar: Since in Urfave we have a [`CommandNotFound` handler](https://github.com/urfave/cli/blob/de9b5b4e61b0cce132f2a2bc73261eb0601d76ce/help.go#L318-L329) does it not make sense to handle this separately?

cmd/nvidia-ctk/hook/hook.go
```diff
@@ -54,6 +55,29 @@ func (m hookCommand) build() *cli.Command {
 			commands.IssueUnsupportedHookWarning(m.logger, cmd)
 			return nil
 		},
+		// Handle unrecognized commands when help is requested (e.g., help
+		// unknowncommand)
+		CommandNotFound: func(ctx context.Context, cmd *cli.Command, commandName string) {
```
elezar: These and the `Action` are now strictly duplicated from `nvidia-cdi-hook`. I don't think this is required with `urfave/v3`. Do we want to refactor this?

ArangoGutierrez: I did a small refactor, creating a shared function so we don't have duplicated code

cmd/nvidia-cdi-hook/commands/commands.go
```diff
@@ -51,3 +54,45 @@ func IssueUnsupportedHookWarning(logger logger.Interface, c *cli.Command) {
 		logger.Warningf("Unsupported CDI hook: %v", args[0])
 	}
 }
+
+// NewHookCommand creates a new hook command with common behavior for handling
```
elezar: Nit: This is specifically CDI hooks, so maybe rename `NewCDIHookCommand`?

cmd/nvidia-cdi-hook/commands/commands.go
```diff
@@ -51,3 +54,45 @@ func IssueUnsupportedHookWarning(logger logger.Interface, c *cli.Command) {
...
+		// older NVIDIA Container Toolkit version or a hook that has been removed
+		// in newer version.
+		Action: func(ctx context.Context, cmd *cli.Command) error {
+			IssueUnsupportedHookWarning(logger, cmd)
+			return nil
+		},
+		// Handle unrecognized commands when help is requested (e.g., help
+		// unknowncommand)
+		CommandNotFound: func(ctx context.Context, cmd *cli.Command, commandName string) {
+			logger.Warningf("Unsupported CDI hook: %v", commandName)
```
elezar: Question: Is there a reason that we don't call `IssueUnsupportedHookWarning` (or at least reuse some of the logic there)?

cmd/nvidia-cdi-hook/commands/commands.go
```diff
@@ -51,3 +54,45 @@ func IssueUnsupportedHookWarning(logger logger.Interface, c *cli.Command) {
...
+			// If this is a flag parsing error for an unrecognized command,
+			// suppress it and let the default Action handle it
+			errMsg := err.Error()
+			if strings.HasPrefix(errMsg, "flag provided but not defined: -") {
+				// Check if the first argument is an unrecognized command
+				args := cmd.Args().Slice()
+				if len(args) > 0 && cmd.Command(args[0]) == nil {
+					// This is an unrecognized hook with flags - the default
+					// Action will handle it
+					return nil
```
elezar: Should we not issue a warning here?

cmd/nvidia-cdi-hook/main.go
```diff
@@ -44,52 +44,37 @@ func main() {
 	// Create a options struct to hold the parsed environment variables or command line flags
 	opts := options{}
 
-	// Create the top-level CLI
-	c := cli.Command{
```
elezar, line 48: Does it make sense to still construct the `Command`s and then call a fuction to modify the members that we want instead?

cmd/nvidia-cdi-hook/commands/commands.go
```diff
@@ -51,3 +54,47 @@ func IssueUnsupportedHookWarning(logger logger.Interface, c *cli.Command) {
 		logger.Warningf("Unsupported CDI hook: %v", args[0])
 	}
 }
+
+// ConfigureCDIHookCommand takes a base command and adds the common CDI hook
+// behavior for handling unsupported hooks.
+// The base command should have Name, Usage, and Version set as desired.
+func ConfigureCDIHookCommand(base *cli.Command, logger logger.Interface) *cli.Command {
```
elezar: The comment seems to indicate that we should call the functions something like `AddUnsupportedHooksChecks` but we do add the subcommands at the end, so it's probably just the comment that needs to be udpated for accuracy.

cmd/nvidia-cdi-hook/main.go
```diff
@@ -44,54 +44,41 @@ func main() {
...
-		// to launch. An unsupported hook could be the result of a CDI specification
-		// referring to a new hook that is not yet supported by an older NVIDIA
-		// Container Toolkit version or a hook that has been removed in newer
-		// version.
-		Action: func(ctx context.Context, cmd *cli.Command) error {
-			commands.IssueUnsupportedHookWarning(logger, cmd)
-			return nil
+	}
+	// Set log-level for all subcommands
+	c.Before = func(ctx context.Context, cmd *cli.Command) (context.Context, error) {
```
elezar: Givent that we're only ADDing to the hook in the new function, I would not expect this to be part of the diff.

elezar:
Let's update the description to show the output for the various test cases.

cmd/nvidia-cdi-hook/commands/commands.go
```diff
@@ -51,3 +54,46 @@ func IssueUnsupportedHookWarning(logger logger.Interface, c *cli.Command) {
 		logger.Warningf("Unsupported CDI hook: %v", args[0])
 	}
 }
+
+// ConfigureCDIHookCommand takes a base command and configures it to handle
+// unsupported CDI hooks gracefully.
```
elezar: This is still not accurate.

ArangoGutierrez: How about now?

cmd/nvidia-cdi-hook/main.go
ArangoGutierrez: The diff has been updated

cmd/nvidia-cdi-hook/commands/commands.go
ArangoGutierrez: Func doc comment has been updated

elezar (approved):

PR #1311, opened 2025-09-22
Add CRI plugin config from source containerd config to drop-in file

cdesiniotis:
This change updates the drop-in file contents that the nvidia-ctk-installer creates for containerd. In addition to adding nvidia runtimes, the nvidia-ctk-installer also includes all existing configuration present in the source containerd configuration for the CRI plugin. That is, all configuration already present in the 'plugins."io.containerd.grpc.v1.cri"' section will be added to our drop-in file.

This is needed due to how containerd merges configuration from multiple files. See https://github.com/containerd/containerd/issues/5837

### DIFF
--- a/cmd/nvidia-ctk-installer/container/container.go
+++ b/cmd/nvidia-ctk-installer/container/container.go
@@ -17,6 +17,7 @@
 package container
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -69,7 +70,23 @@ func (o Options) Unconfigure(cfg engine.Interface) error {
 	if err != nil {
 		return fmt.Errorf("unable to update config: %v", err)
 	}
-	return o.flush(cfg)
+
+	if err := o.flush(cfg); err != nil {
+		return err
+	}
+
+	if o.DropInConfig == "" {
+		return nil
+	}
+	// When a drop-in config is used, we remove the drop-in file explicitly.
+	// This is require for cases where we may have to include other contents
+	// in the drop-in file and as such it may not be empty when we flush it.
+	err = os.Remove(o.DropInConfig)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("failed to remove drop-in config file: %w", err)
+	}
+
+	return nil
 }
 
 // flush flushes the specified config to disk
--- a/pkg/config/engine/containerd/containerd.go
+++ b/pkg/config/engine/containerd/containerd.go
@@ -133,10 +133,15 @@ func New(opts ...Option) (engine.Interface, error) {
 		}
 		dropInConfig := &engine.Config{
 			Source: sourceConfig,
-			// The destinationConfig is an empty config with the same options
-			// as the source config.
+			// The destinationConfig is a minimal config with the same options
+			// as the source config. The starting content of the destinationConfig
+			// is the entire plugins."io.containerd.grpc.v1.cri" section of the source
+			// config. This is needed due to how containerd merges configuration from
+			// multiple files. In particular, plugins are overridden by key and the
+			// content is not merged. This issue is fixed starting in containerd 2.1
+			// Reference: https://github.com/containerd/containerd/issues/5837
 			Destination: &Config{
-				Tree:          toml.NewEmpty(),
+				Tree:          getBaseDropInConfigTree(sourceConfig),
 				configOptions: sourceConfigOptions,
 			},
 		}
@@ -203,4 +208,18 @@ func chrootIfRequired(hostRoot string, commandLine ...string) []string {
 	}
 
 	return append([]string{"chroot", hostRoot}, commandLine...)
+}
+
+// getBaseDropInConfigTree returns the base config to use for the drop-in config file.
+// For older containerd versions (e.g. 1.7) the plugins section of multiple
+// configs are not fully merged, but merged by key instead. This means that we
+// need to duplicate the entire plugin-specific config in the drop in file so as
+// to not override settings applied in the base config.
+func getBaseDropInConfigTree(sourceConfig *Config) *toml.Tree {
+	baseDropInConfigTree := toml.NewEmpty()
+	criPlugin := sourceConfig.GetSubtreeByPath([]string{"plugins", sourceConfig.CRIRuntimePluginName})
+	if criPlugin != nil {
+		baseDropInConfigTree.SetPath([]string{"plugins", sourceConfig.CRIRuntimePluginName}, criPlugin)
+	}
+	return baseDropInConfigTree
 }
--- a/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
@@ -789,6 +789,7 @@ version = 2
     enable_cdi = true
 
     [plugins."io.containerd.grpc.v1.cri".containerd]
+      default_runtime_name = "runc"
 
       [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
 
@@ -809,6 +810,12 @@ version = 2
 
           [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia-legacy.options]
             BinaryName = "/usr/bin/nvidia-container-runtime.legacy"
+
+        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+          runtime_type = "io.containerd.runc.v2"
+
+          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+            BinaryName = "/usr/bin/runc"
 `
 				require.Equal(t, expectedDropIn, string(actualDropIn))
 				return nil
@@ -954,6 +961,12 @@ version = 2
 
           [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia-legacy.options]
             BinaryName = "/usr/bin/nvidia-container-runtime.legacy"
+
+        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+          runtime_type = "io.containerd.runc.v2"
+
+          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+            BinaryName = "/usr/bin/runc"
 `
 				require.Equal(t, expectedDropIn, string(actualDropIn))
 				return nil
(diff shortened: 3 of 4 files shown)


pkg/config/engine/containerd/containerd.go
```diff
@@ -204,3 +208,12 @@ func chrootIfRequired(hostRoot string, commandLine ...string) []string {
 
 	return append([]string{"chroot", hostRoot}, commandLine...)
 }
+
+func getBaseDropInConfigTree(sourceConfig *Config) *toml.Tree {
+	baseDropInConfigTree := toml.NewEmpty()
+	runtimes := sourceConfig.GetSubtreeByPath([]string{"plugins", sourceConfig.CRIRuntimePluginName, "containerd", "runtimes"})
```
elezar: I don't think this is sufficient. Note that plugins are *overwritten* by key https://github.com/containerd/containerd/blob/24b9e2c1a0a72a7ad302cdce7da3abbc4e6295cb/services/server/config/config.go#L308-L310

meaning that the entire content of `plugins[sourceConfig.CRIRuntimePluginName]` will be pulled in from the imported configs.

pkg/config/engine/containerd/containerd.go
```diff
@@ -133,10 +133,14 @@ func New(opts ...Option) (engine.Interface, error) {
...
 		dropInConfig := &engine.Config{
 			Source: sourceConfig,
-			// The destinationConfig is an empty config with the same options
-			// as the source config.
+			// The destinationConfig is a minimal config with the same options
+			// as the source config. The starting content of the destinationConfig
+			// is the entire plugins."io.containerd.grpc.v1.cri".containerd.runtimes
+			// section of the source config. This is needed due to how containerd
+			// merges configuration from multiple files.
+			// Reference: https://github.com/containerd/containerd/issues/5837
```
elezar, line 142: Given that newer containerd versions (v2.1) do seem to do what we want, did we want to make this optional somehow?

pkg/config/engine/containerd/containerd.go
cdesiniotis: You are right. I will update this.

cdesiniotis: Updated.

cdesiniotis:
Initial testing of these changes looks good. Using the GPU Operator (built from https://github.com/NVIDIA/gpu-operator/pull/1710) I have verified several of the tests captured in https://github.com/NVIDIA/nvidia-container-toolkit/blob/b3ce3a4f3855764208b4b14004620d27a947110a/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go.

cmd/nvidia-ctk-installer/container/container.go
```diff
@@ -124,6 +124,10 @@ func (o Options) RevertConfig(cfg engine.Interface) error {
 		}
 	}
 
+	if cfg.UseDropInConfig() {
+		return cfg.CleanupDropInConfig()
+	}
```
cdesiniotis: @elezar note, I remove the nvidia runtimes first before a call to `CleanupDropInConfig()`since the `ConfigWithDropIn` implementation of `RemoveRuntime` removes runtimes from both the top-level config and the drop-in file.

elezar: Thanks. The other option would be to remove the runtimes in the top-level config when ADDING them to the drop in, but that may be more complex than it needs to be.

pkg/config/engine/containerd/config.go
```diff
@@ -154,3 +154,9 @@ func (c *Config) RemoveRuntime(name string) error {
 	*c.Tree = config
 	return nil
 }
+
+func (c *Config) UseDropInConfig() bool {
+	return false
```
elezar: Question: Why is this always `false`? for containerd?

cdesiniotis: It is false for the `Config` implementation, and true for the `ConfigWithDropIn` implemenation.

Thinking about it now, this method is not very useful. I believe we can just call `CleanupDropInConfig` directly and this can be a no-op for all cases except for `ConfigWithDropIn`.

pkg/config/engine/containerd/config_drop_in.go
```diff
@@ -179,3 +180,11 @@ func (c *topLevelConfig) ensureImports(dropInFilename string) {
 func (c *topLevelConfig) RemoveRuntime(name string) error {
 	return c.config.RemoveRuntime(name)
 }
+
+func (c *ConfigWithDropIn) UseDropInConfig() bool {
+	return true
+}
+
+func (c *ConfigWithDropIn) CleanupDropInConfig() {
+	c.Interface = &Config{Tree: toml.NewEmpty()}
```
elezar: Should this not just be:
```suggestion
	c.Tree = toml.NewEmpty()
```

What about the `configOptions` which are used to update the config later? (I understand that we only ever call this AFTER we're done updating the config, but this may not always be the case).

pkg/config/engine/crio/crio.go
```diff
@@ -198,3 +198,11 @@ func chrootIfRequired(hostRoot string, commandLine ...string) []string {
 
 	return append([]string{"chroot", hostRoot}, commandLine...)
 }
+
+func (c *Config) UseDropInConfig() bool {
+	return true
+}
+
+func (c *Config) CleanupDropInConfig() {
+	c.Tree = toml.NewEmpty()
```
elezar: If this works for `crio` then i think we should be able to use it for `containerd` too.

pkg/config/engine/api.go
```diff
@@ -23,6 +23,8 @@ type Interface interface {
 	EnableCDI()
 	GetRuntimeConfig(string) (RuntimeConfig, error)
 	RemoveRuntime(string) error
+	UseDropInConfig() bool
```
elezar: Just as a note to self. We should probably clean up this interface / compose the other two interfaces that we have introduced.

pkg/config/engine/containerd/config_drop_in.go
cdesiniotis: We aren't able to do this without some refactoring. Currently, an `engine.Interface` is a member of `ConfigWithDropIn` and not a `Config`.

pkg/config/engine/containerd/config.go
elezar: Yes. I think the check was only needed if we called this BEFORE we removed the runtimes. Since we're doing that first now, we can just call it unconditionally.

elezar: (and yes. I'm sorry, it's late).

pkg/config/engine/containerd/config_drop_in.go
elezar: Sorry. I think I'm missing some important details here after a very long day. I will have a look again after some rest.

pkg/config/engine/containerd/config.go
cdesiniotis: I have left this as-is for now. Feel free to clean this up further as you see fit.

I spent some time trying to remove the `UseDropInConfig` method entirely (and unconditionally calling `CleanupDropInConfig`), but this did not end up being as straightforward as I would have liked.

cmd/nvidia-ctk-installer/container/container.go
```diff
@@ -69,7 +70,23 @@ func (o Options) Unconfigure(cfg engine.Interface) error {
...
+	if o.DropInConfig == "" {
+		return nil
+	}
+	// When a drop-in config is used, we remove the drop-in file explicitly.
+	// This is require for cases where we may have to include other contents
+	// in the drop-in file and as such it may not be empty when we flush it.
+	err = os.Remove(o.DropInConfig)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("failed to remove drop-in config file: %w", err)
+	}
```
elezar, line 87: @cdesiniotis given the discussions around the additional functions to the interface, I thought it would actually be cleaner to explicitly remove the file here instead of a more complex implementation at a lower level.

I'm going to push out an rc including these changes be we can always revisit if there's something that I missed.

pkg/config/engine/containerd/containerd.go
elezar: I have a local change that I can create a PR from as a follow-up. Let's not overcomplicate this now though.

cmd/nvidia-ctk-installer/container/container.go
ArangoGutierrez, line 87: I think we can simplify this as 
```suggestion
	if o.DropInConfig != "" {
	// When a drop-in config is used, we remove the drop-in file explicitly.
	// This is require for cases where we may have to include other contents
	// in the drop-in file and as such it may not be empty when we flush it.
	err = os.Remove(o.DropInConfig)
	if err != nil && !errors.Is(err, os.ErrNotExist) {
		return fmt.Errorf("failed to remove drop-in config file: %w", err)
	}
```

pkg/config/engine/containerd/config.go
elezar: Could something like https://github.com/NVIDIA/nvidia-container-toolkit/pull/1311#discussion_r2374946387 work? If I recall correctly this is what you proposed initially. This is definitely simpler than trying to "clear" the config after the fact.

cmd/nvidia-ctk-installer/container/container.go
elezar: How does that simplify things? A common pattern in Go is to "quick" return to reduce nesting which is what is employed here.

elezar (approved):

cmd/nvidia-ctk-installer/container/container.go
```diff
@@ -69,7 +70,23 @@ func (o Options) Unconfigure(cfg engine.Interface) error {
...
 	}
-	return o.flush(cfg)
+
+	if err := o.flush(cfg); err != nil {
+		return err
+	}
+
+	if o.DropInConfig == "" {
+		return nil
+	}
```
cdesiniotis, line 80: Question -- if `o.DropInConfig != ""` is it always the case that we are using a drop-in file? Even if this option is specified, my understanding is that a drop-in file would not be used for 1) v1 containerd configs, or 2) cri-o when the `hook` config mode is configured.

cmd/nvidia-ctk-installer/container/container.go
cdesiniotis: Removing the file is definitely cleaner and simplifies implementation. I am onboard as long as we are sure the conditional immediately preceding this is sufficient.

elezar: For use cases where we are modifying the config directly -- or drop ins are not supported -- a user SHOULD be setting `DropInConfig` to `""` explicitly. This would be the case for `docker`, or for v1 containerd configs.

The check for `DropInConfig == ""` also mirrors what we do when deciding whether to flush to the drop-in config or the "regular" config.

Are the situations that you're concerned about?

cdesiniotis: Okay, I see that this is aligned with the code that flushes the config, so I believe this addresses my question. 

> For use cases where we are modifying the config directly -- or drop ins are not supported -- a user SHOULD be setting DropInConfig to "" explicitly. This would be the case for docker, or for v1 containerd configs.

Ah, I wasn't aware that the intention was for users to set `--drop-in-config=""` explicitly when using v1 containerd configs... As long as we cover this in our documentation, this should be fine. Most should be using v2 or v3 configs at this point anyways.

PR #1313, opened 2025-09-23
[no-relnote] Add E2E for containerd

ArangoGutierrez:


ArangoGutierrez:
Builds on #1235

Doesn't include #1311  tests for that should be added as a follow up

ArangoGutierrez:
I'll mark this PR as ready for review once https://github.com/NVIDIA/nvidia-container-toolkit/pull/1235 is merged

ArangoGutierrez:
> I'll mark this PR as ready for review once #1235 is merged

Rebased

ArangoGutierrez:
Rebased

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+	BeforeAll(func(ctx context.Context) {
+		var err error
+		nestedContainerRunner, err = NewNestedContainerRunner(runner, outerContainerImage, installCTK, containerName, localCacheDir)
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	AfterAll(func(ctx context.Context) {
+		// Cleanup: remove the container and the temporary script on the host.
+		// Use || true to ensure cleanup doesn't fail the test
+		runner.Run(fmt.Sprintf("docker rm -f %s 2>/dev/null || true", containerName)) //nolint:errcheck
```
elezar: Instead of the nolint let's just drop the return values.
```suggestion
		_, _, _ = runner.Run(fmt.Sprintf("docker rm -f %s 2>/dev/null || true", containerName))
```
Does it mak sense to at least WARN if the cleanup fails? The `|| true` doesn't ensure that the test doesn't fail, the fact that we don't check the return value does that.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+		_, _, err := nestedContainerRunner.Run(`
+			 # Clean up any previous drop-in configs
+			 rm -rf /etc/containerd/conf.d
+			 rm -rf /etc/containerd/custom.d
+			 mkdir -p /etc/containerd/conf.d
+			 
+			 # Remove any imports line from the config (reset to original state)
+			 if [ -f /etc/containerd/config.toml ]; then
+				 grep -v "^imports = " /etc/containerd/config.toml > /tmp/config.toml.tmp && mv /tmp/config.toml.tmp /etc/containerd/config.toml || true
+			 fi
```
elezar: Why not just make a copy of the original config and restore that after / before each test?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+ 
+			 # Restart containerd to pick up the clean config
+			 systemctl restart containerd
+			 sleep 2
```
elezar: Is there a way to check containerd health?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+
+			_, _, err = installer.Install(nestedContainerRunner)
+			Expect(err).ToNot(HaveOccurred(), "Failed to install toolkit for containerd")
+
+			// Verify the drop-in config was created
+			output, _, err := nestedContainerRunner.Run(`cat /etc/containerd/conf.d/99-nvidia.toml`)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(ContainSubstring(`nvidia`))
+			Expect(output).To(ContainSubstring(`nvidia-cdi`))
+			Expect(output).To(ContainSubstring(`nvidia-legacy`))
```
elezar: As I mentioned in person, we are nolonger triggering the configuration of containerd with the current installation mechanism.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+
+			// Restart containerd to apply config
+			_, _, err = nestedContainerRunner.Run(`systemctl restart containerd`)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Verify containerd loaded the config correctly
+			output, _, err = nestedContainerRunner.Run(`containerd config dump`)
+			Expect(err).ToNot(HaveOccurred())
+			// Verify imports section is in the merged config
+			Expect(output).To(ContainSubstring(`imports = ['/etc/containerd/conf.d/*.toml']`))
```
elezar: I actually think that `config dump` prints that ACTUAL paths of all files processed.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+			// Check for either single or double quotes as containerd v3 uses single quotes
+			Expect(output).To(SatisfyAny(
+				ContainSubstring(`default_runtime_name = "nvidia"`),
+				ContainSubstring(`default_runtime_name = 'nvidia'`),
```
elezar: These should definitely be VERSION specific checks.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+			))
+			Expect(output).To(ContainSubstring(`enable_cdi = true`))
```
elezar: Where do we toggle this behaviour? It is disabled by default.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+			// Verify Kubernetes settings are still present
+			Expect(output).To(ContainSubstring(`SystemdCgroup = true`))
+			// Verify NVIDIA runtime was added
+			Expect(output).To(SatisfyAny(
+				ContainSubstring(`[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia]`),
+				ContainSubstring(`[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.nvidia]`),
```
elezar: Once again, thsi should be version-specific.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+			))
+			Expect(output).To(ContainSubstring(`BinaryName = `)) // The NVIDIA container runtime binary
+		})
+	})
+
+	When("containerd already has a custom default runtime configured", func() {
+		It("should preserve the existing default runtime when --set-as-default=false is specified", func(ctx context.Context) {
```
elezar: `--set-as-default=false` is not specified. It is the default.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+		 [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.runc]
+		   runtime_type = "io.containerd.runc.v2"
+		   
+		   [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.runc.options]
+			 BinaryName = "/usr/bin/runc"
+ EOF
+			 `)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Configure containerd with drop-in config (explicitly not setting as default)
```
elezar: How are we configuring with the drop-in config in this case?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+			Expect(output).To(ContainSubstring(`nvidia`))
+			Expect(output).To(ContainSubstring(`nvidia-cdi`))
+			Expect(output).To(ContainSubstring(`nvidia-legacy`))
+			// CDI should be enabled
+			Expect(output).To(ContainSubstring(`enable_cdi = true`))
+		})
+	})
+
+	When("containerd has multiple custom runtimes and plugins configured", func() {
+		It("should add NVIDIA runtime alongside existing runtimes like kata", func(ctx context.Context) {
```
elezar: As a matter of interest, how is this different to an arbitrarry "custom" runtime?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+   runtime_type = "io.containerd.kata.v2"
+   
+   [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.kata.options]
+	 ConfigPath = "/etc/kata-containers/configuration.toml"
+ EOF
+			 `)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Verify kata runtime was added
+			output, _, err := nestedContainerRunner.Run(`systemctl restart containerd && sleep 2 && containerd config dump | grep -A5 kata`)
```
elezar: Why the `-A5`? Also. Please split the different steps.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(ContainSubstring(`kata`))
+
+			// Configure containerd with drop-in config and set NVIDIA as default
```
elezar: Where are we setting these options?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+	When("using containerd with version 3 configuration format", func() {
+		It("should correctly add NVIDIA runtime to v3 config structure", func(ctx context.Context) {
+			// Create a v3 containerd config
+			_, _, err := nestedContainerRunner.Run(`
+				 rm -f /etc/containerd/config.toml
+				 rm -rf /etc/containerd/conf.d
+				 mkdir -p /etc/containerd/conf.d
+				 
+				 cat > /etc/containerd/config.toml <<'EOF'
+ version = 3
```
elezar: How do we ensure that this is only run on containerd versions that support it?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+			_, _, err := nestedContainerRunner.Run(`
+				 rm -f /etc/containerd/config.toml
+				 rm -rf /etc/containerd/conf.d
+				 rm -rf /etc/containerd/custom.d
+				 mkdir -p /etc/containerd/conf.d
+				 mkdir -p /etc/containerd/custom.d
+				 
+			 # Create a custom config that will be imported
+			 # Use the correct plugin path for containerd v2/v3
+			 cat > /etc/containerd/custom.d/10-custom.toml <<'EOF'
```
elezar: I thought the correct path was: `/etc/containerd/conf.d`? https://github.com/containerd/containerd/pull/12323/files

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+		   runtime_type = "io.containerd.runc.v2"
+		   
+		   [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.runc.options]
+			 BinaryName = "/usr/bin/runc"
+ EOF
+			 `)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Verify containerd can load the custom import before installer
+			_, _, err = nestedContainerRunner.Run(`systemctl restart containerd && sleep 2 && containerd config dump | grep -i myregistry`)
```
elezar: Why are we randomly checking for `myregistry`? Is this sufficient?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+			// Pull a test image
+			output, stderr, err := nestedContainerRunner.Run(`
+				 ctr image pull docker.io/library/busybox:latest
+			 `)
+			Expect(err).ToNot(HaveOccurred(), "Failed to pull image: %s\nSTDERR: %s", output, stderr)
+
+			// Run a container with NVIDIA runtime
+			// Note: We use native snapshotter because overlay doesn't work in nested containers
+			output, stderr, err = nestedContainerRunner.Run(`
+				 ctr run --rm --snapshotter native docker.io/library/busybox:latest test-nvidia echo "Hello from NVIDIA runtime"
```
elezar: How does this use the nvidia runtime?

elezar: Also, if no devices are requested, then the nvidia-runtime is a no-op.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,403 @@
...
+			_, _, err = installer.Install(nestedContainerRunner)
+			Expect(err).ToNot(HaveOccurred(), "Failed to install toolkit for containerd")
+
+			// Restart containerd to apply the new configuration
+			_, _, err = nestedContainerRunner.Run(`systemctl restart containerd`)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Pull a test image
+			output, stderr, err := nestedContainerRunner.Run(`
+				 ctr image pull docker.io/library/busybox:latest
```
elezar: Does ctr automatically use the containerd config?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+
+		// Create a tempdir on the remote runner (not nested container)
+		var tmpdir string
+		tmpdir, _, err = runner.Run("mktemp -d --tmpdir=/tmp nvctk-e2e-test-cacheXXX")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(strings.TrimSpace(tmpdir)).ToNot(BeEmpty())
+		nestedContainerCacheDir = strings.TrimSpace(tmpdir)
+
+		// Prepare cache on the remote runner (where Docker is)
+		installer, err := NewToolkitInstaller(
```
elezar: Is there a reason that we can't use the "globa" `toolkitInstaller`?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+			fi
+			
+			# Backup config.d directory if it exists
+			if [ -d /etc/containerd/config.d ]; then
+				cp -r /etc/containerd/config.d /tmp/containerd-config.d.backup
+			fi
+			
+			# Backup custom.d directory if it exists
+			if [ -d /etc/containerd/custom.d ]; then
+				cp -r /etc/containerd/custom.d /tmp/containerd-custom.d.backup
```
elezar: The containerd source code does not seem to include `custom.d`. Why are we copying this?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+		// Backup original containerd configuration
+		_, _, err = nestedContainerRunner.Run(`
+			# Backup the original config.toml
+			if [ -f /etc/containerd/config.toml ]; then
+				cp /etc/containerd/config.toml /tmp/containerd-config.toml.backup
+			fi
+			
+			# Backup config.d directory if it exists
+			if [ -d /etc/containerd/config.d ]; then
+				cp -r /etc/containerd/config.d /tmp/containerd-config.d.backup
```
elezar: The path that is searched by default is `/etc/containerd/conf.d`. See https://github.com/containerd/containerd/pull/12323

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+		nestedContainerRunner, err = NewNestedContainerRunner(runner, outerContainerImage, installCTK, containerName, nestedContainerCacheDir)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Backup original containerd configuration
+		_, _, err = nestedContainerRunner.Run(`
+			# Backup the original config.toml
+			if [ -f /etc/containerd/config.toml ]; then
+				cp /etc/containerd/config.toml /tmp/containerd-config.toml.backup
+			fi
+
```
elezar: Just as a thought. Does using `-f -r` as arguments for `cp` for all of the things we want to backup make this (and the subsequent cleanup) simpler?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+			rm -rf /etc/containerd/custom.d
+			if [ -d /tmp/containerd-custom.d.backup ]; then
+				cp -r /tmp/containerd-custom.d.backup /etc/containerd/custom.d
+			fi
+			
+			# Restart containerd to apply restored config
+			systemctl restart containerd || true
+		`)
+		if err != nil {
+			GinkgoWriter.Printf("Warning: failed to restore containerd configuration: %v\n", err)
```
elezar: Does `GinkgoLogr` not have an actual `Warning` method?

elezar: Otherwise just use `GinkgoLogr.Error` here?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+		_, _, err = runner.Run(fmt.Sprintf("docker rm -f %s 2>/dev/null || true", containerName))
+		if err != nil {
+			GinkgoWriter.Printf("Warning: failed to cleanup container %s: %v\n", containerName, err)
+		}
+	})
+
+	BeforeEach(func(ctx context.Context) {
+		// Restore from backup to ensure clean state for each test
+		_, _, err := nestedContainerRunner.Run(`
+			# Restore the original config.toml
```
elezar: This seems strange to create the backup in `BeforeAll` and then to restore the backup again directly in `BeforeEach`. Why not restore the backup in `AfterEach` instead? This would mean that we don't need the logic in `AfterAll`?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+				if ctr version >/dev/null 2>&1; then
+					echo "containerd is ready"
+					break
+				fi
+				if [ "$i" -eq 30 ]; then
+					echo "ERROR: containerd failed to become ready"
+					exit 1
+				fi
+				sleep 1
+			done
```
elezar: This should be logically separate from restoring the backups.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+		`)
+		Expect(err).ToNot(HaveOccurred(), "Failed to reset containerd configuration")
+	})
+
+	When("configuring containerd on a Kubernetes node", func() {
```
elezar: nit: We're not on a Kubernetes node.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+		It("should add NVIDIA runtime using drop-in config without modifying the main config", func(ctx context.Context) {
+			// Configure containerd using nvidia-ctk
+			cmd := fmt.Sprintf(nvidiaCtkConfigureContainerdCmd, "--nvidia-set-as-default --cdi.enabled")
```
elezar: Let's definitely just use string concatenation here. I would even go so far as to say that we should just duplicate the command every time we want to run it.

tests/e2e/nvidia-ctk_containerd_test.go
elezar: Having tests that pass in multiple states are prone to be flaky. Please pick the format for the version of containerd that we have installed by default and use that. This makes the differences between SPECIFIC containerd versions more obvious when reading the tests.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+			Expect(output).To(ContainSubstring(`BinaryName = `)) // The NVIDIA container runtime binary
+		})
+	})
+
+	When("containerd already has a custom default runtime configured", func() {
+		It("should preserve the existing default runtime when --set-as-default=false is specified", func(ctx context.Context) {
+			// Create a containerd config with a custom runtime
+			_, _, err := nestedContainerRunner.Run(`
+rm -rf /etc/containerd/config.d
+mkdir -p /etc/containerd/config.d
```
elezar: Why are we removing and creating this folder? Does this not contradict us restoring the backups in `BeforeEach`?

tests/e2e/nvidia-ctk_containerd_test.go
elezar: Note that we don't strictly speaking need a custom runtime to test the behaviour of NOT overriding the default.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+      default_runtime_name = "custom-runtime"
+      
+      [plugins."io.containerd.cri.v1.runtime".containerd.runtimes]
+        [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.custom-runtime]
+          runtime_type = "io.containerd.runc.v2"
+          
+          [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.custom-runtime.options]
+            BinaryName = "/usr/bin/custom-runtime"
+            SystemdCgroup = true
+            CustomOption = "custom-value"
```
elezar: Are theses values relevant? Should we check that the applied config includes these settings for the `nvidia` runtime?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+
+			// Default runtime should still be custom-runtime (handle both quote styles)
+			Expect(output).To(SatisfyAny(
+				ContainSubstring(`default_runtime_name = "custom-runtime"`),
+				ContainSubstring(`default_runtime_name = 'custom-runtime'`),
+			))
+			// Custom runtime should be preserved with all its options
+			Expect(output).To(MatchRegexp(`(?s)custom-runtime.*CustomOption.*custom-value`))
+			// NVIDIA runtime should be added
+			Expect(output).To(ContainSubstring(`nvidia`))
```
elezar: This is a very BROAD check.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+			output, _, err = nestedContainerRunner.Run(`containerd config dump`)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Default runtime should still be custom-runtime (handle both quote styles)
+			Expect(output).To(SatisfyAny(
+				ContainSubstring(`default_runtime_name = "custom-runtime"`),
+				ContainSubstring(`default_runtime_name = 'custom-runtime'`),
+			))
+			// Custom runtime should be preserved with all its options
+			Expect(output).To(MatchRegexp(`(?s)custom-runtime.*CustomOption.*custom-value`))
```
elezar: That's not what this checks. For example the `SystemdCgroups` setting is not checked.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,433 @@
...
+`)
+			Expect(err).ToNot(HaveOccurred())
+
+			// verify kata runtime was added
+			output, _, err := nestedContainerRunner.Run(`containerd config dump`)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(ContainSubstring(`kata`))
+
+			// Configure containerd using nvidia-ctk with nvidia as default
+			cmd := fmt.Sprintf(nvidiaCtkConfigureContainerdCmd, "--nvidia-set-as-default --cdi.enabled")
```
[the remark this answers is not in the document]
elezar: Prefer `--set-as-default` over `--nvidia-set-as-default`.

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: I was creating one during devel. not needed any more

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: Haven't thought on that, now simplified

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: Done

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: You are right, refactored

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: edited

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: Agreed, adopted

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,415 @@
...
+		// Create the nested container with the global cache mounted
+		nestedContainerRunner, err = NewNestedContainerRunner(runner, outerContainerImage, installCTK, containerName, localCacheDir)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Backup original containerd configuration
+		_, _, err = nestedContainerRunner.Run(`
+			# Backup the original conf.d directory
+			if [ -d /etc/containerd/conf.d ]; then
+				cp -r /etc/containerd/conf.d /tmp/containerd-conf.d.backup
+			fi
```
elezar, line 164: What about `/etc/containerd/config.toml` (if it exists)?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,415 @@
...
+		`)
+		Expect(err).ToNot(HaveOccurred(), "Failed to restore containerd configuration")
+
+		// Step 2: Restart containerd and wait for it to be ready
+		err = restartContainerdAndWait(nestedContainerRunner)
+		Expect(err).ToNot(HaveOccurred(), "Failed to restart containerd after configuration restore")
+	})
+
+	When("configuring containerd on a KIND node", func() {
```
elezar: nit: Why reference KIND at all? We only selected the `kind` nodes since this was the simplest way to get `containerd` running in a container.

```suggestion
	When("configuring containerd", func() {
```

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,415 @@
...
+			Expect(err).ToNot(HaveOccurred(), "Failed to restart containerd")
+
+			// Verify containerd loaded the config correctly
+			output, _, err := nestedContainerRunner.Run(`containerd config dump`)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Basic check that nvidia runtime exists
+			Expect(output).To(SatisfyAny(
+				ContainSubstring(`[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia]`),
+				ContainSubstring(`[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.nvidia]`),
```
elezar: As called out a number of times, this check is too broad. We should capture the differences between different containerd versions *explicitly*.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,415 @@
...
+		`)
+		Expect(err).ToNot(HaveOccurred(), "Failed to restore containerd configuration")
+
+		// Step 2: Restart containerd and wait for it to be ready
+		err = restartContainerdAndWait(nestedContainerRunner)
+		Expect(err).ToNot(HaveOccurred(), "Failed to restart containerd after configuration restore")
+	})
+
+	When("configuring containerd on a KIND node", func() {
+		It("should add NVIDIA runtime using drop-in config without modifying the main config", func(ctx context.Context) {
```
elezar: This is only true for containerd versions that INCLUDE the default `/etc/containerd/conf.d` imports.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,415 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			// Basic check that nvidia runtime exists
+			Expect(output).To(SatisfyAny(
+				ContainSubstring(`[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia]`),
+				ContainSubstring(`[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.nvidia]`),
+			))
+
+			// Verify the drop-in config was processed (config dump shows actual imported files)
+			Expect(output).To(ContainSubstring(`/etc/containerd/conf.d/*.toml`))
```
elezar: Can we please load the output config as `toml` and check specific values instead of checking substrings.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,415 @@
...
+			// Check for either single or double quotes as containerd v3 uses single quotes
+			Expect(output).To(SatisfyAny(
+				ContainSubstring(`default_runtime_name = "nvidia"`),
+				ContainSubstring(`default_runtime_name = 'nvidia'`),
+			))
+			Expect(output).To(ContainSubstring(`enable_cdi = true`))
+			// Verify KIND settings are still present
+			Expect(output).To(ContainSubstring(`SystemdCgroup = true`))
```
elezar: How is this a Kind setting? Does the kind node ALWAYS use systemd cgroups?

elezar (commented):
There are still some unanswered questions below. 

Also, the intent of this PR is to test how the merged configs are handled by *different* containerd versions. I don't feel as if this really achieve this since we're only ever testing the version of `containerd` that is included in the kind node image. 

Can one parameterize the test for `1.7.x` and `2.0` so that we can validate the behavior there? (I'm happy to do this as a follow-up, but it would be good to understand how we would handle the expected differences).

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: KIND comes with a preconfigured containerd config.toml file.

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: we now parse the TOML and run the checks from it

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: done

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: done

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: we now check depending on the config version

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,850 @@
...
+# Add imports directive if not present
+if ! grep -q "imports" /etc/containerd/config.toml; then
+	# Create imports line
+	cat > /tmp/imports.line <<EOF
+imports = ["/etc/containerd/conf.d/*.toml"]
+EOF
+	# Prepend to existing config
+	cat /etc/containerd/config.toml >> /tmp/imports.line
+	mv /tmp/imports.line /etc/containerd/config.toml
+fi`)
```
elezar: Why are we doing this? The `nvidia-ctk runtime configure` command should do this in cases where the imports directive does not exist.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,850 @@
...
+					output, _, err := nestedContainerRunner.Run(`containerd config dump`)
+					Expect(err).ToNot(HaveOccurred())
+
+					// Parse the TOML output
+					config, err := parseContainerdConfig(output)
+					Expect(err).ToNot(HaveOccurred(), "Failed to parse containerd config")
+
+					// Verify config version
+					version := config.Get("version")
+					Expect(version).To(Equal(int64(env.configVersion)))
```
elezar: Question: Does gomega have an `EqualValue` matcher so that we don't have to worry about the type?

(Alternatively, just update the type of `env.configVersion` to be `int64`).

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,850 @@
...
+
+					// Verify imports
+					if env.configVersion == 2 {
+						// containerd 1.7 shows actual resolved imports
+						err = validateImports(config, []string{"/etc/containerd/conf.d/99-nvidia.toml"}, true)
```
elezar: This is a sideeffect of the `config dump` command and does not reflect the actual content of the config file. One question I do have is why we're using `validateImports` here and not for the non v2 case?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,850 @@
...
+							Expect(found).To(BeTrue(), "Expected import pattern not found")
+						}
+					}
+
+					// Get plugin configuration
+					pluginConfig, err := getPluginConfig(config, env.configVersion)
+					Expect(err).ToNot(HaveOccurred(), "Failed to get plugin config")
+
+					// Verify CDI is enabled
+					cdiEnabled, err := getCDIEnabled(pluginConfig, env.configVersion)
```
elezar: Why is the version relevant here?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,850 @@
...
+
+	containerdTree, ok := containerdSection.(*toml.Tree)
+	if !ok {
+		return "", fmt.Errorf("containerd section is not a TOML tree")
+	}
+
+	defaultRuntime := containerdTree.Get("default_runtime_name")
+	if defaultRuntime == nil {
+		return "", nil // No default runtime set
+	}
```
elezar: Why don't we use `GetPath`? or `Get("containerd.default_runtime_name")`?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,850 @@
...
+
+	if runtime, ok := defaultRuntime.(string); ok {
+		return runtime, nil
+	}
+
+	return "", fmt.Errorf("default_runtime_name is not a string")
+}
+
+// validateRuntimeConfig validates a specific runtime configuration
+func validateRuntimeConfig(runtime interface{}, expectedType string, expectedOptions map[string]interface{}) error {
```
elezar: Under which conditions do we ever call this with anything other than a `toml.Tree`?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,850 @@
...
+	case map[string]interface{}:
+		runtimeMap = v
+	case *toml.Tree:
+		runtimeMap = tomlTreeToMap(v)
+	default:
+		return fmt.Errorf("runtime is not a map or toml.Tree, got %T", runtime)
+	}
+
+	// Check runtime type
+	runtimeType, ok := runtimeMap["runtime_type"].(string)
```
elezar: This seems like something that we can check quite simply at the call site -- especially since we seem to mostly call this with `runtimeType == ""`.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,850 @@
...
+	if !ok {
+		return fmt.Errorf("runtime_type not found or not a string")
+	}
+
+	if expectedType != "" {
+		// For empty expectedType, we don't check the runtime type
+		// For nvidia runtime, we accept various runtime types
+		if expectedType == "nvidia" {
+			validTypes := []string{"io.containerd.runc.v2", "io.containerd.nvidia.v1", "io.containerd.nvidia.v2"}
```
elezar: When is this ever `io.containerd.nvidia.v1`? Where does this list come from?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,850 @@
...
+	// Check runtime type
+	runtimeType, ok := runtimeMap["runtime_type"].(string)
+	if !ok {
+		return fmt.Errorf("runtime_type not found or not a string")
+	}
+
+	if expectedType != "" {
+		// For empty expectedType, we don't check the runtime type
+		// For nvidia runtime, we accept various runtime types
+		if expectedType == "nvidia" {
```
elezar: We never seem to call this with `expectedType == "nvidia"` what is the purpose of this branch?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,850 @@
...
+			options = v
+		case *toml.Tree:
+			options = tomlTreeToMap(v)
+		case nil:
+			return fmt.Errorf("options not found")
+		default:
+			return fmt.Errorf("options is not a map or toml.Tree, got %T", v)
+		}
+
+		for key, expectedValue := range expectedOptions {
```
elezar: Could we use a gomega `ContainsElements` or `ConsistsOf` instead?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,766 @@
...
+				if !env.hasDefaultImports {
+					// For containerd 1.7.x, ensure a config file exists
+					// nvidia-ctk should add the imports directive automatically
+					_, _, err = nestedContainerRunner.Run(`
+if [ ! -f /etc/containerd/config.toml ]; then
+	mkdir -p /etc/containerd
+	containerd config default > /etc/containerd/config.toml
+fi`)
+					Expect(err).ToNot(HaveOccurred(), "Failed to create default containerd config")
+				}
```
elezar: Could you please explain why this is even required? The nvidia-ctk will create this file if it does not exist. It really should not be needed to create this before the tests.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,766 @@
...
+
+				// Ensure containerd is running
+				_, _, err = nestedContainerRunner.Run(`
+					# Start containerd if not running
+					if ! systemctl is-active --quiet containerd; then
+						systemctl start containerd
+						sleep 2
+					fi
+				`)
+				Expect(err).ToNot(HaveOccurred(), "Failed to ensure containerd is running")
```
elezar: Why do we need this IN addition to `restartContainerdAndWait`? Is there a reason that we can't just call that here if required?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,766 @@
...
+			else
+				# If no backup exists, just clean up
+				rm -rf /etc/containerd/conf.d
+				mkdir -p /etc/containerd/conf.d
+			fi
+			
+			# Restore the original config.toml
+			if [ -f /tmp/containerd-config.toml.backup ]; then
+				cp /tmp/containerd-config.toml.backup /etc/containerd/config.toml
+			fi
```
elezar: Should we REMOVE `/etc/containerd/config.toml` if `/tmp/containerd-config.toml.backup` does not exist?

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,770 @@
...
+
+				// Step 2: Restart containerd and wait for it to be ready
+				err = restartContainerdAndWait(nestedContainerRunner)
+				Expect(err).ToNot(HaveOccurred(), "Failed to restart containerd after configuration restore")
+			})
+
+			When("configuring containerd", func() {
+				It("should add NVIDIA runtime using drop-in config without modifying the main config", func(ctx context.Context) {
+					// Configure containerd using nvidia-ctk
+					_, _, err := nestedContainerRunner.Run(`nvidia-ctk runtime configure --runtime=containerd --config=/etc/containerd/config.toml --drop-in-config=/etc/containerd/conf.d/99-nvidia.toml --set-as-default --cdi.enabled`)
```
elezar: Just as a note: Should we use the defaults here for the `--config` and `drop-in-config` flags? The fact that we're explicitly setting the paths here is one reason that we missed the issue reported in #1404 (see #1407).

tests/e2e/runner.go
```diff
@@ -106,13 +110,24 @@ func NewNestedContainerRunner(runner Runner, baseImage string, installCTK bool,
 
 	var additionalContainerArguments []string
 
+	if requiresGPUs {
```
ArangoGutierrez, line 112: Ha! nice, thanks for this Evan.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,806 @@
...
+		`)
+				Expect(err).ToNot(HaveOccurred(), "Failed to restore containerd configuration")
+
+				// Step 2: Restart containerd and wait for it to be ready
+				err = restartContainerdAndWait(nestedContainerRunner)
+				Expect(err).ToNot(HaveOccurred(), "Failed to restart containerd after configuration restore")
+			})
+
+			When("configuring containerd", func() {
+				It("should add NVIDIA runtime using drop-in config and validate merged configuration", func(ctx context.Context) {
```
elezar: the "and validate merged configuration" is what we're doing in the test, not what we expect.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,806 @@
...
+					// Configure containerd using nvidia-ctk
+					_, _, err := nestedContainerRunner.Run(`nvidia-ctk runtime configure --runtime=containerd --set-as-default --cdi.enabled`)
+					Expect(err).ToNot(HaveOccurred(), "Failed to configure containerd")
```
elezar: I think this is what determines what comes in the "It" statement.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,806 @@
...
+					// Verify drop-in config was created
+					dropInConfigContents, _, err := nestedContainerRunner.Run("cat /etc/containerd/conf.d/99-nvidia.toml")
+					Expect(err).ToNot(HaveOccurred())
+					Expect(dropInConfigContents).ToNot(BeEmpty())
+
+					// Verify main config has imports directive
+					mainConfigContents, _, err := nestedContainerRunner.Run("cat /etc/containerd/config.toml")
+					Expect(err).ToNot(HaveOccurred())
+					Expect(mainConfigContents).To(ContainSubstring("imports"))
+					Expect(mainConfigContents).To(ContainSubstring("/etc/containerd/conf.d"))
```
elezar: If we want to check this, we should parse the toml.

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,806 @@
...
+
+					// Verify config version
+					Expect(config.Get("version")).To(Equal(env.configVersion))
+
+					// Verify imports are present in the merged config
+					imports := config.Get("imports")
+					Expect(imports).ToNot(BeNil())
+					importsList, ok := imports.([]interface{})
+					Expect(ok).To(BeTrue())
+					Expect(len(importsList)).To(BeNumerically(">=", 1))
```
elezar: Checking for the presence of imports is not sufficient. We need to ensure that we ALSO check that the desired folder (or file) is imported. For example, see the matcher here: https://github.com/ArangoGutierrez/nvidia-container-toolkit/pull/1/files#diff-056e23fcacb20f063e793331c10a8fedec7eedb12c88a3c7e857ad995ace49daR234-R249

(I'm not saying that we need to use a matcher, but that we should check the value properly).

ArangoGutierrez: Edited, to now look for the file depending on the config version

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: edited

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: edited

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: it nows uses toml parsing and matching

tests/e2e/nvidia-ctk_containerd_test.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+// Integration tests for containerd drop-in config functionality.
+// These tests verify that nvidia-ctk runtime configure correctly applies
+// configuration changes while preserving existing settings. The preservation
+// validation is critical for containerd < 2.1 where plugin merge behavior
+// (commit 598c632) requires duplicating the CRI plugin section.
+var _ = Describe("containerd", Ordered, ContinueOnFailure, Label("container-runtime"), func() {
+	// Construct the test configs using the same matrix as the original tests
+	var testsConfigs []*testConfig
```
elezar: What are the "original tests" here?

ArangoGutierrez: removed, at some point, I had a `nvidia-ctl_containerdv2_test.go` file for testing things, it was a self note comment

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: Yes we could, but that would make this section more complex. 

The current implementation is actually the right choice here. `ContainElements` and `ConsistsOf` work on slices/arrays, not maps, so we'd need conversion logic to use them.
The current approach gives us better error messages when tests fail. If an option validation fails, we get a specific error telling us exactly which key had the issue:

```go
return fmt.Errorf("error matching option %s: %v", key, err)
```

We immediately know which option failed rather than having to dig through generic matcher output.
Also, this approach validates that our expected options exist without failing if there are additional options in the map, which is the right semantics here. ConsistsOf would fail if containerd adds any extra keys we're not explicitly checking for.
If we wanted to use a single assertion, we could do:

```go
matchers := []types.GomegaMatcher{}
for key, expectedValue := range expectedOptions {
    matchers = append(matchers, HaveKeyWithValue(key, expectedValue))
}
Expect(options).To(And(matchers...))
matchers := []types.GomegaMatcher{}for key, expectedValue := range expectedOptions {    matchers = append(matchers, HaveKeyWithValue(key, expectedValue))}Expect(options).To(And(matchers...))
```

But this would give us a worse error context and more complex code without any real benefit.

tests/e2e/nvidia-ctk_containerd_test.go
ArangoGutierrez: You are right, I have edited this to accept `map[string]interface{}`

ArangoGutierrez:
> What do you think of [ArangoGutierrez#3](https://github.com/ArangoGutierrez/nvidia-container-toolkit/pull/3)?

🥇

elezar:
@ArangoGutierrez let's squash and merge this.

PR #1314, opened 2025-09-23
Deprecate the hook config mode for cri-o

cdesiniotis:
The GPU Operator is switching from 'hook' to 'config' by default in the next release. This change ensures the default behavior of the toolkit container and GPU Operator are aligned. We also add deprecation messages to cases where an OCI hook is explicitly requested.

See thread for additional context: https://github.com/NVIDIA/gpu-operator/pull/1578#discussion_r2361236018

### DIFF
--- a/cmd/nvidia-ctk-installer/container/runtime/crio/crio.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/crio/crio.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/container"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/config"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
 	"github.com/NVIDIA/nvidia-container-toolkit/pkg/config/engine"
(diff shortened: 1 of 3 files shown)


ArangoGutierrez (commented):
LGTM, I'll leave final approval to ELezar

cmd/nvidia-ctk-installer/container/runtime/crio/crio.go
```diff
@@ -35,7 +35,7 @@ import (
 const (
 	Name = "crio"
 
-	defaultConfigMode = "hook"
+	defaultConfigMode = "config"
```
elezar: My one question was whether we wanted to print a deprecation warning when `hook` is configured?

In this case we would also need to update the `nvidia-ctk runtime configure` command to issue a similar warning.

elezar (changes requested):
Minor UX improvement in terms of issuing a deprecation warning.

In order to more easily capture this in the release notes, maybe we can update the commit message to something like:
```
Deprecate hook config mode for cro-o
```.

elezar (approved):
Thanks @cdesiniotis.

I added the deprecation log messages as changes on-top.

cdesiniotis:
Thanks @elezar

PR #1315, opened 2025-09-24
Allow OpenRM drivers to be discovered in CSV mode

elezar:
On Tegra-based systems with OpenRM (e.g. Thor-based systems). We need to also discover driver libraries in the same way was what we do on regular desktop systems. This change ensures that these driver libraries and meta devices are discovered.

### DIFF
--- a/pkg/nvcdi/lib-csv.go
+++ b/pkg/nvcdi/lib-csv.go
@@ -82,5 +82,25 @@ func (l *csvlib) GetDeviceSpecs() ([]specs.Device, error) {
 
 // GetCommonEdits generates a CDI specification that can be used for ANY devices
 func (l *csvlib) GetCommonEdits() (*cdi.ContainerEdits, error) {
-	return edits.FromDiscoverer(discover.None{})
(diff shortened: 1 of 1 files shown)


pkg/nvcdi/lib-csv.go
```diff
@@ -82,5 +82,25 @@ func (l *csvlib) GetDeviceSpecs() ([]specs.Device, error) {
 
 // GetCommonEdits generates a CDI specification that can be used for ANY devices
 func (l *csvlib) GetCommonEdits() (*cdi.ContainerEdits, error) {
-	return edits.FromDiscoverer(discover.None{})
+	if l.featureFlags["disable-openrm-drivers-in-csv"] {
```
ArangoGutierrez, line 85: Question: Where is this FeatureFlag documented?

elezar: It isn't yet. This is something that I need to follow-up on if this PR moves out of draft.

elezar:
Superceded by #1461

PR #1247, opened 2025-08-15
Generate CDI specs for coherent and non-coherent devices

elezar:
This change updates CDI spec generation to generation specifications for coherent and non-coherent devices. At its core, this is acheived by adding device-level annotations to the CDI specification for devices indicating whether a device is coherent or not. The `nvidia-ctk cdi generate` command then splits the generated spec by the specified annotation and generates additional specs for `nvidia.com/gpu.coherent` and / or `nvidia.com/gpu.noncoherent`devices. 

For example, when running on a device that supports coherence we see the following:
```
$ ./nvidia-ctk cdi generate --output $(pwd)/test/nvidia.yaml
$ ls -l test/
total 48
-rw-r--r-- 1 local-elezar local-elezar 20820 Aug 15 13:22 nvidia.coherent.yaml
-rw-r--r-- 1 local-elezar local-elezar 20811 Aug 15 13:22 nvidia.yaml
```
With the only difference between the two specs being the device kind:
```
$ diff test/nvidia.yaml test/nvidia.coherent.yaml
3c3
< kind: nvidia.com/gpu
---
```
The following specs are generated:
```
$ nvidia-ctk cdi list --spec-dir=$(pwd)/test/
INFO[0000] Found 6 CDI devices
nvidia.com/gpu.coherent=0
nvidia.com/gpu.coherent=GPU-ca81aac1-36e7-2d26-8a15-0aa6ef17c627
nvidia.com/gpu.coherent=all
nvidia.com/gpu=0
nvidia.com/gpu=GPU-ca81aac1-36e7-2d26-8a15-0aa6ef17c627
nvidia.com/gpu=all
```

This is blocked by:
* https://github.com/NVIDIA/go-nvml/pull/156
* https://github.com/NVIDIA/go-nvlib/pull/68

PR #1319, opened 2025-09-26
Disable generation of nvidia.com/gpu.[non]coherent CDI specs by default

elezar:
This change bumps the `github.com/NVIDIA/go-nvlib` dependency from `v0.8.0` to `v0.8.1` to ensures that `nvidia.com/gpu.coherent` CDI specs are only generated for devices with an addressing mode of ATS.

It also disables the functionality for splitting generated CDI specifications based on device coherence by default. This was added in #1247, but due to discussions around whether coherence is an property that should be exposed, we are disabling this by default.

Note that users can opt in to the feature by running the `nvidia-ctk cdi generate` with the `--feature-flag=enable-coherent-annotations` command line flag. Alternatively the `nvidia-ctk cdi generate` command can be run with the `NVIDIA_CTK_CDI_GENERATE_FEATURE_FLAGS` enviroment set to include `"enable-coherent-annotations"` (in a comma-separated list).

The same feature flag can be passed to the `nvcdi` API when constructing a CDI spec generator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # NVIDIA Container Toolkit Changelog
 
+## v1.18.0-rc.5
+- Update go-nvlib to restrict nvidia.com/gpu.coherent devices to devices with an ATS addressing mode.
+
 ## v1.18.0-rc.4
 
 - Add drop-in file support for containerd and crio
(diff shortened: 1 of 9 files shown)


klueska (approved):
What was the primary motivation for the change in semantics?

elezar:
> It also disables the functionality for splitting generated CDI specifications based on device coherence by default. This was added in #1247, but due to discussions around whether coherence is an property that should be exposed, we are disabling this by default.

As noted in the description:
> due to discussions around whether coherence is an property that should be exposed, we are disabling this by default.

With regards to what constitutes coherence, it was called out after the initial implementation that HMM addressing mode is not actually coherent. This was initially included due to an incorrect requirement in the original request.

PR #1323, opened 2025-09-26
add support for building artifacts with custom GOPROXY

tariq1890:


.github/workflows/image.yaml
```diff
@@ -109,12 +113,26 @@ jobs:
...
         env:
           IMAGE_NAME: ghcr.io/nvidia/container-toolkit
           VERSION: ${{ inputs.version }}
           PUSH_ON_BUILD: "true"
           BUILD_MULTI_ARCH_IMAGES: ${{ inputs.build_multi_arch_images }}
+          OIDC_USER: ${{ steps.jfrog.outputs.oidc-user }}
+          OIDC_TOKEN: ${{ steps.jfrog.outputs.oidc-token }}
+          OIDC_ARTIFACTORY_ENDPOINT: ${{ secrets.OIDC_ARTIFACTORY_ENDPOINT }}
         run: |
+          OIDC_USER_ENCODED=$(python3 -c 'import urllib.parse, os; print(urllib.parse.quote_plus(os.environ["OIDC_USER"]))')
```
elezar: Question: Why do we not just output the encoded user in the jfrog step?

elezar: Or better yet, have a single step that generates the goproxy to be reused elsewhere?

.github/workflows/golang.yaml
```diff
@@ -100,4 +96,19 @@ jobs:
...
+        with:
+          oidc-provider-name: nvgithub
+          oidc-audience: ${{ secrets.OIDC_ARTIFACTORY_ENDPOINT }}
+      - env:
+          OIDC_USER: ${{ steps.jfrog.outputs.oidc-user }}
+          OIDC_TOKEN: ${{ steps.jfrog.outputs.oidc-token }}
+          OIDC_ARTIFACTORY_ENDPOINT: ${{ secrets.OIDC_ARTIFACTORY_ENDPOINT }}
+        run: |
+          OIDC_USER_ENCODED=$(python3 -c 'import urllib.parse, os; print(urllib.parse.quote_plus(os.environ["OIDC_USER"]))')
+          export GOPROXY="https://${OIDC_USER_ENCODED}:${OIDC_TOKEN}@${OIDC_ARTIFACTORY_ENDPOINT}/artifactory/api/go/edge-go-remote-virtual"
```
elezar: Does the goproxy have to be set for the standard go builds too? The reason that this is run on PRs is that it is lightweight and doesn't need any specific permissions. The artifacts of the build are not used in any production images.

tariq1890: > Does the goproxy have to be set for the standard go builds too? 

No, it really is optional. I had added this as it was easier to debug any issues stemming from the custom GOPROXY when enabling it here as opposed to the image builds. If you prefer, I can remove the GOPROXY enablement from this step

elezar: The point on debugging failures is valid. With that in mind, I think we can leave it in place, but I would like to see a simplification where we ONLY set the envvar in the go build job instead of extracting the encoded values there.

deployments/container/Dockerfile
```diff
@@ -42,6 +42,9 @@ RUN set -eux; \
 ENV GOPATH=/go
 ENV PATH=$GOPATH/bin:/usr/local/go/bin:$PATH
 
+ARG GOPROXY=direct
```
elezar: The default value of this is not `direct` (see https://go.dev/ref/mod#private-module-privacy). Should we set this accordingly?

tariq1890: Oh good catch. Yes, I think we should set it to that

.github/workflows/golang.yaml
tariq1890: Sure, let me see what can be done about that

deployments/container/Dockerfile
tariq1890: Updated. Please check

.github/workflows/golang.yaml
tariq1890: I have created a composite github action which can reused. Please check

.github/workflows/image.yaml
tariq1890: Please check now

cdesiniotis (approved):

cdesiniotis (approved):

elezar:
Thanks @tariq1890. This looks much cleaner now that we're reusing the logic.

PR #1329, opened 2025-10-01
[no-relnote] Add basic regression test for nvidiascape

elezar:
This backports a test from #1327

### DIFF
--- a/tests/e2e/nvidia-container-toolkit_test.go
+++ b/tests/e2e/nvidia-container-toolkit_test.go
@@ -215,4 +215,94 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
 			Expect(ldconfigOut).To(ContainSubstring("/usr/lib64"))
 		})
 	})
+
+	When("Running a container with LD_PRELOAD", Ordered, func() {
+		BeforeAll(func(ctx context.Context) {
+			// Create the source for the poc.
+			_, _, err := r.Run(`cat <<EOF > poc.c
+/**
+Code taken from https://youtu.be/56vcNIh35PA?si=gmh7Cx9P-lNTbl4L&t=328
+**/
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <fcntl.h>
+__attribute__((constructor))
+void init() {
+// Ultra-minimal exploit just create a marker file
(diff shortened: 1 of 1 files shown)


tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -215,4 +215,94 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
 			Expect(ldconfigOut).To(ContainSubstring("/usr/lib64"))
 		})
 	})
+
+	When("Running a container with LD_PRELOAD", Ordered, func() {
+		BeforeAll(func(ctx context.Context) {
+			// Create the source for the poc.
+			_, _, err := r.Run(`cat <<EOF > poc.c
+/**
+Code taken from https://youtu.be/56vcNIh35PA?si=gmh7Cx9P-lNTbl4L&t=328
```
jgehrcke, line 224: You just made me watch this. Thank you!

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -215,4 +215,94 @@ var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
...
+EOF`)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Create the local Dockerfile
+			_, _, err = r.Run(`cat <<EOF > Dockerfile.nvidiascape
+FROM ubuntu AS build
+RUN apt-get update && \
+	apt-get install -y gcc \
+	&& \
+	rm -rf /var/lib/apt/lists/*
```
jgehrcke, line 248: Is this a tiny bit of an annoyance in terms of how long it takes?

Could using https://hub.docker.com/_/gcc be faster?

elezar: That's a good question and could be valid.

One issue is that we may need to loosly match (or not exceed) the glibc version on the host since the library is actually preloaded there. This is simpler with a base image like `ubuntu` than something more opague like a `gcc` image. 

Also, since spinning up and tearing down the AWS instances to run the tests still takes the most time in this CI (see below), I don't want to make this optimisation in a backport.

Screenshot 2025-10-01 at 14 28 12

PR #1331, opened 2025-10-01
Add ability to opt-out of drop-in files

cdesiniotis:
This change allows the use of drop-in config files for containerd and cri-o to be disabled when setting `RUNTIME_DROP_IN_CONFIG` to `""`. This allows better support for use cases such as `microk8s` where the top-level config may be a template that is rendered at runtime.

### DIFF
--- a/cmd/nvidia-ctk-installer/container/runtime/containerd/containerd.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/containerd/containerd.go
@@ -183,6 +183,7 @@ func getRuntimeConfig(o *container.Options, co *Options) (engine.Interface, erro
 		containerd.WithRuntimeType(co.runtimeType),
 		containerd.WithUseLegacyConfig(co.useLegacyConfig),
 		containerd.WithContainerAnnotations(co.containerAnnotationsFromCDIPrefixes()...),
+		containerd.WithDisableDropInConfig(o.DropInConfig == ""),
 	}
 	if o.DropInConfigHostPath != "" && o.DropInConfig != "" {
 		options = append(options,
--- a/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/containerd/config_test.go
@@ -1780,6 +1780,113 @@ version = 3
 				return nil
 			},
 		},
+		{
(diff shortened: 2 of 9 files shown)


elezar:
@cdesiniotis note that we explicitly distinguish between an empty value for `RUNTIME_DROP_IN_CONFIG` and the default value for a specific runtime. If a user sets `RUNTIME_DROP_IN_CONFIG=""` this *should* disable the use of drop-in configs meaning that an additional flag is not required.

**Update:** While adding a unit test to validate this behaviour, I noticed that this was not happening. I would still like to avoid adding an extra config option though. I have updated the PR. Let me know what you think.

elezar (approved):
Feel free to squash my commit on top and merge.

cmd/nvidia-ctk-installer/container/runtime/containerd/containerd.go
```diff
@@ -183,6 +183,7 @@ func getRuntimeConfig(o *container.Options, co *Options) (engine.Interface, erro
 		containerd.WithRuntimeType(co.runtimeType),
 		containerd.WithUseLegacyConfig(co.useLegacyConfig),
 		containerd.WithContainerAnnotations(co.containerAnnotationsFromCDIPrefixes()...),
+		containerd.WithDisableDropInConfig(o.DropInConfig == ""),
```
tariq1890, line 186: I don't quite understand the benefit of adding a new field here, if we can just directly evaluate this conditional internally when processing the runtime config

```
o.DropInConfig == ""
```

elezar: The `DropInConfig` path is not passed to the configs themselves.

ArangoGutierrez (approved):

ArangoGutierrez:
After comment https://github.com/NVIDIA/nvidia-container-toolkit/pull/1251#issuecomment-3373921015 , do we still want this patch?

cdesiniotis:
> After comment https://github.com/NVIDIA/nvidia-container-toolkit/pull/1251#issuecomment-3373921015 , do we still want this patch?

No, I think we can hold off on this. Closing for now and we can revisit this if the need to disable drop-in files resurfaces.

Issue #1340, opened 2025-10-07
Building a previous version from github source


jgforbes:
Is there documentation somewhere on how to build an older release from the source tree. I need to build for a centos8 target and need the 1.17.8-1 version. I can build the latest release candidate.

elezar:
@jgforbes is there a specific reason that you need to build from source? Are you only interested in the pacakges, or also the container image?

The process that we would use is:
```
git checkout v1.17.8
./scripts/build-packages.sh centos8-x86_64
```

Note that the `centos8-x86_64` target is not something that we use often so there may be minor issues in this flow.

jgforbes:
The RPMs as supplied will not work with RHEL-8 or RHEL-9 in FIPS mode. RHEL-9 FIPS mode will not tolerate SHA1
See issue  #116
Building for the centos8 target partially solves the problem for fips mode RHEL-8.
From what I understand, RHEL7/CENTOS7 RPM tools will allow building with SHA256 for digest, but I am not sure where to modify the build scripts to enable this. 

Can you help out with this?

elezar:
@jgforbes would something like https://github.com/NVIDIA/nvidia-container-toolkit/pull/1351 work for this? Looking at the produced packages we have:
```
rpm --checksig --verbose  nvidia-container-toolkit-1.18.0~rc.6-1.aarch64.rpm
nvidia-container-toolkit-1.18.0~rc.6-1.aarch64.rpm:
    Header SHA256 digest: OK
    Header SHA1 digest: OK
    Payload SHA256 digest: OK
    MD5 digest: OK
```
which seems to indicate that these have the required digests (due to the newer `rpmbuild` version used). We would have to do something similar for the `libnvidia-container` packages, but this should be possible with a similar approach.

jgforbes:
@elezar Yes, I believe it will. Here is the same test of a redhat RHEL-9.6 RPM package:
```
$rpm --checksig --verbose /tmp/NetworkManager-1.52.0-7.el9_6.x86_64.rpm
/tmp/NetworkManager-1.52.0-7.el9_6.x86_64.rpm:
    Header V4 RSA/SHA256 Signature, key ID fd431d51: OK
    Header SHA256 digest: OK
    Header SHA1 digest: OK
    Payload SHA256 digest: OK
    MD5 digest: OK
```
Other Nvidia RPMs have the same header digest, e.g.:
```
[...]

PR #1351, opened 2025-10-13
Produce RPM packages with SHA256 header digests

elezar:
See #1340

### DIFF
--- a/docker/Dockerfile.rpm-yum
+++ b/docker/Dockerfile.rpm-yum
@@ -15,7 +15,7 @@
 # This is the dockerfile for building packages on yum-based RPM systems.
 
 ARG BASEIMAGE
-FROM ${BASEIMAGE}
+FROM ${BASEIMAGE} AS build
 
 # centos:stream8 is EOL.
 # We switch to the vault repositories for this base image.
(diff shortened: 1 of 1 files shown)


elezar:
We would need to make similar changes to `libnvidia-container`.

docker/Dockerfile.rpm-yum
```diff
@@ -15,7 +15,7 @@
 # This is the dockerfile for building packages on yum-based RPM systems.
 
 ARG BASEIMAGE
-FROM ${BASEIMAGE}
+FROM ${BASEIMAGE} AS build
```
ebendler, line 18: why not just change the BASEIMAGE for rhel to `registry.access.redhat.com/ubi8/ubi:latest` and not have to use a janky EOL centos image? It gets you `rpm-build-4.14.3-32.el8_10.x86_64` which (I believe) defaults to the sha256 header you need.

ebendler: Ah, it's the build stuff in libnvidia-container. Hold on a sec and I'll send some PRs that use the Oracle versions instead.

elezar: The reason that we use the "janky" image for `BASEIMAGE` is to ensure that we build the executables with a low enough glibc version to allow compatibility with our support matrix. For RPM-based packages this INCLUDES RHEL7 since we need to support *at least* all active driver branches.

antheas:
Bump

2 days to go

ArangoGutierrez:
Does https://github.com/NVIDIA/nvidia-container-toolkit/pull/1411 supersede this PR?

elezar:
Closing in favour of #1411

PR #1360, opened 2025-10-16
[nvidia-ctk-installer] do not revert cri-o config on shutdown

cdesiniotis:
This commit updates the behavior of the nvidia-ctk-installer for cri-o.
On shutdown, we no longer delete the drop-in config file as long as
none of the nvidia runtime handlers are set as the default runtime.
This change was made to workaround an issue observed when uninstalling
the gpu-operator -- management containers launched with the nvidia
runtime handler would get stuck in the terminating state with the below
error message:

```
failed to find runtime handler nvidia from runtime list map[crun:... runc:...], failed to "KillPodSandbox" for ...
```

There appears to be a race condition where the nvidia-ctk-installer removes the drop-in file
and restarts cri-o. After the cri-o restart, if there are still pods / containers to terminate
that were started with the nvidia runtime, then cri-o fails to terminate them. The behavior
of cri-o, and its in-memory runtime handler cache, appears to differ from that of containerd as
we have never encountered such an issue with containerd.

This commit can be considered a stop-gap solution until more robust solution is developed.

Signed-off-by: Christopher Desiniotis <cdesiniotis@nvidia.com>

### DIFF
--- a/cmd/nvidia-ctk-installer/container/runtime/crio/crio.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/crio/crio.go
@@ -133,7 +133,7 @@ func setupHook(o *container.Options, co *Options) error {
 func setupConfig(o *container.Options) error {
 	log.Infof("Updating config file")
 
-	cfg, err := getRuntimeConfig(o)
+	cfg, err := getRuntimeConfig(o, false)
 	if err != nil {
 		return fmt.Errorf("unable to load config: %v", err)
 	}
@@ -186,14 +186,18 @@ func cleanupConfig(o *container.Options) error {
 
 	log.Infof("Reverting config file modifications")
 
-	cfg, err := getRuntimeConfig(o)
+	cfg, err := getRuntimeConfig(o, true)
 	if err != nil {
 		return fmt.Errorf("unable to load config: %v", err)
 	}
 
-	err = o.Unconfigure(cfg)
+	err = cfg.UpdateDefaultRuntime(o.RuntimeName, engine.UpdateActionUnset)
 	if err != nil {
-		return fmt.Errorf("unable to unconfigure cri-o: %v", err)
+		return fmt.Errorf("failed to unset %q as the default runtime: %w", o.RuntimeName, err)
+	}
+
+	if err := o.Flush(cfg); err != nil {
+		return err
 	}
 
 	err = RestartCrio(o)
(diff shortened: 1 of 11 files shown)


elezar:
One question: Even if we don't delete the drop-in file, do we no still remove the binaries?

cmd/nvidia-ctk-installer/container/runtime/crio/crio.go
```diff
@@ -180,6 +180,10 @@ func cleanupHook(co *Options) error {
 
 // cleanupConfig removes the NVIDIA container runtime from the cri-o config
 func cleanupConfig(o *container.Options) error {
+	if !o.SetAsDefault {
+		return nil
+	}
```
elezar, line 185: Don't we still need to check whether we're using a drop-in file? We *could* be modifying the top-level config directly.

cdesiniotis: I am not sure I follow why that matters? Regardless of whether we use a drop-in file or not, removing the nvidia runtime from the cri-o config can lead to the issues described in the PR description.

The reason I thought to add this conditional is that we must attempt to revert the config if nvidia is the default runtime, or else all future pods will fail to function (after the nvidia runtime binaries are removed).

cdesiniotis:
> One question: Even if we don't delete the drop-in file, do we no still remove the binaries?

I believe we still do, yes.

pkg/config/engine/config.go
```diff
@@ -43,6 +43,7 @@ type RuntimeConfigDestination interface {
 	AddRuntimeWithOptions(string, string, bool, interface{}) error
 	EnableCDI()
 	RemoveRuntime(string) error
+	UnsetDefaultRuntime(string)
```
tariq1890: Thoughts on having a more generic method here?

What about this?
```
SetRuntimeProperty(propertyName, propertyValue)
```

cdesiniotis: Not opposed. But in this case I want to delete a field, so `SetRuntimeProperty` may not allow for that.

tariq1890: How about this?
```
UpdateRuntimeProperty(propertyName, propertyValue, operation)
```

operation could be one of - CREATE, DELETE, MODIFY

So in your case, it would look like 
```
UpdateRuntimeProperty("default_runtime", "", DELETE)
```

pkg/config/engine/containerd/config.go
```diff
@@ -154,3 +154,33 @@ func (c *Config) RemoveRuntime(name string) error {
 	*c.Tree = config
 	return nil
 }
+
+// UpdateDefaultRuntime updates the default runtime setting in the config.
+// When action is 'set' the provided runtime name is set as the default.
+// When action is 'unset' we make sure the provided runtime name is not
+// the default.
+func (c *Config) UpdateDefaultRuntime(name string, action string) error {
+	if action != "set" && action != "unset" {
```
tariq1890: can we move "set" and "unset" to constant vars?

tariq1890: Can we add this in the package level, so that it can reused in multiple places?

cdesiniotis: I just added constants for these in the `engine` package.

elezar:
> > One question: Even if we don't delete the drop-in file, do we no still remove the binaries?
> 
> I believe we still do, yes.

Just to close the loop on this. In checking the implementation yesterday, we do unconfigure the runtime, but do NOT remove the binaries. This is the reason that containerd still functions and why CRIO functions with this change. This should be a sufficient workaround for the time being, but the implications of being able to trigger the `nvidia` runtime even if it may not currently be configured / available in the strictest sense should be investigated.

PR #1361, opened 2025-10-16
Change RPM builder to Oracle Linux 8 to get sha256 digests by default

ebendler:
Change all CI to RHEL8 target.

Remove all traces of centos 7 use in the build.

CI will fail until the https://github.com/NVIDIA/libnvidia-container/pull/328 is in and the submodule reference here is updated to it.


elezar:
Which glibc version would the binaries be built with in RHEL8?

ebendler:
> Which glibc version would the binaries be built with in RHEL8?

```
# ld.so --version
ld.so (GNU libc) stable release version 2.28.
Copyright (C) 2020 Free Software Foundation, Inc.
```
Not sure what the OS support policy is for CTK, but the CUDA stack does not release on any EOL platforms.

ebendler:
I'm over-complicating this. The best fix is probably to just continue using the el7 builder for el7 packages and then add/post el8 packages for everything that needs the newer format

PR #1362, opened 2025-10-17
[nvidia-ctk-installer] introduce unconfigure mode to determine cleanu…

cdesiniotis:
…p behavior

### DIFF
--- a/cmd/nvidia-ctk-installer/container/container.go
+++ b/cmd/nvidia-ctk-installer/container/container.go
@@ -57,6 +57,8 @@ type Options struct {
 	HostRootMount string
 
 	ConfigSources []string
+
+	UnconfigureMode string
 }
 
 // Configure applies the options to the specified config
@@ -66,6 +68,29 @@ func (o Options) Configure(cfg engine.Interface) error {
 		return fmt.Errorf("unable to update config: %v", err)
 	}
 	return o.flush(cfg)
+}
+
+func (o Options) UnsetDefaultRuntime(cfg engine.Interface) error {
+	runtimes := operator.GetRuntimes(
+		operator.WithNvidiaRuntimeName(o.RuntimeName),
+		operator.WithSetAsDefault(o.SetAsDefault),
+		operator.WithRoot(o.RuntimeDir),
+	)
+	defaultRuntimeName := ""
+	for name, runtime := range runtimes {
+		if runtime.SetAsDefault {
+			defaultRuntimeName = name
+		}
+	}
+	if defaultRuntimeName == "" {
+		return nil
+	}
+
+	cfg.UnsetDefaultRuntime(defaultRuntimeName)
+	if err := o.flush(cfg); err != nil {
+		return err
+	}
+	return nil
 }
 
 // Unconfigure removes the options from the specified config
--- a/pkg/config/engine/containerd/config.go
+++ b/pkg/config/engine/containerd/config.go
@@ -123,6 +123,20 @@ func (c *Config) EnableCDI() {
 	*c.Tree = config
 }
 
+func (c *Config) UnsetDefaultRuntime(name string) {
+	if c == nil || c.Tree == nil {
+		return
+	}
+
+	config := *c.Tree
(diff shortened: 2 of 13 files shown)


cmd/nvidia-ctk-installer/container/container.go
```diff
@@ -68,6 +70,29 @@ func (o Options) Configure(cfg engine.Interface) error {
...
+	cfg.UnsetDefaultRuntime(defaultRuntimeName)
+	if err := o.flush(cfg); err != nil {
+		return err
+	}
+	return nil
+}
+
 // Unconfigure removes the options from the specified config
 func (o Options) Unconfigure(cfg engine.Interface) error {
 	err := o.RevertConfig(cfg)
```
elezar, line 98: What about checking the mode here instead of adding another top-level function to the Options. This *should* mean that we don't need to make changes for each runtime.

pkg/config/engine/containerd/config.go
```diff
@@ -123,6 +123,20 @@ func (c *Config) EnableCDI() {
 	*c.Tree = config
 }
 
+func (c *Config) UnsetDefaultRuntime(name string) {
```
elezar, line 126: Should we invoke this from `RemoveRuntime`?

pkg/config/engine/containerd/config_drop_in.go
```diff
@@ -88,6 +88,10 @@ func (c *ConfigWithDropIn) Save(dropInPath string) (int64, error) {
 	return bytesWritten, nil
 }
 
+func (c *ConfigWithDropIn) UnsetDefaultRuntime(name string) {
+	c.Interface.UnsetDefaultRuntime(name)
```
elezar, line 92: Note that this unsets it in the drop-in file and not in the top-level config.

pkg/config/engine/containerd/config_v1.go
```diff
@@ -75,6 +75,26 @@ func (c ConfigV1) DefaultRuntime() string {
...
+	if !ok || runtimePath == "" {
+		runtimePath, _ = config.GetPath([]string{"plugins", "cri", "containerd", "runtimes", name, "options", "Runtime"}).(string)
+	}
+	defaultRuntimePath, ok := config.GetPath([]string{"plugins", "cri", "containerd", "default_runtime", "options", "BinaryName"}).(string)
+	if !ok || defaultRuntimePath == "" {
+		defaultRuntimePath, _ = config.GetPath([]string{"plugins", "cri", "containerd", "default_runtime", "options", "Runtime"}).(string)
+	}
+	if runtimePath != "" && defaultRuntimePath != "" && runtimePath == defaultRuntimePath {
+		config.DeletePath([]string{"plugins", "cri", "containerd", "default_runtime"})
+	}
```
elezar, line 95: In other implementations we have `*c.Tree = config` at the end of modifying functions. Do we not need those?

cmd/nvidia-ctk-installer/container/container.go
cdesiniotis: Yeah, I think that should be possible...

pkg/config/engine/containerd/config.go
cdesiniotis: Yes, this was a copy-paste from the code currently in `RemoveRuntime`. I can look to reuse this new function there.

pkg/config/engine/containerd/config_v1.go
cdesiniotis: I probably missed this...

pkg/config/engine/containerd/config_drop_in.go
cdesiniotis: That was intentional. When using drop-ins, we set the `default_runtime_name` field only in the drop-in file. The purpose of `UnsetDefaultRuntime` is to revert that setting (in cases where nvidia is set as the default runtime).

cdesiniotis:
Closing in favor of #1360

PR #1364, opened 2025-10-17
Add rockylinux 8 as an RPM builder

ebendler:
This makes sure that RPMs have the fields that newer releases require.

Also:
- scrub centos from the release process, its dead
- update golang URLs to newer google endpoints
- replace all centos-stream:8 usage with rockylinux:8
- add "dist" tag to RPMs to distinguish el7 and el8 builds

This will fail CI until this libnvidia-container PR is merged. Come back and update this PR with new third_party pointer when that happens. https://github.com/NVIDIA/libnvidia-container/pull/329

### DIFF
--- a/docker/Dockerfile.debian
+++ b/docker/Dockerfile.debian
@@ -25,7 +25,7 @@ RUN set -eux; \
         aarch64 | arm64) ARCH='arm64' ;; \
         *) echo "unsupported architecture" ; exit 1 ;; \
     esac; \
-    wget -nv -O - https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-${ARCH}.tar.gz \
+    wget -nv -O - https://dl.google.com/go/go${GOLANG_VERSION}.linux-${ARCH}.tar.gz \
     | tar -C /usr/local -xz
 
 ENV GOPATH /go
--- a/scripts/packages-sign-all.sh
+++ b/scripts/packages-sign-all.sh
@@ -61,7 +61,7 @@ function sign() {
     case ${src_dist} in
     amazonlinux*) pkg_type=rpm
         ;;
-    centos* | rpm) pkg_type=rpm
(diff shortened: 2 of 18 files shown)


docker/Dockerfile.debian
```diff
@@ -25,7 +25,7 @@ RUN set -eux; \
         aarch64 | arm64) ARCH='arm64' ;; \
         *) echo "unsupported architecture" ; exit 1 ;; \
     esac; \
-    wget -nv -O - https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-${ARCH}.tar.gz \
+    wget -nv -O - https://dl.google.com/go/go${GOLANG_VERSION}.linux-${ARCH}.tar.gz \
```
ArangoGutierrez, line 28: This change looks out of scope for the commit. (PR description does mention it. 

I would prefer to have the 4 items listed on the PR description as standalone commits:

- scrub centos from the release process, its dead
- update golang URLs to newer google endpoints
- replace all centos-stream:8 usage with rockylinux:8
- add "dist" tag to RPMs to distinguish el7 and el8 builds

scripts/packages-sign-all.sh
```diff
@@ -61,7 +61,7 @@ function sign() {
     case ${src_dist} in
     amazonlinux*) pkg_type=rpm
         ;;
-    centos* | rpm) pkg_type=rpm
+    centos* | rhel* | rpm) pkg_type=rpm
```
ArangoGutierrez, line 64: If we are getting rid of CentOS, does it make sense to continue to list it here?

docker/Dockerfile.debian
ebendler: My understanding is that you're moving to kitmaker in the near-ish future, so this PR is largely pointless. I thought @elezar was going to close it, but don't spend any energy on it.

PR #1365, opened 2025-10-17
invoke the actual default low-level runtime in the nvidia-ctk wrapper script

tariq1890:
Currently, the nvidia-container-runtime wrapper binary falls back to `runc` whenever it finds that the nvidia driver modules are no longer loaded. This will not work in clusters that have newer versions CRIO where only `crun` is installed and there is no `runc` binary present.

With this change, we pick the first element from the `opts.ContainerRuntimeRuntimes` list which is already populated with low level runtime binary paths extracted from the runtime config TOML. This makes for a more robust logic when choosing the fallback runtime as opposed to a hardcoded `"runc"`

### DIFF
--- a/cmd/nvidia-ctk-installer/toolkit/toolkit.go
+++ b/cmd/nvidia-ctk-installer/toolkit/toolkit.go
@@ -26,6 +26,7 @@ import (
 	"tags.cncf.io/container-device-interface/pkg/cdi"
 	"tags.cncf.io/container-device-interface/pkg/parser"
 
+	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/container/runtime/crio"
 	"github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk-installer/toolkit/installer"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/config"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
@@ -281,7 +282,7 @@ func (t *Installer) ValidateOptions(opts *Options) error {
 
 // Install installs the components of the NVIDIA container toolkit.
 // Any existing installation is removed.
-func (t *Installer) Install(cli *cli.Command, opts *Options) error {
+func (t *Installer) Install(cli *cli.Command, opts *Options, runtime string) error {
 	if t == nil {
 		return fmt.Errorf("toolkit installer is not initilized")
 	}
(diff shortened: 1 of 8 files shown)


tariq1890:
### Scenario 1: Cluster with CRIO

ContainerRuntimes list
```
[nvidia-container-runtime]
  debug = "/dev/null"
  log-level = "info"
  mode = "cdi"
  runtimes = ["/usr/libexec/crio/crun", "/usr/libexec/crio/runc", "runc", "crun"]
```

Resultant nvidia-container-runtime

```
$ cat /usr/local/nvidia/toolkit/nvidia-container-runtime
#! /bin/sh
cat /proc/modules | grep -e "^nvidia " >/dev/null 2>&1
if [ "${?}" != "0" ]; then
	echo "nvidia driver modules are not yet loaded, invoking /usr/libexec/crio/crun directly"
	exec /usr/libexec/crio/crun "$@"
fi
NVIDIA_CTK_CONFIG_FILE_PATH=/usr/local/nvidia/toolkit/.config/nvidia-container-runtime/config.toml \
PATH=/usr/local/nvidia/toolkit:$PATH \
	/usr/local/nvidia/toolkit/nvidia-container-runtime.real \
		"$@"
```

tariq1890:
### Scenario 2: Cluster with Containerd

ContainerRuntimes list
```
[nvidia-container-runtime]
  debug = "/dev/null"
  log-level = "info"
  mode = "cdi"
  runtimes = ["/usr/bin/runc", "runc", "crun"]
```

Resultant nvidia-container-runtime

```
$ cat /usr/local/nvidia/toolkit/nvidia-container-runtime
#! /bin/sh
cat /proc/modules | grep -e "^nvidia " >/dev/null 2>&1
if [ "${?}" != "0" ]; then
	echo "nvidia driver modules are not yet loaded, invoking /usr/bin/runc directly"
	exec /usr/bin/runc "$@"
fi
NVIDIA_CTK_CONFIG_FILE_PATH=/usr/local/nvidia/toolkit/.config/nvidia-container-runtime/config.toml \
PATH=/usr/local/nvidia/toolkit:$PATH \
	/usr/local/nvidia/toolkit/nvidia-container-runtime.real \
		"$@"
```

tariq1890:
### Scenario 3: Openshift Cluster with CRIO + runc

ContainerRuntimes list
```
[nvidia-container-runtime]
  debug = "/dev/null"
  log-level = "info"
  mode = "cdi"
  runtimes = ["/usr/bin/runc", "/usr/bin/crun", "runc", "crun"]
```

Resultant nvidia-container-runtime
```
$ cat /usr/local/nvidia/toolkit/nvidia-container-runtime
#! /bin/sh
cat /proc/modules | grep -e "^nvidia " >/dev/null 2>&1
if [ "${?}" != "0" ]; then
	echo "nvidia driver modules are not yet loaded, invoking /usr/bin/runc directly"
	exec /usr/bin/runc "$@"
fi
NVIDIA_CTK_CONFIG_FILE_PATH=/usr/local/nvidia/toolkit/.config/nvidia-container-runtime/config.toml \
PATH=/usr/local/nvidia/toolkit:$PATH \
	/usr/local/nvidia/toolkit/nvidia-container-runtime.real \
		"$@"
```

cmd/nvidia-ctk-installer/toolkit/toolkit.go
```diff
@@ -295,18 +296,35 @@ func (t *Installer) Install(cli *cli.Command, opts *Options) error {
...
 	}
 
+	var defaultRuntimeExecutable string
+	if len(opts.ContainerRuntimeRuntimes) > 0 {
+		defaultRuntimeExecutable = opts.ContainerRuntimeRuntimes[0]
+	}
+
+	// If opts.ContainerRuntimeRuntimes is empty (highly unlikely), we fall back to either runc or crun
+	// depending on the runtime
+	if len(defaultRuntimeExecutable) == 0 {
```
elezar, line 306: As a follow-up. Should we not error out here if it's unexpected, or at the very least log that we're using this value. Switching a default silently seems like something that will be difficult to debug when it causes problems.

elezar: Also. Should we expose this as a user-configurable value?

tariq1890: This conditional will never evaluate to true as `opts.ContainerRuntimeRuntimes` will always default to `["runc", "crun"]` at minimum. If the config toml has absolute paths of low-level runtime binaries, they will be added too in which case `opts.ContainerRuntimeRuntimes` would look something like `["/usr/bin/runc", "runc", "crun"]`.

tariq1890: > Also. Should we expose this as a user-configurable value?

No, I think we should be falling back to the default runtime/current runtime (the runtime handler used by the toolkit pod) when rendering the nvidia-container-runtime thin wrapper binary

PR #1372, opened 2025-10-21
Make systems/nvdevices package public

elezar:
This change promotes the system/nvdevices package to public so that it can directly be consumed by mig-parted.


elezar:
This was superceded by https://github.com/NVIDIA/nvidia-container-toolkit/pull/1577

PR #1382, opened 2025-10-23
Fix handling of existing imports in containerd

elezar:
This change fixes the handling of existing imports in containerd config files. Without this fix the existing imports were dropped and replaced with the import directive for the generated drop-in file.

Before this change:
```
$ cat test-config.toml
imports = ["/foo/bar/*.toml"]
version = 2
```

Run:
```
nvidia-ctk runtime configure --runtime=containerd --config=test-config.toml --drop-in-config=$(pwd)/99-nvidia.toml
```

```
$  cat test-config.toml
imports = ["/Users/elezar/dev/container-toolkit/*.toml"]
version = 2
```

With this change:
```
$ cat test-config.toml
imports = ["/foo/bar/*.toml", "/Users/elezar/dev/container-toolkit/*.toml"]
version = 2
```

### DIFF
--- a/cmd/nvidia-ctk/runtime/configure/configure_test.go
+++ b/cmd/nvidia-ctk/runtime/configure/configure_test.go
@@ -44,6 +44,36 @@ func TestConfigureLifecycle(t *testing.T) {
 	}{
 		// Containerd v2 test cases
 		{
+			description: "containerd: config exists with imports",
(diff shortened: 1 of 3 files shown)


tariq1890 (approved):

tariq1890:
LGTM. Thanks @elezar !

Issue #123, opened 2021-10-11
nvidia-container-runtime doesn't respect container /etc/ld.so.conf paths


k0nstantinv:
I have configured nvidia-container-runtime with a proper docker config. I have no question about how to configure it. Docker is configured and working
```
host# cat /etc/docker/daemon.json
{
    "live-restore": true,
    "default-runtime": "nvidia",
    "runtimes": {
        "nvidia": {
            "path": "/usr/bin/nvidia-container-runtime",
            "runtimeArgs": []
        }
    }
}
```
Seems like I'm getting a bug

I have a custom Debian "buster" image with python 3.7.9. Another version of python 3.7.3 is also there by default. So python libs are located at `/usr/local/lib` and `/usr/lib` respectively

Here are my ld.so.confs inside the container instance based on that image

```
container# cat /etc/ld.so.conf
include /etc/ld.so.conf.d/*.conf

container# # cat /etc/ld.so.conf.d/*
/usr/local/cuda-10.2/targets/x86_64-linux/lib
/usr/local/lib
...
```
If I run container without nvidia-container-runtime I get the proper image version of python - 3.7.9. When nvidia-container-runtime is enabled it makes something with libs that ends up in version mismatch
```
host# docker run --rm 'image:3.7.9' sh -cx 'python --version && ldconfig && python --version'
+ python --version
Python 3.7.3
+ ldconfig
+ python --version
Python 3.7.9
```
As you can see the versions have changed

A diff before and right after running `ldconfig`:

```
host# docker run --rm 'image:3.7.9' sh -c 'ldconfig -p | grep python'

	libpython3.7m.so.1.0 (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libpython3.7m.so.1.0
	libpython3.7m.so.1.0 (libc6,x86-64) => /usr/local/lib/libpython3.7m.so.1.0
	libpython3.7m.so (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libpython3.7m.so
	libpython3.7m.so (libc6,x86-64) => /usr/local/lib/libpython3.7m.so
	libpython3.so (libc6,x86-64) => /usr/local/lib/libpython3.so
	libpython2.7.so.1.0 (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libpython2.7.so.1.0
	libpython2.7.so (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libpython2.7.so
[...]

PR #1403, opened 2025-10-31
Filter already tracked directories from ldcache update

elezar:
This change ensures that directories that are already included in the ld.conf are not added to the NVCR-specific config files. The intent is to not inadventently promote existing directories to a higher priority.

See #123 where this was already an issue in the libnvidia-container-based implementation.

## Testing

build the following docker image
```
docker build -t libordering \
            - <<EOF
FROM ubuntu
ENV NVIDIA_VISIBLE_DEVICES=all
RUN mkdir -p /extra/lib
RUN cp /usr/lib/$(uname -m)-linux-gnu/libc.so.? /extra/lib/
RUN echo "/extra/lib" > /etc/ld.so.conf.d/00-xxx.conf
RUN ldconfig
EOF
```

Without the `nvidia` runtime:
```
$ docker run --rm -ti --runtime=runc libordering bash -c "ldconfig -p | grep libc.so."
        libc.so.6 (libc6,AArch64) => /extra/lib/libc.so.6
        libc.so.6 (libc6,AArch64) => /lib/aarch64-linux-gnu/libc.so.6
```

Without this change:
```
$ docker run --rm -ti --runtime=nvidia libordering bash -c "ldconfig -p | grep libc.so."
        libc.so.6 (libc6,AArch64) => /usr/lib/aarch64-linux-gnu/libc.so.6
        libc.so.6 (libc6,AArch64) => /extra/lib/libc.so.6
```
(Note that the `libc.so.6` in `/usr/lib/aarch64-linux-gnu` takes precedence).

With this change:
```
$ docker run --rm -ti --runtime=nvidia libordering bash -c "ldconfig -p | grep libc.so."
        libc.so.6 (libc6,AArch64) => /extra/lib/libc.so.6
        libc.so.6 (libc6,AArch64) => /lib/aarch64-linux-gnu/libc.so.6
```

### DIFF
--- a/internal/ldconfig/ldconfig.go
+++ b/internal/ldconfig/ldconfig.go
@@ -18,11 +18,13 @@
 package ldconfig
 
 import (
+	"bufio"
 	"flag"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/config"
@@ -38,9 +40,11 @@ const (
 )
 
 type Ldconfig struct {
-	ldconfigPath string
-	inRoot       string
-	directories  []string
+	ldconfigPath          string
+	inRoot                string
+	isDebianLikeHost      bool
+	isDebianLikeContainer bool
+	directories           []string
 }
 
 // NewRunner creates an exec.Cmd that can be used to run ldconfig.
@@ -49,6 +53,9 @@ func NewRunner(id string, ldconfigPath string, containerRoot string, additionala
 		id,
 		"--ldconfig-path", strings.TrimPrefix(config.NormalizeLDConfigPath("@"+ldconfigPath), "@"),
 		"--container-root", containerRoot,
+	}
+	if isDebian() {
+		args = append(args, "--is-debian-like-host")
 	}
 	args = append(args, additionalargs...)
 
@@ -66,6 +73,10 @@ func NewRunner(id string, ldconfigPath string, containerRoot string, additionala
 //	--ldconfig-path=LDCONFIG_PATH	the path to ldconfig on the host
 //	--container-root=CONTAINER_ROOT	the path in which ldconfig must be run
 //
+// The following flags are optional:
+//
+//	--is-debian-like-host	Indicates that the host system is debian-based.
+//
 // The remaining args are folders where soname symlinks need to be created.
 func NewFromArgs(args ...string) (*Ldconfig, error) {
 	if len(args) < 1 {
@@ -74,6 +85,7 @@ func NewFromArgs(args ...string) (*Ldconfig, error) {
 	fs := flag.NewFlagSet(args[1], flag.ExitOnError)
 	ldconfigPath := fs.String("ldconfig-path", "", "the path to ldconfig on the host")
 	containerRoot := fs.String("container-root", "", "the path in which ldconfig must be run")
+	isDebianLikeHost := fs.Bool("is-debian-like-host", false, "the hook is running from a Debian-like host")
 	if err := fs.Parse(args[1:]); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


internal/ldconfig/ldconfig.go
```diff
@@ -75,21 +76,81 @@ func (l *Ldconfig) UpdateLDCache(directories ...string) error {
...
 
-	if err := createLdsoconfdFile(ldsoconfdFilenamePattern, directories...); err != nil {
+	if err := createLdsoconfdFile(ldsoconfdFilenamePattern, filteredDirectories...); err != nil {
 		return fmt.Errorf("failed to update ld.so.conf.d: %w", err)
 	}
 
 	return SafeExec(ldconfigPath, args, nil)
 }
 
+func (l *Ldconfig) filterDirectories(configFilePath string, directories ...string) ([]string, error) {
```
ArangoGutierrez: Question: If the function doesn't need to access or modify the state of an Ldconfig object, why not to be a regular standalone function?

internal/ldconfig/ldconfig.go
```diff
@@ -75,21 +76,81 @@ func (l *Ldconfig) UpdateLDCache(directories ...string) error {
...
+					return nil, err
+				}
+				ldconfigFilenames = append(ldconfigFilenames, includes...)
+			default:
+				ldconfigDirs[line] = ldconfigFilename
+			}
+		}
+	}
+
+	var filtered []string
```
ArangoGutierrez:
```suggestion
    filtered := make([]string, 0, len(directories))
```

ArangoGutierrez (approved):
LGTM
2 non Blocking comments

internal/ldconfig/ldconfig.go
elezar: Do you have a justification for this? I would prefer that `filtered` be a `nil` slice if we don't have any directories to include.

internal/ldconfig/ldconfig.go
elezar: I was originally accessing a member, but I can remove it.

internal/ldconfig/ldconfig.go
ArangoGutierrez: That's Ok, I thought having it initialised is better, but is not a blocking comment.

internal/ldconfig/ldconfig.go
```diff
@@ -75,21 +76,81 @@ func (l *Ldconfig) UpdateLDCache(directories ...string) error {
 		return err
 	}
 
+	// Explicitly specify using /etc/ld.so.conf since the host's ldconfig may
+	// be configured to use a different config file by default.
+	configFilePath := "/etc/ld.so.conf"
+	filteredDirectories, err := filterDirectories(configFilePath, directories...)
```
jasonzlai: (Jason from Uber AI Infra here - we filed the ticket) What will be passed in as `directories`?

elezar: `directories` contains the container paths of parent directories for any CUDA user-mode driver libraries that are mounted from the host. Note that at present the contianer paths of the libraries are the same as the host path.

On an ubuntu based system the list is typically:
```
/usr/lib/${ARCH_SPECIFIC_LIB_DIR}/
/usr/lib/${ARCH_SPECIFIC_LIB_DIR}/vdpau
```
which results in the following CDI hook:
```
        - hookName: createContainer
          path: /usr/bin/nvidia-cdi-hook
          args:
            - nvidia-cdi-hook
            - update-ldcache
            - --folder
            - /lib/x86_64-linux-gnu
            - --folder
            - /lib/x86_64-linux-gnu/vdpau
          env:
            - NVIDIA_CTK_DEBUG=false
```

If you wanted to know what the list is on your system, you could run the `nvidia-ctk cdi generate` command and check the output for the generated `update-ldcache` hook.

elezar:
@jasonzlai do you have a simple reproducer Dockerfile that we could add to our test suite to address this?

PR #1409, opened 2025-11-03
Fix trigger of CDI refresh service

### LINKED ISSUE #1395: Stopped at an ExecCondition to run nvidia-cdi-refresh.service
Since the last update of `nvidia-container-toolkit` (`v1.18.0`), I wanted to follow the [manual](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html#manual-cdi-specification-generation) and update the CDI, or at least generate it in my case. So I ran the following command:

```bash
sudo systemctl restart nvidia-cdi-refresh.service
```

The problem is that I'm blocked by a condition in the file `/etc/systemd/system/nvidia-cdi-refresh.service`.
Here it is:

```
ExecCondition=/usr/bin/grep -qE '/nvidia.ko' /lib/modules/%v/modules.dep
```

As shown by this command, I have the NVIDIA drivers installed on my computer, and everything works fine with Docker containers.

```bash
$> lsmod | grep nvidia
nvidia_uvm           1540096  4
nvidia_drm             77824  0
nvidia_modeset       1314816  1 nvidia_drm
video                  65536  1 nvidia_modeset
drm_kms_helper        212992  4 mgag200,nvidia_drm
nvidia              56860672  29 nvidia_uvm,nvidia_modeset
drm                   614400  6 drm_kms_helper,drm_shmem_helper,nvidia,mgag200,nvidia_drm
```

Just in case, I updated the `modules.dep` file using the following command:

```bash
sudo depmod -a
```

However, this still did not fulfill the condition.

Here is the content of the `module.dep` file with a less strict `grep` command:

```bash
$> cat /lib/modules/$(uname -r)/modules.dep | grep "nvidia"
kernel/drivers/net/ethernet/nvidia/forcedeth.ko:
kernel/drivers/usb/typec/altmodes/typec_nvidia.ko: kernel/drivers/usb/typec/altmodes/typec_displayport.ko kernel/drivers/usb/typec/typec.ko kernel/drivers/gpu/drm/drm.ko
kernel/drivers/platform/x86/nvidia-wmi-ec-backlight.ko: kernel/drivers/acpi/video.ko kernel/drivers/platform/x86/wmi.ko
updates/dkms/nvidia-current-peermem.ko:
updates/dkms/nvidia-current-modeset.ko: updates/dkms/nvidia-current.ko kernel/drivers/acpi/video.ko kernel/drivers/platform/x86/wmi.ko kernel/drivers/gpu/drm/drm.ko
updates/dkms/nvidia-current-drm.ko: updates/dkms/nvidia-current-modeset.ko updates/dkms/nvidia-current.ko kernel/drivers/acpi/video.ko kernel/drivers/platform/x86/wmi.ko kernel/drivers/gpu/drm/drm_kms_helper.ko kernel/drivers/gpu/drm/drm.ko
updates/dkms/nvidia-current-uvm.ko: updates/dkms/nvidia-current.ko kernel/drivers/gpu/drm/drm.ko
updates/dkms/nvidia-current.ko: kernel/drivers/gpu/drm/drm.ko
```

So I wonder if the prerequisites are too strict and not permissive enough. Because in my case, the Nvidia drivers were installed correctly using the method provided by Debian with the official packages.

In terms of my configuration, I am running `Debian 12 Bookworm` with `kernel 6.1.0-40-amd64` and I installed the NVIDIA drivers as recommended by Debian [here](https://wiki.debian.org/fr/NvidiaGraphicsDrivers#Debian_12_.2BAKs_Bookworm_.2BALs-).

elezar:
This change also allows the nvidia-cdi-refresh.service to execute when the nvidia-current kernel module is present instead of restricting this to the nvidia
module.

Fixes #1395

### DIFF
--- a/deployments/systemd/nvidia-cdi-refresh.service
+++ b/deployments/systemd/nvidia-cdi-refresh.service
@@ -23,7 +23,7 @@ Type=oneshot
 # Values from Environment will be replaced if defined in EnvironmentFile
 Environment=NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/var/run/cdi/nvidia.yaml
 EnvironmentFile=-/etc/nvidia-container-toolkit/nvidia-cdi-refresh.env
-ExecCondition=/usr/bin/grep -qE '/nvidia.ko' /lib/modules/%v/modules.dep
+ExecCondition=/usr/bin/grep -qE '/(nvidia|nvidia-current)\.ko[:]' /lib/modules/%v/modules.dep
 ExecStart=/usr/bin/nvidia-ctk cdi generate
 CapabilityBoundingSet=CAP_SYS_MODULE CAP_SYS_ADMIN CAP_MKNOD
 


deployments/systemd/nvidia-cdi-refresh.service
```diff
@@ -23,7 +23,7 @@ Type=oneshot
 # Values from Environment will be replaced if defined in EnvironmentFile
 Environment=NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/var/run/cdi/nvidia.yaml
 EnvironmentFile=-/etc/nvidia-container-toolkit/nvidia-cdi-refresh.env
-ExecCondition=/usr/bin/grep -qE '/nvidia.ko' /lib/modules/%v/modules.dep
+ExecCondition=/usr/bin/grep -qE '/nvidia.ko|/nvidia-current.ko' /lib/modules/%v/modules.dep
```
ArangoGutierrez: on https://github.com/NVIDIA/nvidia-container-toolkit/issues/1395 @tr4cks proposes

```bash
ExecCondition=/usr/bin/grep -qE '/nvidia(-current)?\.ko' /lib/modules/%v/modules.dep
```
What do you think

elezar: I don't really have a strong feeling on this -- although I though the explicit OR to be more easily readable instead of trying to parse a more complex regex. It should strictly speaking be:
```suggestion
ExecCondition=/usr/bin/grep -qE '/nvidia\.ko|/nvidia-current\.ko' /lib/modules/%v/modules.dep
```
Anyway.

tr4cks: I agree, it's more meaningful, especially if you need to add something else in the future.

However, I might have done it this way. Without forgetting the $ at the end if you don't want it to match `nividia.ko.someting-else`. But maybe that's what you want, I don't know, I don't have the context.

```bash
ExecCondition=/usr/bin/grep -qE '/(nvidia|nvidia-current)\.ko$' /lib/modules/%v/modules.dep
```

elezar: On one system I checked, I don't think that the `$` will do what we wanted:
```
$ cat modules.dep | grep -E '/nvidia\.ko\:'
updates/dkms/nvidia.ko:
```
Note the trailing `.ko:`. 

I think what we may want is:
```
$ cat /lib/modules/6.11.0-1013-nvidia-64k/modules.dep | grep -E '/(nvidia|nvidia-current)\.ko[:]'
updates/dkms/nvidia.ko:
```

elezar: Actually: looking at https://man7.org/linux/man-pages/man5/modules.dep.5.html it is stated that:

> These files are not intended for editing or use by any additional
       utilities as their format is subject to change in the future. You
       should use the [modinfo(8)](https://man7.org/linux/man-pages/man8/modinfo.8.html) command to obtain information about
       modules in a future proof and compatible fashion rather than
       touching these files.

Does `modinfo nvidia --field version` work as expected on your system where the module file is `nvidia-current.ko`?

I get:
```
$ modinfo nvidia --field version
580.82.07
```

tr4cks: No, it doesn't work.

```bash
$ modinfo nvidia --field version
modinfo: ERROR: Module nvidia not found.
```

Here are the available modules:

```bash
$ modinfo nvidia-
nvidia-current           nvidia-current-modeset   nvidia-current-uvm
nvidia-current-drm       nvidia-current-peermem   nvidia-wmi-ec-backlight
$ modinfo nvidia-
```

So forced to do this with `nvidia-current`:

```bash
$ modinfo nvidia-current --field version
535.247.01
```

In addition, the executable is located in the `/usr/sbin/modinfo` path, and I am not sure if it is always in the same place on all systems—to be verified.

```bash
$ whereis modinfo
modinfo: /usr/sbin/modinfo /usr/share/man/man8/modinfo.8.gz
```

tr4cks: After doing some research, I think your method is probably the best compromise. It would also be possible to search for modules in the file system, but that would require executing more complicated commands with pipes.

It would probably be necessary to iterate and track the changes.

Issue #1398, opened 2025-10-29
Segmentation fault in `nvsandboxutils._Cfunc_nvSandboxUtilsShutdown()` on Ubuntu 22.04 (v1.18.0)


timpalpant:
After installing `nvidia-container-toolkit=1.18.0-1` I encounter the following crash on Google Cloud `n1-standard-4` instance with V100 GPU. Downgrading to `1.17.9-1` resolves it.

```
$ docker run -it --gpus=all alpine                   
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: nvidia-container-runtime did not terminate successfully: exit status 2: free(): invalid pointer
SIGABRT: abort
PC=0x723e54e969fc m=0 sigcode=18446744073709551610
signal arrived during cgo execution

goroutine 1 gp=0xc000002380 m=0 mp=0xc6ad60 [syscall]:
runtime.cgocall(0x6d4990, 0xc00015ae60)
	/usr/local/go/src/runtime/cgocall.go:167 +0x4b fp=0xc00015ae38 sp=0xc00015ae00 pc=0x48378b
github.com/NVIDIA/nvidia-container-toolkit/internal/nvsandboxutils._Cfunc_nvSandboxUtilsShutdown()
	_cgo_gotypes.go:171 +0x45 fp=0xc00015ae60 sp=0xc00015ae38 pc=0x6762e5
github.com/NVIDIA/nvidia-container-toolkit/internal/nvsandboxutils.nvSandboxUtilsShutdown(...)
	/go/src/nvidia-container-toolkit/internal/nvsandboxutils/nvsandboxutils.go:42
github.com/NVIDIA/nvidia-container-toolkit/internal/nvsandboxutils.(*library).Shutdown(0xc000173b90)
	/go/src/nvidia-container-toolkit/internal/nvsandboxutils/impl.go:36 +0x19 fp=0xc00015ae78 sp=0xc00015ae60 pc=0x6768f9
github.com/NVIDIA/nvidia-container-toolkit/pkg/nvcdi.(*nvmllib).tryShutdown(0xc0000d7520)
	/go/src/nvidia-container-toolkit/pkg/nvcdi/lib-nvml.go:229 +0x33 fp=0xc00015aec8 sp=0xc00015ae78 pc=0x6a77d3
[...]

PR #1419, opened 2025-11-04
Allow nvcdi feature flags to be set for the `jit-cdi` mode

elezar:
This changs allows nvcdi feature flags to be set for the (now default) `jit-cdi` mode. This can be used as a workaround for issues such as #1398 where there are issues with a specific driver version.

With the change from https://github.com/NVIDIA/k8s-device-plugin/pull/1495 it will also be possible to set this in the device plugin.

To opt-in to this feature for the `jit-cdi` mode, run:
```
sudo nvidia-ctk config --in-place --set nvidia-container-runtime.modes.jit-cdi.nvcdi-feature-flags=enable-nvsandboxutils 
```

To opt-in for `nvidia-ctk cdi generate` run:
```
nvidia-ctk cdi generate --feature-flag=enable-nvsandboxutils
```

To opt-in for the `nvidia-cdi-refresh.service` add:
```
NVIDIA_CTK_CDI_GENERATE_FEATURE_FLAGS=enable-nvsandboxutils
```
to `/etc/nvidia-container-toolkit/nvidia-cdi-refresh.env`.

### DIFF
--- a/pkg/nvcdi/api.go
+++ b/pkg/nvcdi/api.go
@@ -80,9 +80,11 @@ const (
 	// FeatureEnableExplicitDriverLibraries enables the inclusion of a list of
 	// explicit driver libraries.
 	FeatureEnableExplicitDriverLibraries = FeatureFlag("enable-explicit-driver-libraries")
+
 	// FeatureDisableNvsandboxUtils disables the use of nvsandboxutils when
 	// querying devices.
-	FeatureDisableNvsandboxUtils = FeatureFlag("disable-nvsandbox-utils")
+	FeatureDisableNvsandboxUtils = FeatureFlag("disable-nvsandboxutils")
+
 	// FeatureEnableCoherentAnnotations enables the addition of annotations
 	// coherent or non-coherent devices.
 	FeatureEnableCoherentAnnotations = FeatureFlag("enable-coherent-annotations")
--- a/pkg/nvcdi/options.go
+++ b/pkg/nvcdi/options.go
@@ -183,6 +183,11 @@ func WithFeatureFlags[T string | FeatureFlag](featureFlags ...T) Option {
 			o.featureFlags = make(map[FeatureFlag]bool)
 		}
 		for _, featureFlag := range featureFlags {
+			// The initial release of the FeatureDisableNvsandboxUtils feature
+			// flag included a typo which we handle here.
+			if string(featureFlag) == "disable-nvsandbox-utils" {
+				featureFlag = T(FeatureDisableNvsandboxUtils)
+			}
 			o.featureFlags[FeatureFlag(featureFlag)] = true
 		}
 	}


pkg/nvcdi/api.go
```diff
@@ -80,9 +80,23 @@ const (
 	// FeatureEnableExplicitDriverLibraries enables the inclusion of a list of
 	// explicit driver libraries.
 	FeatureEnableExplicitDriverLibraries = FeatureFlag("enable-explicit-driver-libraries")
+
 	// FeatureDisableNvsandboxUtils disables the use of nvsandboxutils when
 	// querying devices.
+	//
+	// Deprecated: nvsandboxutils is now disabled by default. To opt-in use the
```
ArangoGutierrez: Do we have a policy on how long to keep a deprecated feature before total removal?

elezar: No we don't. 

For this feature specifically, I know that @jgehrcke mentioned using this while working on the dynamic MIG for DRA.

cdesiniotis:
Just a general question -- what is nvsandboxutils used for currently? As in, what parts of the CDI spec does it help us generate?

elezar:
> Just a general question -- what is nvsandboxutils used for currently? As in, what parts of the CDI spec does it help us generate?

It's functionality depends on driver version. Currently we use it here for device node symlinks https://github.com/NVIDIA/nvidia-container-toolkit/blob/754cfa7480d9cef7dc956f30ef25112af44dc788/internal/platform-support/dgpu/nvsandboxutils.go#L66-L107

https://github.com/NVIDIA/nvidia-container-toolkit/pull/825 aimed to refactor the driver discovery so that we could add additional functionality, but it has not made it in.

@cdesiniotis thinking about this again, do we want to have opt-in or opt-out behaviour here. The issues that users are experiencing were limited to the 565 driver branch, and we may not want to roll back things now and lose the experience that we would gain by having this on by default.

cdesiniotis:
The same question crossed my mind when initially reviewing this. Do we know what driver branches / versions besides 565 are affected? 565 is a short-lived branch and is no longer supported, see https://docs.nvidia.com/datacenter/tesla/drivers/supported-drivers-and-cuda-toolkit-versions.html. If the fix is present in all of the active branches (535 / 570 / 580), then I would definitely be in favor of continuing to use `nvsandboxutils` by default and providing `opt-out` behavior.

elezar:
> If the fix is present in all of the active branches (535 / 570 / 580), then I would definitely be in favor of continuing to use nvsandboxutils by default and providing opt-out behavior. 

If I recall correctly, `nvsandboxutils` was only added with a driver version after 535 meaning that this bug is not applicable there. The bug that was triggered in 565 was already fixed in 570 at the point where we discovered it. The 580 branch also includes this fix.

PR #1420, opened 2025-11-04
[no-relnote] Add GitHub action for marking issues and PRs as stale

elezar:
This change adds a GitHub action to mark inactive issues and PRs as stale. A lifecycle/stale label is applied after 90 days before an issue / PR is closed after an additional 30 days of inaction.

Adding a lifecycle/frozen label bypasses these checks.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,29 @@
+name: Stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "21 4 * * *"
+
+jobs:
+  stale:
+    permissions:
+      issues: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


cdesiniotis (approved):

ArangoGutierrez (approved):

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,29 @@
+name: Stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "21 4 * * *"
```
tariq1890, line 5: curious. What's the reasoning behind this cron schedule?

elezar: This is what was configured in the `container-device-interface`. It was an arbitrary value that I copied from another workflow for https://github.com/cncf-tags/container-device-interface/pull/207.

PR #1431, opened 2025-11-06
Fix duplicate CDI spec generation in jit-cdi mode

### LINKED ISSUE #1429: NVIDIA_VISIBLE_DEVICES=void is in the container spec file twice
nvidia-container-toolkit: 1.18.0
containerd: 1.7.22

When I create a GPU Pod with one GPU card like `nvidia.com/l40s: '1'`, and check the container spec file in the node:
`
cat /data/containerd-state/io.containerd.runtime.v2.task/k8s.io/da111ad6430ef318fd9941d1db03b613a5e559cd471f289432dc2125280befd7/config.json
`

the NVIDIA_VISIBLE_DEVICES env is like this, and env NVIDIA_VISIBLE_DEVICES=void has twice:

Image

I check the code, https://github.com/NVIDIA/nvidia-container-toolkit/blob/v1.18.0/internal/modifier/cdi.go#L171-L187 , when devices=[runtime.nvidia.com/gpu=GPU-XXX]，the uniqueModes will be {"auto", "auto"}, so the code will add env NVIDIA_VISIBLE_DEVICES=void twice.

So, my question is:
(1) env NVIDIA_VISIBLE_DEVICES=void should be in the container spec file? Is it a bug?
(2) env NVIDIA_VISIBLE_DEVICES=void has twice, Is it a bug?

elezar:
Fixes #1429 

As a proxy for this we can use a duplicate cuda compat `ld.so.conf.d` file as seen in:
https://github.com/NVIDIA/nvidia-container-toolkit/actions/runs/19136472132/job/54690347765?pr=1431
```
docker when Testing CUDA Forward compatibility [It] should create a single ld.so.conf.d config file
/home/runner/_work/nvidia-container-toolkit/nvidia-container-toolkit/tests/e2e/nvidia-container-toolkit_test.go:202
  [FAILED] Expected
      <[]string | len:3, cap:3>: [
          "-rw-r--r-- 1 root root 28 Nov  6 13:16 /etc/ld.so.conf.d/00-compat-3060868491.conf",
          "-rw-r--r-- 1 root root 28 Nov  6 13:16 /etc/ld.so.conf.d/00-compat-3160397803.conf",
          "",
      ]
  to have length 1
  In [It] at: /home/runner/_work/nvidia-container-toolkit/nvidia-container-toolkit/tests/e2e/nvidia-container-toolkit_test.go:205 @ 11/06/25 08:16:49.078
```

### DIFF
--- a/internal/modifier/cdi.go
+++ b/internal/modifier/cdi.go
@@ -211,7 +211,7 @@ type cdiModeIdentifiers struct {
 func cdiModeIdentfiersFromDevices(devices ...string) *cdiModeIdentifiers {
 	perModeIdentifiers := make(map[string][]string)
 	perModeDeviceClass := map[string]string{"auto": automaticDeviceClass}
-	uniqueModes := []string{"auto"}
+	var uniqueModes []string
 	seen := make(map[string]bool)
 	for _, device := range devices {
 		mode, id := getModeIdentifier(device)
--- a/internal/modifier/cdi_test.go
+++ b/internal/modifier/cdi_test.go
@@ -181,7 +181,7 @@ func Test_cdiModeIdentfiersFromDevices(t *testing.T) {
 			description: "empty device list",
 			devices:     []string{},
 			expected: &cdiModeIdentifiers{
-				modes:             []string{"auto"},
+				modes:             nil,
 				idsByMode:         map[string][]string{},
 				deviceClassByMode: map[string]string{"auto": "gpu"},
 			},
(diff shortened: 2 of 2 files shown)


internal/modifier/cdi.go
```diff
@@ -216,6 +202,35 @@ func newAutomaticCDISpecModifier(logger logger.Interface, cfg *config.Config, de
 	return modifiers, nil
 }
 
+type cdiModeIdentifiers struct {
+	modes             []string
+	idsByMode         map[string][]string
+	deviceClassByMode map[string]string
+}
+
+func cdiModeIdentfiersFromDevices(devices ...string) *cdiModeIdentifiers {
```
ArangoGutierrez, line 211: My only comment on this func is the missing 

```go
		logger.Debugf("Mapped %v to %v: %v", device, mode, id)
```

Are we ok with removing this Debug Log?

elezar: Yes, that was intentional. The logging should have been handled by testing in the first place.

PR #1444, opened 2025-11-11
Fix ldcache update when host and container distributions do not match

elezar:
This change fixes the update of the ldcache when running a non-Debian-like (e.g. RHEL, Fedora) container on a Debian-like (e.g. Debian, Ubuntua) system. Here the ldcache in the container does not initially include system libraries from `/lib64` or `/usr/lib64` since this `ldconfig` from the host that is executed in the container's namespace does not process those paths by default.

## Testing

The expected behaviour when using `runc` directly:
```
$ docker run --rm --runtime=runc redhat/ubi9 bash -c "ldconfig -p | grep libc.so."
        libc.so.6 (libc6,AArch64, OS ABI: Linux 3.7.0) => /lib64/libc.so.6
```

The current behaviour:
```
$ docker run --rm --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all redhat/ubi9 bash -c "ldconfig -p | grep libc.so."
```
(note that `libc.so.6` is not present in the ldcache since `/lib64` (linked to `/usr/lib64`) is not processed).

The updated behaviour:
```
$ docker run --rm --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all redhat/ubi9 bash -c "ldconfig -p | grep libc.so."
        libc.so.6 (libc6,AArch64, OS ABI: Linux 3.7.0) => /lib64/libc.so.6
```

### DIFF
--- a/internal/ldconfig/ldconfig.go
+++ b/internal/ldconfig/ldconfig.go
@@ -37,6 +37,12 @@ const (
 	// higher precedence than other libraries on the system, but lower than
 	// the 00-cuda-compat that is included in some containers.
 	ldsoconfdFilenamePattern = "00-nvcr-*.conf"
+	// ldsoconfdSystemDirsFilenamePattern specifies the filename pattern for the drop-in conf file
+	// that includes the expected system directories for the container.
+	// This is chosen to have a high likelihood of being lexicographically last in
+	// in the list of config files, since system search paths should be
+	// considered last.
+	ldsoconfdSystemDirsFilenamePattern = "zz-nvcr-*.conf"
 	// defaultTopLevelLdsoconfFilePath is the standard location of the top-level ld.so.conf file.
 	// Most container images based on a distro will have this file, but distroless container images
 	// may not.
@@ -63,7 +69,7 @@ func NewRunner(id string, ldconfigPath string, containerRoot string, additionala
 		"--ldconfig-path", strings.TrimPrefix(config.NormalizeLDConfigPath("@"+ldconfigPath), "@"),
 		"--container-root", containerRoot,
 	}
-	if isDebian() {
+	if isDebianLike() {
 		args = append(args, "--is-debian-like-host")
 	}
 	args = append(args, additionalargs...)
@@ -76,7 +82,8 @@ func NewRunner(id string, ldconfigPath string, containerRoot string, additionala
 // This struct is used to perform operations on the ldcache and libraries in a
 // particular root (e.g. a container).
 //
-// args[0] is the reexec initializer function name
+// args[0] is the reexec initializer function name and is required.
+//
 // The following flags are required:
 //
 //	--ldconfig-path=LDCONFIG_PATH	the path to ldconfig on the host
@@ -84,17 +91,22 @@ func NewRunner(id string, ldconfigPath string, containerRoot string, additionala
 //
 // The following flags are optional:
 //
-//	--is-debian-like-host	Indicates that the host system is debian-based.
+//	--is-debian-like-host	Indicates that the host system is debian-like (e.g. Debian, Ubuntu)
+//	                     	as opposed to non-Debian-like (e.g. RHEL, Fedora)
+//	                     	See https://github.com/NVIDIA/nvidia-container-toolkit/pull/1444
 //
 // The remaining args are folders where soname symlinks need to be created.
 func NewFromArgs(args ...string) (*Ldconfig, error) {
 	if len(args) < 1 {
 		return nil, fmt.Errorf("incorrect arguments: %v", args)
 	}
-	fs := flag.NewFlagSet(args[1], flag.ExitOnError)
+	fs := flag.NewFlagSet("ldconfig-options", flag.ExitOnError)
 	ldconfigPath := fs.String("ldconfig-path", "", "the path to ldconfig on the host")
 	containerRoot := fs.String("container-root", "", "the path in which ldconfig must be run")
-	isDebianLikeHost := fs.Bool("is-debian-like-host", false, "the hook is running from a Debian-like host")
+	isDebianLikeHost := fs.Bool("is-debian-like-host", false, `indicates that the host system is debian-based.
+This allows us to handle the case where there are  differences in behavior
+between the ldconfig from the host (as executed from an update-ldcache hook) and
+ldconfig in the container. Such differences include system search paths.`)
 	if err := fs.Parse(args[1:]); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


internal/ldconfig/ldconfig.go
```diff
@@ -39,37 +40,60 @@ const (
...
+//
+// args[0] is the reexec initializer function name
+// The following flags are required:
+//
+//	--ldconfig-path=LDCONFIG_PATH	the path to ldconfig on the host
+//	--container-root=CONTAINER_ROOT	the path in which ldconfig must be run
+//
+// The remaining args are folders where soname symlinks need to be created.
+func NewFromArgs(args ...string) (*Ldconfig, error) {
+	if len(args) < 1 {
```
cdesiniotis, line 100: I am not following this. The previous implementation expected there to be at least 3 args. What changed?

internal/ldconfig/ldconfig.go
```diff
@@ -39,37 +40,60 @@ const (
...
+		return nil, fmt.Errorf("incorrect arguments: %v", args)
+	}
+	fs := flag.NewFlagSet(args[1], flag.ExitOnError)
```
cdesiniotis: Question -- is `args[1]` the right name to give to this flag set? What about something along the lines of `ldconfig-options`?

internal/ldconfig/ldconfig.go
cdesiniotis: Okay I think I am following why this change was made. While processing the flags below we error out if the required flags are not provided...

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -533,4 +533,23 @@ EOF`)
 			Expect(err).ToNot(HaveOccurred())
 		})
 	})
+
+	When("running a container a ubi9 container", Ordered, func() {
```
cdesiniotis: nit: `s/running a container a ubi9 container/running a ubi9 container/`

internal/ldconfig/ldconfig.go
```diff
@@ -37,39 +38,72 @@ const (
...
+//
+// args[0] is the reexec initializer function name
+// The following flags are required:
+//
+//	--ldconfig-path=LDCONFIG_PATH	the path to ldconfig on the host
+//	--container-root=CONTAINER_ROOT	the path in which ldconfig must be run
+//
+// The following flags are optional:
+//
+//	--is-debian-like-host	Indicates that the host system is debian-based.
```
cdesiniotis: To aid a future reader, can we add information about why this flag was introduced? Either a link to an issue / PR or an in-line comment.

internal/ldconfig/ldconfig.go
elezar: You're right. This was definitely a copy-paste error on my part.

internal/ldconfig/ldconfig.go
elezar: I have decided to expand the usage string for the flag instead of adding to the comment.

internal/ldconfig/ldconfig.go
```diff
@@ -247,7 +266,7 @@ func (l *Ldconfig) getLdsoconfDirectories(configFilePath string) (map[string]str
 	return ldconfigDirs, nil
 }
 
-func (l *Ldconfig) getSystemSerachPaths() []string {
+func (l *Ldconfig) getSystemSearchPaths() []string {
 	if l.isDebianLikeContainer {
 		debianSystemSearchPaths()
```
jfroy: I've been debugging issues on my homelab cluster with the latest CTK release, and I just spotted this code here. I think it's missing a `return`. I can send a PR for the existing code, or we can roll that fix into this PR.

jfroy: See #1457

elezar: Thanks for the fix. In order to minimize backport churn, let's add the commit to this PR.

jfroy:
I'm preparing a follow up / alternative to this patch to handle additional cases I've ran into recently. Should have it out as a branch in an hour or so.

internal/ldconfig/ldconfig.go
```diff
@@ -331,22 +355,83 @@ func isDebian() bool {
...
+// path.system_dirs[0x0]="/lib64/"
+// path.system_dirs[0x1]="/usr/lib64/"
 //
-// This list was taken from the output of:
+// On most distributions, including Fedora and derivatives, this yields the following
+// ldconfig system search paths.
 //
-//	docker run --rm -ti redhat/ubi9 /usr/lib/ld-linux-aarch64.so.1 --help | grep -A6 "Shared library search path"
+// TODO: Add other architectures that have custom `add_system_dir` macros (e.g. riscv)
+// TODO: Replace with executing the container's dynamlic linker with `--list-diagnostics`?
```
ArangoGutierrez, line 393: Once we merge this PR we should turn this into GitHub issues to track this TODO's and get them implemented

internal/ldconfig/ldconfig.go
```diff
@@ -331,22 +355,83 @@ func isDebian() bool {
...
+// x86_64
+// path.system_dirs[0x0]="/lib/x86_64-linux-gnu/"
+// path.system_dirs[0x1]="/usr/lib/x86_64-linux-gnu/"
+// path.system_dirs[0x2]="/lib/"
+// path.system_dirs[0x3]="/usr/lib/"
+//
+// This yields the following ldconfig system search paths.
+//
+// TODO: Add other architectures that have custom `add_system_dir` macros (e.g. riscv)
+// TODO: Replace with executing the container's dynamlic linker with `--list-diagnostics`?
```
ArangoGutierrez, line 441: Once we merge this PR we should turn this into GitHub issues to track this TODO's and get them implemented

ArangoGutierrez:
Thanks a lot for your contributions on this topic @jfroy !

internal/ldconfig/ldconfig.go
```diff
@@ -37,6 +37,12 @@ const (
 	// higher precedence than other libraries on the system, but lower than
 	// the 00-cuda-compat that is included in some containers.
 	ldsoconfdFilenamePattern = "00-nvcr-*.conf"
+	// ldsoconfdSystemDirsFilenamePattern specifies the filename pattern for the drop-in conf file
+	// that includes the expected system directories for the container.
+	// This is chosen to have a high likelihood of being lexographically last in
```
jgehrcke: "lexicographically" :)

We could also use the prefix "yolo-" 🤡

jfroy: 🤦

elezar:
> Fascinating, thank you.
> 
> > non-debian (including ubuntu)
> 
> From the perspective of this PR, Ubuntu is actually non-Debian?

No. Ubuntu is debian in this context. Let me update the description to be clearer.

jfroy:
> Fascinating, thank you.
> 
> > non-debian (including ubuntu)
> 
> From the perspective of this PR, Ubuntu is actually non-Debian?

@elezar is that what you meant (I think this is quoting https://github.com/NVIDIA/nvidia-container-toolkit/pull/1444#issue-3611946076)?

@jgehrcke Ubuntu _is_ "debian-like" because `/etc/debian_version` exists on Ubuntu systems. This PR does not change that.

internal/ldconfig/ldconfig.go
elezar: I believe this was me. Updating this and references to Debian and non-Debian for clarity.

elezar:
> @elezar is that what you meant (I think this is quoting [#1444 (comment)](https://github.com/NVIDIA/nvidia-container-toolkit/pull/1444#issue-3611946076))?

Yes, the code also used `Debian` and `non-Debian` in comments (and a function name). I have updated to use `Debian-like` and `non-Debian-like` in the comments with examples of distros to make this clearer.

PR #1457, opened 2025-11-15
Add missing return in `getSystemSerachPaths` for debian like containers

jfroy:
@elezar @cdesiniotis 

Might impact #1444


elezar:
Thanks @jfroy. I have pulled this commit into #1444.

PR #1458, opened 2025-11-15
ldconfig: Determine container "debian-ness" after root pivot

jfroy:
@elezar @cdesiniotis

Might impact https://github.com/NVIDIA/nvidia-container-toolkit/pull/1444


elezar:
Thanks @jfroy. I've also added this commit to #1444.

PR #1459, opened 2025-11-16
ldconfig: Create ld.so.conf file if missing

jfroy:
Distroless container images (e.g. `nvcr.io/nvidia/k8s/dcgm-exporter:4.4.1-4.6.0-distroless`, `nvcr.io/nvidia/gpu-operator:v25.10.0`) do not have a top-level ld.so.conf file. Emitting a drop-in for these container images will do nothing. This patch adds a function to create a "standard" top-level ld.so.conf file if it is missing that includes "standard" /etc/ld.so.conf.d/*.conf drop-in files.

@elezar @cdesiniotis

Might impact #1444

### DIFF
--- a/internal/ldconfig/ldconfig.go
+++ b/internal/ldconfig/ldconfig.go
@@ -37,6 +37,15 @@ const (
 	// higher precedence than other libraries on the system, but lower than
 	// the 00-cuda-compat that is included in some containers.
 	ldsoconfdFilenamePattern = "00-nvcr-*.conf"
+	// defaultTopLevelLdsoconfFilePath is the standard location of the top-level ld.so.conf file.
+	// Most container images based on a distro will have this file, but distroless container images
+	// may not.
+	defaultTopLevelLdsoconfFilePath = "/etc/ld.so.conf"
+	// defaultLdsoconfdDir is the standard location for the ld.so.conf.d drop-in directory. Most
+	// container images based on a distro will have this directory included by the top-level
+	// ld.so.conf file, but some may not. And some container images may not have a top-level
+	// ld.so.conf file at all.
+	defaultLdsoconfdDir = "/etc/ld.so.conf.d"
 )
 
 type Ldconfig struct {
@@ -115,20 +124,22 @@ func (l *Ldconfig) UpdateLDCache() error {
 
 	// Explicitly specify using /etc/ld.so.conf since the host's ldconfig may
 	// be configured to use a different config file by default.
-	const topLevelLdsoconfFilePath = "/etc/ld.so.conf"
-	filteredDirectories, err := l.filterDirectories(topLevelLdsoconfFilePath, l.directories...)
+	filteredDirectories, err := l.filterDirectories(defaultTopLevelLdsoconfFilePath, l.directories...)
 	if err != nil {
 		return err
 	}
 
 	args := []string{
 		filepath.Base(ldconfigPath),
-		"-f", topLevelLdsoconfFilePath,
+		"-f", defaultTopLevelLdsoconfFilePath,
 		"-C", "/etc/ld.so.cache",
 	}
 
-	if err := createLdsoconfdFile(ldsoconfdFilenamePattern, filteredDirectories...); err != nil {
-		return fmt.Errorf("failed to update ld.so.conf.d: %w", err)
+	if err := ensureLdsoconfFile(defaultTopLevelLdsoconfFilePath, defaultLdsoconfdDir); err != nil {
+		return fmt.Errorf("failed to ensure ld.so.conf file: %w", err)
+	}
+	if err := createLdsoconfdFile(defaultLdsoconfdDir, ldsoconfdFilenamePattern, filteredDirectories...); err != nil {
+		return fmt.Errorf("failed to create ld.so.conf.d drop-in file: %w", err)
 	}
 
 	return SafeExec(ldconfigPath, args, nil)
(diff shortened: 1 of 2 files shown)


internal/ldconfig/ldconfig.go
```diff
@@ -179,22 +187,30 @@ func (l *Ldconfig) filterDirectories(configFilePath string, directories ...strin
...
-	if err := os.MkdirAll(ldsoconfdDir, 0755); err != nil {
-		return fmt.Errorf("failed to create ld.so.conf.d: %w", err)
-	}
-
-	configFile, err := os.CreateTemp(ldsoconfdDir, pattern)
-	if err != nil {
-		return fmt.Errorf("failed to create config file: %w", err)
+	var configFile *os.File
+	if _, err := os.Stat(topLevelLdsoconfFilePath); os.IsNotExist(err) {
+		configFile, err = os.OpenFile(topLevelLdsoconfFilePath, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644)
```
elezar: Since we also rely on drop-in files to enable CUDA forward compatibility, would it be better to create an `ld.so.conf` file with the following contents if it doesn't exist:
```
$ cat /etc/ld.so.conf
include /etc/ld.so.conf.d/*.conf
```
Then we don't have to also have a special case for the compat libraries.

internal/ldconfig/ldconfig.go
```diff
@@ -115,19 +124,18 @@ func (l *Ldconfig) UpdateLDCache() error {
...
 
 	args := []string{
 		filepath.Base(ldconfigPath),
-		"-f", topLevelLdsoconfFilePath,
+		"-f", defaultTopLevelLdsoconfFilePath,
 		"-C", "/etc/ld.so.cache",
 	}
 
-	if err := createLdsoconfdFile(ldsoconfdFilenamePattern, filteredDirectories...); err != nil {
+	if err := createLdsoconfdFile(defaultTopLevelLdsoconfFilePath, defaultLdsoconfdDir, ldsoconfdFilenamePattern, filteredDirectories...); err != nil {
```
elezar: Considering the comment below, does it make sense to split this into two:
```golang
if err := ensureLdsoconfFile(defaultTopLevelLdsoconfFilePath, defaultLdsoconfdDir); err != nil {
    ...
}
if err := createLdsoconfdFile(ldsoconfdFilenamePattern, filteredDirectories...); err != nil {
    ...
```

internal/ldconfig/ldconfig.go
jfroy: Yes I agree. I started thinking about that but didn't have time this weekend to write it. I'll send an updated patch.

internal/ldconfig/ldconfig.go
jfroy: Yes, will do that.

internal/ldconfig/ldconfig.go
jfroy: I've seen Debian-based distroless containers that retained a multi-arch drop-in that should also be processed.

cdesiniotis (approved):

internal/ldconfig/ldconfig.go
```diff
@@ -220,6 +227,19 @@ func createLdsoconfdFile(pattern string, dirs ...string) error {
...
+// ensureLdsoconfFile creates a "standard" top-level ld.so.conf file if none exists.
+//
+// The created file will contain a single include statement for "`ldsoconfdDir`/*.conf".
+func ensureLdsoconfFile(topLevelLdsoconfFilePath, ldsoconfdDir string) error {
+	configFile, err := os.OpenFile(topLevelLdsoconfFilePath, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644)
+	if err != nil && !os.IsExist(err) {
+		return fmt.Errorf("failed to create top-level ld.so.conf file: %w", err)
+	}
+	defer configFile.Close()
+	configFile.WriteString("include " + ldsoconfdDir + "/*.conf\n")
```
elezar: One thing about this implementation is that we add this line even if the config file already exists OR already includes this line. Should we we check these conditions (or at least not modify the file if it already exists)?

internal/ldconfig/ldconfig_test.go
```diff
@@ -124,3 +125,124 @@ include INCLUDED_PATTERN*
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			tmpDir := t.TempDir()
+			confFilePath := filepath.Join(tmpDir, "ld.so.conf")
+
+			if tc.existingContent != "" {
+				err := os.WriteFile(confFilePath, []byte(tc.existingContent), 0644)
```
elezar: We should be able to replace this with:

```suggestion
				err := os.WriteFile(confFilePath, []byte(tc.existingContent), 0600)
```
to keep the linter happy, since we're creating and running the code as the same user.

internal/ldconfig/ldconfig.go
jfroy: The code _accidentally_ works because if the file exists `os.OpenFile` with `os.O_CREATE|os.O_EXCL` will fail and `configFile` will be `nil`. But I will clarify the code.

internal/ldconfig/ldconfig_test.go
jfroy: Though for drop-in files we `chmod` to `0644`. I can swap to a separate `chmod` call, or add a lint exception.

PR #1461, opened 2025-11-17
Handle multiple GPUs in CDI spec generation from CSV

elezar:
This change allows CDI specs to be generated for multiple
devices when using CSV mode. This can be used in cases where
a Tegra-based system consists of an iGPU and dGPU.

This behavior can be opted out of using the `disable-multiple-csv-devices`
feature flag. This can be specified by adding the
```
            --feaure-flags=disable-multiple-csv-devices
```
command line option to the nvidia-ctk cdi generate command or to the
automatic CDI spec generation by adding
```
    NVIDIA_CTK_CDI_GENERATE_FEATURE_FLAGS=disable-multiple-csv-devices
```
to the /etc/nvidia-container-toolkit/nvidia-cdi-refresh.env file.

### DIFF
--- a/pkg/nvcdi/lib-csv.go
+++ b/pkg/nvcdi/lib-csv.go
@@ -18,9 +18,16 @@ package nvcdi
 
 import (
 	"fmt"
+	"slices"
+	"strconv"
 
 	"tags.cncf.io/container-device-interface/pkg/cdi"
 	"tags.cncf.io/container-device-interface/specs-go"
+
+	"github.com/NVIDIA/go-nvlib/pkg/nvlib/device"
+	"github.com/NVIDIA/go-nvlib/pkg/nvlib/info"
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+	"github.com/google/uuid"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/discover"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/edits"
@@ -29,9 +36,31 @@ import (
 
 type csvlib nvcdilib
 
+type mixedcsvlib nvcdilib
+
 var _ deviceSpecGeneratorFactory = (*csvlib)(nil)
 
+// DeviceSpecGenerators creates a set of generators for the specified set of
+// devices.
+// If NVML is not available or the disable-multiple-csv-devices feature flag is
+// enabled, a single device is assumed.
 func (l *csvlib) DeviceSpecGenerators(ids ...string) (DeviceSpecGenerator, error) {
+	if l.featureFlags[FeatureDisableMultipleCSVDevices] {
+		return l.purecsvDeviceSpecGenerators(ids...)
+	}
+	hasNVML, _ := l.infolib.HasNvml()
+	if !hasNVML {
+		return l.purecsvDeviceSpecGenerators(ids...)
+	}
+	mixed, err := l.mixedDeviceSpecGenerators(ids...)
+	if err != nil {
+		l.logger.Warningf("Failed to create mixed CSV spec generator; falling back to pure CSV implementation: %v", err)
+		return l.purecsvDeviceSpecGenerators(ids...)
+	}
+	return mixed, nil
+}
+
+func (l *csvlib) purecsvDeviceSpecGenerators(ids ...string) (DeviceSpecGenerator, error) {
 	for _, id := range ids {
 		switch id {
 		case "all":
@@ -40,35 +69,42 @@ func (l *csvlib) DeviceSpecGenerators(ids ...string) (DeviceSpecGenerator, error
 			return nil, fmt.Errorf("unsupported device id: %v", id)
 		}
 	}
+	g := &csvDeviceGenerator{
+		csvlib: l,
+		index:  0,
+		uuid:   "",
+	}
+	return g, nil
+}
 
-	return l, nil
+func (l *csvlib) mixedDeviceSpecGenerators(ids ...string) (DeviceSpecGenerator, error) {
+	return (*mixedcsvlib)(l).DeviceSpecGenerators(ids...)
+}
+
+// A csvDeviceGenerator generates CDI specs for a device based on a set of
+// platform-specific CSV files.
+type csvDeviceGenerator struct {
+	*csvlib
+	index int
+	uuid  string
+}
+
+func (l *csvDeviceGenerator) GetUUID() (string, error) {
+	return l.uuid, nil
 }
 
 // GetDeviceSpecs returns the CDI device specs for a single device.
-func (l *csvlib) GetDeviceSpecs() ([]specs.Device, error) {
-	d, err := tegra.New(
-		tegra.WithLogger(l.logger),
-		tegra.WithDriverRoot(l.driverRoot),
-		tegra.WithDevRoot(l.devRoot),
-		tegra.WithHookCreator(l.hookCreator),
-		tegra.WithLdconfigPath(l.ldconfigPath),
-		tegra.WithLibrarySearchPaths(l.librarySearchPaths...),
-		tegra.WithMountSpecs(
-			tegra.Transform(
-				tegra.MountSpecsFromCSVFiles(l.logger, l.csvFiles...),
-				tegra.IgnoreSymlinkMountSpecsByPattern(l.csvIgnorePatterns...),
-			),
-		),
-	)
+func (l *csvDeviceGenerator) GetDeviceSpecs() ([]specs.Device, error) {
+	deviceNodeDiscoverer, err := l.deviceNodeDiscoverer()
 	if err != nil {
-		return nil, fmt.Errorf("failed to create discoverer for CSV files: %v", err)
+		return nil, fmt.Errorf("failed to create discoverer for device nodes from CSV files: %w", err)
 	}
-	e, err := edits.FromDiscoverer(d)
+	e, err := edits.FromDiscoverer(deviceNodeDiscoverer)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create container edits for CSV files: %v", err)
 	}
 
-	names, err := l.deviceNamers.GetDeviceNames(0, uuidIgnored{})
+	names, err := l.deviceNamers.GetDeviceNames(l.index, l)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get device name: %v", err)
 	}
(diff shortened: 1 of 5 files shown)


internal/platform-support/tegra/csv.go
```diff
@@ -17,36 +17,30 @@
...
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/platform-support/tegra/csv"
 )
 
-// newDiscovererFromCSVFiles creates a discoverer for the specified CSV files. A logger is also supplied.
-// The constructed discoverer is comprised of a list, with each element in the list being associated with a
-// single CSV files.
-func (o tegraOptions) newDiscovererFromCSVFiles() (discover.Discover, error) {
-	if len(o.csvFiles) == 0 {
-		o.logger.Warningf("No CSV files specified")
+func (o options) newDiscovererFromMountSpecs() (discover.Discover, error) {
```
ArangoGutierrez: Instead of deleting the func comment, could we update it?

pkg/nvcdi/lib-csv.go
```diff
@@ -18,20 +18,49 @@ package nvcdi
...
+// If NVML is not available or the disable-multiple-csv-devices feature flag is
+// enabled, a single device is assumed.
 func (l *csvlib) DeviceSpecGenerators(ids ...string) (DeviceSpecGenerator, error) {
+	if l.featureFlags[FeatureDisableMultipleCSVDevices] {
+		return l.purecsvDeviceSpecGenerators(ids...)
+	}
+	hasNVML, _ := l.infolib.HasNvml()
+	if !hasNVML {
+		return l.purecsvDeviceSpecGenerators(ids...)
+	}
```
ArangoGutierrez, line 54: Nit: this should be in the order that is documented above

```go
// If NVML is not available or the disable-multiple-csv-devices feature flag is
// enabled, a single device is assumed.
```

ArangoGutierrez (commented):
LGTM, just 2 non-blocking nits

pkg/nvcdi/lib-csv.go
ArangoGutierrez: discussed offline

internal/platform-support/tegra/csv.go
elezar: What do you mean by update it? This function is no longer used and this is why it was removed. We now construct a set of mount specs from CSV files and pass these in when constructing the type that processes the CSV files. This allows us more flexibility such as filtering and adding entities -- which is required to handle multiple devices.

ArangoGutierrez: If I understand correctly, here we are renaming `newDiscovererFromCSVFiles` to `newDiscovererFromMountSpecs` . The new func `newDiscovererFromMountSpecs` has no docu-comment. 

Maybe I'll reword my initial comment, to - let's add a comment to this new func.

elezar: Sure. I'll add an updated comment.

elezar:
I have split two of the commits originally included here into their own PRs: #1511 and #1512

pkg/nvcdi/lib-csv.go
```diff
@@ -80,7 +120,204 @@ func (l *csvlib) GetDeviceSpecs() ([]specs.Device, error) {
...
+		return dgpu, nil
+	}
+
+	return g, nil
+}
+
+func isIntegratedGPUID(id device.Identifier) bool {
+	_, err := uuid.Parse(string(id))
+	return err == nil
+}
```
cdesiniotis, line 298: Question -- would this method not also return true for a discrete GPU identifier? The `isIntegratedGPUID` method name seems to indicate this is unique to discrete GPUs...

EDIT: Okay I think I see why this method is needed. Based on reading other parts of the code, I am assuming `id.IsGpuUUID()` returns false for integrated GPU UUIDs? Some more context would help here.

pkg/nvcdi/lib-csv.go
```diff
@@ -80,7 +120,204 @@ func (l *csvlib) GetDeviceSpecs() ([]specs.Device, error) {
...
+		return false, fmt.Errorf("failed to get PCI info: %v", ret)
+	}
+
+	if pciInfo.Domain != 0 {
+		return false, nil
+	}
+	if pciInfo.Bus != 1 {
+		return false, nil
+	}
+	return pciInfo.Device == 0, nil
```
cdesiniotis, line 322: Question -- does this mean that integrated GPUs (even though they are not attached to the PCI bus) always appear to have a PCI address of `0000:01:00` (domain:bus:device)? 

nit: as a reader, this may be easier to grok if rewritten as

```suggestion
    if pciInfo.Domain == 0 && pciInfo.Bus == 1 && pci.Device == 0 {
      return true, nil
    }
    return false, nil
```

pkg/nvcdi/lib-csv.go
```diff
@@ -80,7 +120,204 @@ func (l *csvlib) GetDeviceSpecs() ([]specs.Device, error) {
...
+			uuid:    uuid,
+			index:   index,
+			// For the CSV case, we include the control device nodes at a
+			// device level.
+			additionalDiscoverers: []discover.Discover{
+				(*nvmllib)(l).controlDeviceNodeDiscoverer(),
+				csvDeviceNodeDiscoverer,
+			},
+			featureFlags: l.featureFlags,
+		})
```
cdesiniotis, line 288: Question -- What are the differences between the device specs generated for dGPUs and iGPUs? Is the addition of control device nodes (e.g. `nvidiactl`, `nvidia-uvm`) the main difference?

pkg/nvcdi/lib-csv.go
```diff
@@ -80,7 +120,204 @@ func (l *csvlib) GetDeviceSpecs() ([]specs.Device, error) {
...
+		// the full GPU.
+		dgpu := (*nvmllib)(l).withInit(&fullGPUDeviceSpecGenerator{
+			nvmllib: (*nvmllib)(l),
+			uuid:    uuid,
+			index:   index,
+			// For the CSV case, we include the control device nodes at a
+			// device level.
+			additionalDiscoverers: []discover.Discover{
+				(*nvmllib)(l).controlDeviceNodeDiscoverer(),
+				csvDeviceNodeDiscoverer,
```
cdesiniotis, line 285: Question -- Conceptually speaking, why do we have to add the `csvDeviceNodeDiscoverer` here? I ask sicne the `fullGPUDeviceSpecGenerator` will, by default, construct and use a device node discoverer [here](https://github.com/NVIDIA/nvidia-container-toolkit/blob/2bffb664439ccdc8f7d8c006bdae665b220196bf/pkg/nvcdi/full-gpu-nvml.go#L162-L167).

pkg/nvcdi/lib-csv.go
elezar: This is one of those heuristics I keep mentioning. Currently, UUIDs for discrete GPUs have a `GPU-` prefix (`MIG-` for MIG devices), and on all Tegra-based systems that I have had access to the UUID is a "standard" UUID for example:
```
$ nvidia-smi -L
GPU 0: Orin (nvgpu) (UUID: 1833c8b5-9aa0-5382-b784-68b7e77eb185)
```

We have been pushing the NVML team for an "IsIntegrated" API, but have not had a commitment.

Let me update the function comment.

pkg/nvcdi/lib-csv.go
elezar: The device specs generated for iGPUs depend entirely on the contents of the `/etc/nvidia-container-runtime/host-files-for-container.d/devices.csv` file that is constructed by the platform team. For example, on an Orin-based system I have:
```
$ cat /etc/nvidia-container-runtime/host-files-for-container.d/devices.csv
dev, /dev/dri/card*
dev, /dev/dri/renderD*
dir, /dev/dri/by-path
dev, /dev/fb0
dev, /dev/fb1
dev, /dev/host1x-fence
dev, /dev/nvhost-as-gpu
dev, /dev/nvhost-ctrl-gpu
dev, /dev/nvhost-ctrl-nvdla0
dev, /dev/nvhost-ctrl-nvdla1
dev, /dev/nvhost-ctrl-pva0
dev, /dev/nvhost-ctxsw-gpu
dev, /dev/nvhost-dbg-gpu
dev, /dev/nvhost-gpu
dev, /dev/nvhost-nvsched-gpu
dev, /dev/nvhost-power-gpu
dev, /dev/nvhost-prof-ctx-gpu
dev, /dev/nvhost-prof-dev-gpu
dev, /dev/nvhost-prof-gpu
dev, /dev/nvhost-sched-gpu
dev, /dev/nvhost-tsg-gpu
dev, /dev/nvgpu/igpu0/as
dev, /dev/nvgpu/igpu0/channel
dev, /dev/nvgpu/igpu0/ctrl
dev, /dev/nvgpu/igpu0/ctxsw
dev, /dev/nvgpu/igpu0/dbg
dev, /dev/nvgpu/igpu0/nvsched
dev, /dev/nvgpu/igpu0/power
dev, /dev/nvgpu/igpu0/prof
dev, /dev/nvgpu/igpu0/prof-ctx
dev, /dev/nvgpu/igpu0/prof-dev
dev, /dev/nvgpu/igpu0/sched
dev, /dev/nvgpu/igpu0/tsg
dev, /dev/nvidia-modeset
dev, /dev/nvidia0
dev, /dev/nvidiactl
dev, /dev/nvmap
dev, /dev/nvsciipc
dev, /dev/v4l2-nvdec
dev, /dev/v4l2-nvenc
```
This file is provided by the `nvidia-l4t-init` package:
```
$ dpkg -S  /etc/nvidia-container-runtime/host-files-for-container.d/devices.csv
nvidia-l4t-init: /etc/nvidia-container-runtime/host-files-for-container.d/devices.csv
```

Note that this includes `/dev/nvidia0` and `/dev/nvidiactl` for this system. In the case of Thor-systems, this would include `/dev/nvidia0`, `/dev/nvidia1`, and `/dev/nvidiactl`. 

For the purpose of this discussion then, the primary difference between the device nodes for the two devices are that the dGPU includes the `/dev/nvidia-uvm` and `/dev/nvidia-uvm-tools` devices that are required for actually running CUDA applications. On a Thor-based system using nvgpu, the container also needs access to the OTHER device nodes mentioned in the CSV file. We currently include all of them, but this list *could* probably be reduced.

Also note that on a Thor-based system that includes a dGPU, the second (rendering) device node for the iGPU is `/dev/nvidia2` and NOT `/dev/nvidia1`.

pkg/nvcdi/lib-csv.go
elezar: We add this because in addition to the "Standard" dGPU device nodes that are returned by the `fullGPUDIscovererer` that we construct as linked, we *ALSO* need access to (at least some of) the device nodes defined in the CSV file. The csvDeviceNodeDiscoverer in this case *should* be filtering out the specific device nodes (e.g. `/dev/nvidia0` and `/dev/nvidia2`) associated with the iGPU.

pkg/nvcdi/lib-csv.go
elezar: Created https://github.com/NVIDIA/nvidia-container-toolkit/pull/1674 to update the comment.

pkg/nvcdi/lib-csv.go
elezar: At least in the case of Thor-based systems that I have had access to, this has been the case. Orin-based systems that I have had access to do not support getting PCI information. I will update the implementation for clarity.

pkg/nvcdi/lib-csv.go
cdesiniotis: Ah okay, this makes sense now! Thanks for the additional context and raising the PR.

PR #1462, opened 2025-11-17
Write system paths to lexicographically last ld.so.conf.d drop-in

jfroy:
This is a followup patch that builds on #1444. It covers more cases by having a slightly wider set of system paths.

@elezar @cdesiniotis

### DIFF
--- a/internal/ldconfig/ldconfig.go
+++ b/internal/ldconfig/ldconfig.go
@@ -46,6 +46,9 @@ const (
 	// ld.so.conf file, but some may not. And some container images may not have a top-level
 	// ld.so.conf file at all.
 	defaultLdsoconfdDir = "/etc/ld.so.conf.d"
+	// ldsoconfdSystemDirsFilenamePattern specifies the filename pattern for the drop-in conf file
+	// that includes the expected system directories for the container.
+	ldsoconfdSystemDirsFilenamePattern = "zz-nvcr-*.conf"
 )
 
 type Ldconfig struct {
@@ -131,7 +134,7 @@ func (l *Ldconfig) UpdateLDCache() error {
 
 	// Explicitly specify using /etc/ld.so.conf since the host's ldconfig may
 	// be configured to use a different config file by default.
-	filteredDirectories, ldconfigDirs, err := l.filterDirectories(defaultTopLevelLdsoconfFilePath, l.directories...)
+	filteredDirectories, err := l.filterDirectories(defaultTopLevelLdsoconfFilePath, l.directories...)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


internal/ldconfig/ldconfig.go
```diff
@@ -303,22 +333,61 @@ func isDebian() bool {
...
+// look for "path.system_dirs". For example
+// `docker run --rm -ti fedora:latest /lib64/ld-linux-x86-64.so.2 --list-diagnostics | grep path.system_dirs`.
 //
-// This list was taken from the output of:
+// On most distributions, including Fedora and derivatives, this yields the following
+// ldconfig system search paths.
 //
-//	docker run --rm -ti redhat/ubi9 /usr/lib/ld-linux-aarch64.so.1 --help | grep -A6 "Shared library search path"
+// TODO: Add other architectures that have custom `add_system_dir` macros (e.g. riscv)
+// TODO: Replace with executing the container's dynamlic linker with `--list-diagnostics`?
```
elezar, line 373: Just a note on this. Since containers are user-supplied, we need to be careful about executing something from the container. This (and the fact that not all containers include `ldconfig`) is the reason that we don't run `ldconfig` form the container.

internal/ldconfig/ldconfig.go
```diff
@@ -131,32 +138,36 @@ func (l *Ldconfig) UpdateLDCache() error {
 		return fmt.Errorf("failed to update ld.so.conf.d: %w", err)
 	}
 
+	// In most cases, the hook will be executing a host ldconfig that may be configured widely
+	// differently from what the container image expects. The common case is Debian vs non-Debian.
+	// But there are also hosts that configure ldconfig to search in a glibc prefix
+	// (e.g. /usr/lib/glibc). To avoid all these cases, append the container's expected system
+	// search paths to the top-level ld.so.conf. This will ensure they get scanned but won't
+	// materially change the scan order.
+	if err := appendSystemSearchPathsToLdsoconf(topLevelLdsoconfFilePath, l.getSystemSearchPaths()...); err != nil {
```
elezar: My one question here would be whether we want to use a drop in file. I suppose the issue there is that we're not guaranteed that the file we create will be lexographically last in the list of included files.

Since we're already processing the list of included files, we could ensure that this is the case for the execution of this hook though.

jfroy: Yes we could do that, being careful to match the sort order of ldconfig. But it felt easier to just append to the top-level config which will always be processed last (but before the built-in (aka system) directories).

internal/ldconfig/ldconfig.go
jfroy: Yeah this is why I wrote it as a question. Two follow up:

- The suggestion/question is about the dynamic linker, not ldconfig. I imagine almost all container images that expect to run our software are going to have it. 
- I probably lack the historical or technical background why this hook is not a `startContainer` hook that could run the container image's ldconfig (if any) or ld.so. Naively, the runtime is about to execute the main container process, which is just as untrusted. Anyways I don't mean to start a big technical discussion with this comment; I am just curious about the way things are and the security posture.

elezar: The reason this is a createContainer hook and not a startContainer hook is that the hooks we run (i.e. `nvidia-cdi-hook`) are not available in the container. This is not to say that it is impossible to ensure that these are injected and available to the container, but it is not something that we have worked on.

Note that logic like "if any" is not really expressible in the current OCI Runtime hook spec which is why we rely on more complex logic being backed into an executable. What we could consider doing is:

1. Run `createContainer` hooks to:
    2. create `/etc/ld.so.conf.d/` drop in files for injected libraries and CUDA compat libraries.
    3. create a hook at a well known path in the container that covers optionally running `ldconfig` / `ldconfig.real` in the container.
4. Run a `startContainer` hook referencing the created hook.

Note that since we would then be running `ldconfig` in the container as a `startContainer` hook, we would be able to leverage the isolation that is already provided by low-level runtimes such as `runc` and it would also simplify the logic around running `ldconfig` since we would not have to handle differences between the host and the container distributions.

One caveat here is that we would NOT be able to handle containers that do not have `ldconfig` in the container -- although in this case we may be able to fall back to a host executable mounted into the container.

internal/ldconfig/ldconfig.go
elezar: The one reason to create a new file is that this provides a clearer signal that these are files that are managed by the NVIDIA Container Toolkit. If we make modifications to the top-level config, a user may inadvertently change this config.

One more thing to note is that this discussion is only relevant to the lifetime of the hook, since once the container is created the `ldconfig` that would be executed would be from the container and not from the host.

jfroy: > The one reason to create a new file is that this provides a clearer signal that these are files that are managed by the NVIDIA Container Toolkit. If we make modifications to the top-level config, a user may inadvertently change this config.

I think it's probably not very likely that the top-level config would be modified, but yeah. We could emit a comment ("# GENERATED DO NOT EDIT"), or we could emit something like "zzz-nv-system-path.conf" in the drop-ins. 

> 
> 
> 
> One more thing to note is that this discussion is only relevant to the lifetime of the hook, since once the container is created the `ldconfig` that would be executed would be from the container and not from the host.

Yes, indeed. I think the patch's approach will not make the ldconfig that comes with container images misbehave.

internal/ldconfig/ldconfig.go
jfroy: It's appealing to consider a `startContainer` approach, but because of containers without an `ldconfig` binary and the need to emit one or more drop-ins, it wouldn't save any code or reduce complexity.

jfroy:
Updated patch to write a drop-in instead of appending to the top-level conf file.

internal/ldconfig/ldconfig.go
```diff
@@ -37,6 +37,9 @@ const (
 	// higher precedence than other libraries on the system, but lower than
 	// the 00-cuda-compat that is included in some containers.
 	ldsoconfdFilenamePattern = "00-nvcr-*.conf"
+	// ldsoconfdSystemDirsFilenamePattern specifies the filename pattern for the drop-in conf file
+	// that includes the expected system directories for the container.
+	ldsoconfdSystemDirsFilenamePattern = "99-nvcr-*.conf"
```
elezar: Should this be `zz-nvcr-*.conf` to ensure that this is lexographcially last?

internal/ldconfig/ldconfig.go
```diff
@@ -220,11 +242,29 @@ func createLdsoconfdFile(pattern string, dirs ...string) error {
 	return nil
 }
 
+func appendSystemSearchPathsToLdsoconf(configFilePath string, dirs ...string) error {
```
elezar: This is only used in tests. Can it be removed?

internal/ldconfig/ldconfig.go
jfroy: It's a cat-mouse. Someone could use zzz-. When I sat down to actually write this and look around what other ecosystems like udev and sysctld were doing, 99- was the more popular/common option. See https://man.archlinux.org/man/sysctl.d.5.en

But sure I can swap to zz.

internal/ldconfig/ldconfig.go
jfroy: Ah yeah. I was tired when I made the change and didn't think about it! The test kept it in-use and it didn't get flagged.

jfroy:
Superseded by #1444

PR #1486, opened 2025-11-20
Update release tooling

elezar:
These changes update release tooling to address some minor issues discovered in the past. Primarily:

* We create a release tracking issue against this REPO and not against the `cloud-native-team` repo.
* We also allow for the scripts to be run against release branches.
* Removed validate_semver function to simplify the scripts.

### DIFF
--- a/hack/prepare-release.sh
+++ b/hack/prepare-release.sh
@@ -20,58 +20,25 @@ this=`basename $0`
 
 usage () {
 cat << EOF
-Usage: $this [-h] [-a] RELEASE_VERSION
+Prepare for an NVIDIA Container Toolkit release
+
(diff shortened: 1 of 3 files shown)


hack/prepare-release.sh
```diff
@@ -20,112 +20,59 @@ this=`basename $0`
...
 
 Example:
 
-  $this {{ VERSION }}
+  $this --previous-version {{ PREVIOUS_VERSION}} --version {{ VERSION }}
 
 EOF
 }
 
-validate_semver() {
```
ArangoGutierrez, line 36: Not needed anymore?  if so let's document this on the PR description.

PR #1491, opened 2025-11-20
Default to strict decoding of OCI runtime spec

elezar:
This change defaults to strict decoding of the OCI
runtime spec to avoid dropping unknown fields when making
OCI runtime modifications.

We also add an allow-unknown-oci-spec-fields feature flag
that can be used to opt-out of this behaviour.


ArangoGutierrez (approved):
LGTM

PR #1496, opened 2025-11-20
Bump release v1.18.1

elezar:
This bumps the version for the v1.18.1 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # NVIDIA Container Toolkit Changelog
 
+## v1.18.1
+- Fix update of ldcache for non-matching host and container distros
+- Use enable-cuda-compat hook when ldcache does not exist
+- Fix update of ldcache if container does not have an ld.so.conf file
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
 # NVIDIA Container Toolkit Changelog
 
+## v1.18.1
+- Fix update of ldcache for non-matching host and container distros
```
tariq1890, line 4: shouldn't we include the runc bump as well? It fixes a high CVE

cdesiniotis:
I see two commits with the same commit message.

CHANGELOG.md
elezar: We generally drop the dependency bumps from the CHANGELOG.

Also, although runc had a CVE, I don't think the runtime was affected by it directly. I would have to check the code we import to see whether any of that was affected.

elezar: I'm definitely open to discussing this though, but don't want to block this release on it.

PR #1498, opened 2025-11-22
implement NRI plugin server to inject management CDI devices

tariq1890:
This PR adds support for an NRI Plugin which will intercept `CreateContainer` events to inject CDI devices to management containers.

Here is a summary of the changes:
- NRI Plugin will listen for `CreateContainer` events, check if the reserved annotation (for e.g. - `"nvidia.cdi.k8s.io/container.{{container-name}}"` ) exists in the pod
- The NRI Plugin is disabled by default.
- When enabled, the `runtime.Setup` method is skipped. Runtime config TOML will not be modified and the runtime will no longer need to be restarted
- If the user is using a container runtime version (for e.g. `containerd v1.7.30`) where NRI and CDI are disabled by default, they will be have to be enabled out of a band as a pre-requisite

### Future Scope
- Should we add a flag to force runtime class creation when NRI plugin is enabled?
- Add mechanisms to restrict who/which pods can inject these devices. One idea is to set a list of allowed namespaces. The NRI Plugin could limit CDI device injections to pods in an allowed namespace.


cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+// Compile-time interface checks
+var (
+	_ nriplugin.Plugin = (*Plugin)(nil)
+)
+
+const (
+	// nodeResourceCDIDeviceKey is the prefix of the key used for CDI device annotations.
+	nodeResourceCDIDeviceKey = "cdi-devices.noderesource.dev"
+	// nriCDIDeviceKey is the prefix of the key used for CDI device annotations.
+	nriCDIDeviceKey = "cdi-devices.nri.io"
```
elezar: Could you comment on why we need two prefixes here? Are these not determined by US? (https://github.com/NVIDIA/gpu-operator/pull/1950/files#diff-e6f52ba1392796db4c79e078d3f1067c50e3bfde9d90f3aaaad3eb3e3f4d84fbR20-R21). Why not only respond to one of them?

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+
+func getAnnotation(annotations map[string]string, mainKey, oldKey, ctr string) []byte {
+	for _, key := range []string{
+		mainKey + "/container." + ctr,
+		oldKey + "/container." + ctr,
+		mainKey + "/pod",
+		oldKey + "/pod",
+		mainKey,
+		oldKey,
+	} {
```
elezar: Instead of having to deal with two keys, could we rather have a single function that we call for each of the keys?

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+		nriSocketPath = defaultNRISocket
+	}
+	_, err := os.Stat(nriSocketPath)
+	if err != nil {
+		return fmt.Errorf("failed to find valid nri socket in %s: %w", nriSocketPath, err)
+	}
+
+	var pluginOpts []nriplugin.Option
+	pluginOpts = append(pluginOpts, nriplugin.WithPluginIdx(nriPluginIdx))
+	pluginOpts = append(pluginOpts, nriplugin.WithSocketPath(nriSocketPath))
```
elezar:
```suggestion
	pluginOpts := []nriplugin.Option{
		nriplugin.WithPluginIdx(nriPluginIdx),
		nriplugin.WithSocketPath(nriSocketPath),
	}
```

pkg/config/engine/containerd/config_v1.go
```diff
@@ -145,3 +145,9 @@ func (c *ConfigV1) EnableCDI() {
 	config.SetPath([]string{"plugins", "cri", "containerd", "enable_cdi"}, true)
 	*c.Tree = config
 }
+
+func (c *ConfigV1) EnableNRI() {
+	config := *c.Tree
+	config.SetPath([]string{"plugins", nriPluginName, "disable"}, false)
+	*c.Tree = config
+}
```
elezar:
```suggestion
func (c *ConfigV1) EnableNRI() {
	(*Config)(c).EnableNRI()
}
```

pkg/config/engine/containerd/containerd.go
```diff
@@ -27,6 +27,7 @@ import (
 const (
 	defaultConfigVersion = 2
 	defaultRuntimeType   = "io.containerd.runc.v2"
+	nriPluginName        = "io.containerd.nri.v1.nri"
```
elezar: If we we only have a single implementation for for `EnableNRI`, then we can just use this string there directly instead of using a constant.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,146 @@
+package nri
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/containerd/nri/pkg/api"
+	nriplugin "github.com/containerd/nri/pkg/stub"
+	"sigs.k8s.io/yaml"
```
elezar: Does it make sense to use a comma-separated list of devices instead of having to parse the YAML?

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	if len(devices) == 0 {
+		p.logger.Debugf("%s: no CDI devices annotated...", containerName(pod, ctr))
+		return nil
+	}
+
+	for _, name := range devices {
+		a.AddCDIDevice(
+			&api.CDIDevice{
```
elezar, line 80: As far as I am aware, this introduces restrictions on compatible containerd / cri-o versions.

cmd/nvidia-ctk-installer/main.go
```diff
@@ -70,13 +75,15 @@ func main() {
 type app struct {
 	logger logger.Interface
 
-	toolkit *toolkit.Installer
+	nriPlugin *nri.Plugin
```
elezar: Do we really need to embed this here? Does it make sense to just instantiate it when required and use a deferred shutdown?

elezar: Something like:
```
diff --git a/cmd/nvidia-ctk-installer/main.go b/cmd/nvidia-ctk-installer/main.go
index 39ee1abe..c725ee3e 100644
--- a/cmd/nvidia-ctk-installer/main.go
+++ b/cmd/nvidia-ctk-installer/main.go
@@ -75,15 +75,13 @@ func main() {
 type app struct {
 	logger logger.Interface
 
-	nriPlugin *nri.Plugin
-	toolkit   *toolkit.Installer
+	toolkit *toolkit.Installer
 }
 
 // NewApp creates the CLI app fro the specified options.
 func NewApp(logger logger.Interface) *cli.Command {
 	a := app{
-		logger:    logger,
-		nriPlugin: nri.NewPlugin(logger),
+		logger: logger,
 	}
 	return a.build()
 }
@@ -229,11 +227,12 @@ func (a *app) Run(ctx context.Context, c *cli.Command, o *options) error {
 	}
 
 	if !o.noDaemon {
-		if o.runtimeOptions.EnableNRI {
-			if err = a.startNRIPluginServer(ctx, o.runtimeOptions); err != nil {
-				a.logger.Errorf("unable to start NRI plugin server: %v", err)
-			}
+		nriPlugin, err := a.startNRIPluginServer(ctx, o.runtimeOptions)
+		if err != nil {
+			a.logger.Errorf("unable to start NRI plugin server: %v", err)
 		}
+		defer nriPlugin.Stop()
+
 		err = a.waitForSignal()
 		if err != nil {
 			return fmt.Errorf("unable to wait for signal: %v", err)
@@ -299,11 +298,15 @@ func (a *app) waitForSignal() error {
 	return nil
 }
 
-func (a *app) startNRIPluginServer(ctx context.Context, opts runtime.Options) error {
+func (a *app) startNRIPluginServer(ctx context.Context, opts runtime.Options) (*nri.Plugin, error) {
+	if !opts.EnableNRI {
+		return nil, nil
+	}
 	a.logger.Infof("Starting the NRI Plugin server....")
 
+	plugin := nri.NewPlugin(a.logger)
 	retriable := func() error {
-		return a.nriPlugin.Start(ctx, opts.NRISocket, opts.NRIPluginIndex)
+		return plugin.Start(ctx, opts.NRISocket, opts.NRIPluginIndex)
 	}
 	var err error
 	for i := 0; i < maxRetryAttempts; i++ {
@@ -318,19 +321,13 @@ func (a *app) startNRIPluginServer(ctx context.Context, opts runtime.Options) er
 	}
 	if err != nil {
 		a.logger.Errorf("Max retries reached %d/%d, aborting", maxRetryAttempts, maxRetryAttempts)
-		return err
+		return nil, err
 	}
-	return nil
+	return plugin, nil
 }
 
 func (a *app) shutdown(pidFile string) {
 	a.logger.Infof("Shutting Down")
-
-	if a.nriPlugin != nil {
-		a.logger.Infof("Stopping NRI plugin server...")
-		a.nriPlugin.Stop()
-	}
-
 	err := os.Remove(pidFile)
 	if err != nil {
 		a.logger.Warningf("Unable to remove pidfile: %v", err)
```

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: We don't. Yes, these should be  defined by us and one annotation should be sufficient. This is taken the from example plugin code of `containerd/nri`

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Given that we've moved to native CDI, the minimum supported containerd is now v1.7. Will that be affected by this change?

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: The plan is to definitely deal with just one annotation key for now. This was taken from the example plugin code

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: thanks

pkg/config/engine/containerd/containerd.go
tariq1890: I prefer having a constant though, it makes it a bit more readable. Not a strong opinion however

cmd/nvidia-ctk-installer/main.go
tariq1890: Agreed, this is much better. Thanks for the suggestion!

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: yes, I am open to that

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Please check now.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: I have updated this to use just one annotation and the prefix is `nvidia.cdi.k8s.io`

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: This is done, please check now.

cmd/nvidia-ctk-installer/main.go
tariq1890: Done

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
elezar: We would need to check when CDI Devices were added to the NRI APIs. The versions of containerd / cri-o that then started responding to these fields would be the minimum versions.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,130 @@
+package nri
```
elezar, line 17: nit: License header.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
elezar: Here was the NRI commit: https://github.com/containerd/nri/commit/c9b4798b3d344d1ef6f7d1897501d8a699fc3e6f
Tagged as v0.7.0. 

We would also need to check which version of containerd includes the related adjustments. (see https://github.com/containerd/containerd/blob/6dc89d8b392dbea8651ab0b141765f196836205c/internal/cri/nri/nri_api_linux.go#L321-L326).

(I wasn't able to find similar code in v1.7 ir v2.0).

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,130 @@
...
+	if len(annotation) == 0 {
+		return nil, nil
+	}
+
+	cdiDevices := strings.Split(annotation, ",")
+	return cdiDevices, nil
+}
+
+func getAnnotation(annotations map[string]string, key, ctr string) string {
+	nriPluginAnnotationKey := fmt.Sprintf("%s/container.%s", key, ctr)
```
elezar: I know that the original example allowed all containers in a pod to automatically have access to devices using a different key. Here, I understand that we're only allowing a SINGLE form, correct?

pkg/config/engine/containerd/containerd.go
elezar: I'm not going to block on this.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
klihub: @elezar CDI injection from an NRI plugin does not work in 1.7.x. I have a [fix for it](https://github.com/klihub/containerd/tree/fixes/1.7.x/nri-cdi-device-injection) which I never filed a PR from. But if it's important enough for you guys, I can file it and then see if we could get it merged.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
elezar: I see that the functionality for providing device access to ALL containers in a pod was removed. Let's keep the implementation simple to start with and not respond to too many signals.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
elezar: Looks good.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
elezar: Thanks @klihub. Having this supported in 1.7.x would simplify things quite a bit on our end in terms of version support. (It also seems to me as if the code to actually make the CDI device modifications is also missing from the `release/2.0` branch).

The diff doesn't look too big, so maybe it's worth seeing what the community thinks about it. Feel free to tag me or @tariq1890 to get input.

klihub: @elezar Here is the filed PR against the 1.7 maintenance branch: https://github.com/containerd/containerd/pull/12650

klihub: @elezar And here is one filed against the 2.0 maintenance branch: https://github.com/containerd/containerd/pull/12651

tariq1890: Thanks @klihub !

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Yes, correct we are only allowing a single form. Please let me know if you have any concerns

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Thanks for catching that

tariq1890: Fixed

pkg/config/engine/containerd/containerd.go
tariq1890: I've changed this to using the string directly as we are referencing the string constant only once right now. Thanks!

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Sure, sounds good

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
klihub: @elezar @tariq1890 Not sure how realistic is to get https://github.com/containerd/containerd/pull/12651 in. 2.0 is EOL'd. 12651 would require a new tagged release, and if that is an absolute no-go (it's usually not that inflexible), then it's a tough call. Anyway, you might want to chime in there and comment.

klihub: > @elezar @tariq1890 Not sure how realistic is to get [containerd/containerd#12651](https://github.com/containerd/containerd/pull/12651) in. 2.0 is EOL'd. 12651 would require a new tagged release, and if that is an absolute no-go (it's usually not that inflexible), then it's a tough call. Anyway, you might want to chime in there and comment.

The v1.7 version of the fix has been now commented as being considered for (a late) inclusion in the upcoming v1.30 release. You might want to chime in there as well.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+// Compile-time interface checks
+var (
+	_ nriplugin.Plugin = (*Plugin)(nil)
+)
+
+const (
+	// nriCDIDeviceDomain is the domain name used to denote an nvidia cdi device
+	nriCDIDeviceDomain = "nvidia.cdi.k8s.io"
+	// nriCDIDeviceKey is the prefix of the key used for CDI device annotations
+	nriCDIDeviceKey = nriCDIDeviceDomain + "/container"
```
elezar: nit: Since we don't use `nriCDIDeviceDomain` elsewhere, using a single constant make things easier to read.
```suggestion
	// nriCDIDeviceKeyPrefix is the prefix of the key used for CDI device annotations
	nriCDIDeviceKeyPrefix = "nvidia.cdi.k8s.io/container."
```
(note that trailing `.`).

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+	// defaultNRISocket represents the default path of the NRI socket
+	defaultNRISocket = "/var/run/nri/nri.sock"
```
elezar: Let's not define this here and instead ensure that we pass in the default value. (We probably want to add the socket as a CLI flag so that we COULD override it.

Nit: We also have `api.DefaultSocketPath` that we could use.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+	annotation := getCDIDeviceFromAnnotation(annotations, ctr)
+	if len(annotation) == 0 {
+		return nil, nil
+	}
+
+	cdiDevices := strings.Split(annotation, ",")
+	return cdiDevices, nil
+}
+
+func getCDIDeviceFromAnnotation(annotations map[string]string, ctr string) string {
```
elezar: nit: Technically we're just getting the annotation value here -- which MAY be a comma-separated list of devices. Do we want to rename the function?

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+	// If there isn't an exact match, we look for a wildcard character match
+	for annotationKey := range annotations {
+		if after, found := strings.CutPrefix(annotationKey, fmt.Sprintf("%s.", nriCDIDeviceKey)); found {
+			if after == "*" {
+				return annotations[annotationKey]
+			}
+		}
+	}
+
+	return ""
```
elezar: We could introduce a new constant that is equal to `"nvidia.cdi.k8s.io/container."` to simplify this quite a bit. Also note that we need not use `strings.CutPrefix`, since we known the full key for the "wildcard" annotation beforehand. (It's also possible to use a recursive implementation, but this is not required).
```suggestion
	nriPluginAnnotationKeyPrefix := nriCDIDeviceKey + "."
    
    containerNriPluginAnnotationKey := nriPluginAnnotationKeyPrefix + ctr
	if value, ok := annotations[containerNriPluginAnnotationKey]; ok {
		return value
	}

    if ctr == "*" {
        return ""
    }
	// If there isn't an exact match, we look for the key that matches ALL containers, indicated by the container name '*':
    return getDeviceFromAnnoations(annotations, "*")
```

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+		return pod.Name + "/" + container.Name
+	}
+	return container.Name
+}
+
+// Start starts the NRI plugin
+func (p *Plugin) Start(ctx context.Context, nriSocketPath, nriPluginIdx string) error {
+	if len(nriSocketPath) == 0 {
+		nriSocketPath = defaultNRISocket
+	}
```
elezar: This isn't required since the plugin implementation already uses the default path. Let's rather remove this conditional entirely.


```suggestion
```

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+	_, err := os.Stat(nriSocketPath)
+	if err != nil {
+		return fmt.Errorf("failed to find valid nri socket in %s: %w", nriSocketPath, err)
+	}
+
+	pluginOpts := []nriplugin.Option{
+		nriplugin.WithPluginIdx(nriPluginIdx),
```
elezar: One thing to note is that the `nriplugin.New` implementation also considers envvars for the plugin name and index.

```
		name:       os.Getenv(api.PluginNameEnvVar),
		idx:        os.Getenv(api.PluginIdxEnvVar),
```

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+
+// Start starts the NRI plugin
+func (p *Plugin) Start(ctx context.Context, nriSocketPath, nriPluginIdx string) error {
+	if len(nriSocketPath) == 0 {
+		nriSocketPath = defaultNRISocket
+	}
+	_, err := os.Stat(nriSocketPath)
+	if err != nil {
+		return fmt.Errorf("failed to find valid nri socket in %s: %w", nriSocketPath, err)
+	}
```
elezar: OK. maybe it is useful to specify the socket if we're adding this check. I would still argue that this is not the place to set the default though.

Let's do the following:
1. Require that a non-empty socket be specified.
2. Require that the socket exist / is queriable.

As a follow-up. Do we need write permissions on the socket? Is that something that we can check with the `os.Stat` call?

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+
+	pluginOpts := []nriplugin.Option{
```
elezar: Is there a reason that we don't also set the plugin name explicitly?

Looking at the `Stub` implementation there is logic to read this from the executable:
```
// ensureIdentity sets plugin index and name from the binary if those are unset.
func (stub *stub) ensureIdentity() error {
	if stub.idx != "" && stub.name != "" {
		return nil
	}

	if stub.idx != "" {
		stub.name = filepath.Base(os.Args[0])
		return nil
	}

	idx, name, err := api.ParsePluginName(filepath.Base(os.Args[0]))
	if err != nil {
		return err
	}

	stub.name = name
	stub.idx = idx

	return nil
}
```
but setting this explicitly is probably clearer.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+		return fmt.Errorf("plugin exited with error: %w", err)
+	}
+	return nil
+}
+
+// Stop stops the NRI plugin
+func (p *Plugin) Stop() {
+	if p != nil && p.stub != nil {
+		p.stub.Stop()
+	}
```
elezar:
```suggestion
    if p == nil || p.stub == nil {
        return
    }
    p.stub.Stop()
```

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -34,6 +34,8 @@ const (
 	// defaultRuntimeName specifies the NVIDIA runtime to be use as the default runtime if setting the default runtime is enabled
 	defaultRuntimeName   = "nvidia"
 	defaultHostRootMount = "/host"
+	defaultNRIPluginIdx  = "10"
+	defaultNRISocket     = "/var/run/nri/nri.sock"
```
elezar: Note that we're duplicating this constant. Does it make sense to use the value from the `nri` API directly?
```
	// DefaultSocketPath is the default socket path for external plugins.
	DefaultSocketPath = "/var/run/nri/nri.sock"
```

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -94,6 +96,27 @@ func Flags(opts *Options) []cli.Flag {
...
+			Value:       true,
+			Sources:     cli.EnvVars("RUNTIME_ENABLE_NRI"),
+		},
+		&cli.StringFlag{
+			Name:        "nri-plugin-index",
+			Usage:       "Specify the plugin index to register to NRI",
+			Value:       defaultNRIPluginIdx,
+			Destination: &opts.NRIPluginIndex,
+			Sources:     cli.EnvVars("RUNTIME_NRI_PLUGIN_INDEX"),
+		},
```
elezar: Looking at the NRI stub implementation, this needs to map to a two-digit string in the range [00, 99]. Does making the following change to treat this as a `uint` make sense. We can then use `fmt.Sprintf("%02d", opts.NRIPluginIndex)` when converting this to an int?
```suggestion
		&cli.UintFlag{
			Name:        "nri-plugin-index",
			Usage:       "Specify the plugin index to register to NRI",
			Value:       10,
			Destination: &opts.NRIPluginIndex,
			Sources:     cli.EnvVars("RUNTIME_NRI_PLUGIN_INDEX"),
			Action: func(ctx context.Context, c *cli.Command, u uint) error {
				if u > 99 {
					return fmt.Errorf("nri-plugin-index must be in the range [0,99]")
				}
				return nil
			},
		},
```

cmd/nvidia-ctk-installer/main.go
```diff
@@ -287,6 +303,31 @@ func (a *app) waitForSignal() error {
...
+func (a *app) startNRIPluginServer(ctx context.Context, opts runtime.Options) (*nri.Plugin, error) {
+	a.logger.Infof("Starting the NRI Plugin server....")
+
+	plugin := nri.NewPlugin(a.logger)
+	retriable := func() error {
+		return plugin.Start(ctx, opts.NRISocket, opts.NRIPluginIndex)
+	}
+	var err error
+	for i := 0; i < maxRetryAttempts; i++ {
+		err = retriable()
```
elezar, line 353: Question: Is it ok that we lose the initial errors and don't log them?

cmd/nvidia-ctk-installer/main.go
```diff
@@ -26,6 +28,9 @@ const (
 	toolkitSubDir            = "toolkit"
 
 	defaultRuntime = "docker"
+
+	retryBackoff     = 2 * time.Second
+	maxRetryAttempts = 5
```
elezar: Since these are only related to the nri plugin, does it make sense to rename them, or declare them as local constants? (in the function that starts the NRI plugin)?

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -34,6 +34,8 @@ const (
 	// defaultRuntimeName specifies the NVIDIA runtime to be use as the default runtime if setting the default runtime is enabled
 	defaultRuntimeName   = "nvidia"
 	defaultHostRootMount = "/host"
+	defaultNRIPluginIdx  = "10"
```
elezar: Question: why `10`?

cmd/nvidia-ctk-installer/main.go
```diff
@@ -216,12 +221,23 @@ func (a *app) Run(c *cli.Command, o *options) error {
 		return fmt.Errorf("unable to install toolkit: %v", err)
 	}
 
-	err = runtime.Setup(c, &o.runtimeOptions, o.runtime)
-	if err != nil {
-		return fmt.Errorf("unable to setup runtime: %v", err)
+	// if NRI is enabled, we don't need to modify the container runtime config TOML.
+	if !o.runtimeOptions.EnableNRI {
```
elezar: The behaviour of this `main` funtion should NOT depend on the `runtimeOptions`. If we wanted to do a quick return in `runtime.Setup` based on the option, we can do that there.

cmd/nvidia-ctk-installer/main.go
```diff
@@ -216,12 +221,23 @@ func (a *app) Run(c *cli.Command, o *options) error {
...
+		err = runtime.Setup(c, &o.runtimeOptions, o.runtime)
+		if err != nil {
+			return fmt.Errorf("unable to setup runtime: %v", err)
+		}
 	}
 
 	if !o.noDaemon {
+		if o.runtimeOptions.EnableNRI {
```
elezar: This change in relation to the comment above seems to indicate that we need two flags to really do what we want to do. The one that we already have controls whether we want to enable `nri` in the runtime. We need another to control whether we want to start the NRI plugin. I would go so far as to say that the socket and index flags should then NOT be associated with the runtime -- unless we expect them to affect the runtime configuration.

As an explaination for this. Assuming that a user has configured their runtime with NRI support and is using the default socket (and assuming that the runtime already supports CDI injection). All we want to do here is start the NRI plugin. We MAY also want to skip the runtime configuration depending on whether we want to register `nvidia` runtimes for any runtime handlers.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,157 @@
...
+	// nriCDIDeviceDomain is the domain name used to denote an nvidia cdi device
+	nriCDIDeviceDomain = "nvidia.cdi.k8s.io"
+	// nriCDIDeviceKey is the prefix of the key used for CDI device annotations
+	nriCDIDeviceKey = nriCDIDeviceDomain + "/container"
+	// defaultNRISocket represents the default path of the NRI socket
+	defaultNRISocket = "/var/run/nri/nri.sock"
+)
+
+type Plugin struct {
+	logger logger.Interface
```
elezar, line 44: Note that the `nri.Stub` implementation has a `Logger` associated with it. Can we convert this logger to an `nri.Logger`?

elezar: Looks like the following type should do what we want:
```
type toNriLogger struct {
	logger.Interface
}

func (l toNriLogger) Debugf(_ context.Context, fmt string, args ...interface{}) {
	l.Interface.Debugf(fmt, args)
}

func (l toNriLogger) Errorf(_ context.Context, fmt string, args ...interface{}) {
	l.Interface.Errorf(fmt, args)
}

func (l toNriLogger) Infof(_ context.Context, fmt string, args ...interface{}) {
	l.Interface.Infof(fmt, args)
}

func (l toNriLogger) Warnf(_ context.Context, fmt string, args ...interface{}) {
	l.Interface.Warningf(fmt, args)
}
```
(there may be a cleaner way to do this though).

elezar (commented):
Thanks @tariq1890.

As discussed yesterday, I think the approach is good in general. It would be good to define what level of control we want to afford users w.r.t configuring runtimes. (e.g. Do we want to allow them to enable NRI (and CDI) *and* configure the runtime classes?).

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Sounds good, I can make the change

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Sure, will do. Thanks for the suggestion

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Does `getCDIDeviceNamesFromAnnotation` work? Open to suggestions here

cmd/nvidia-ctk-installer/container/runtime/runtime.go
tariq1890: I based this off the `device-injector` implementation from the upstream nri samples

cmd/nvidia-ctk-installer/main.go
tariq1890: agreed, will make the change

cmd/nvidia-ctk-installer/main.go
tariq1890: I can move `EnableNRI` up to the top-level options struct. Or introduce a new config object `o.nriOptions`

tariq1890: > The one that we already have controls whether we want to enable nri in the runtime. 

Thanks for pointing this out. My intention is to actually remove the enablement of NRI in the config TOML. I forgot to update that. If we want to enable NRI in the toml and do so via a new flag, we could explore that in a follow-up PR, perhaps?

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Thanks again @klihub. I have tested this PR on a cluster using containerd v1.7.30 and can confirm it works as expected

cmd/nvidia-ctk-installer/main.go
tariq1890: Will do

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Done. please check

cmd/nvidia-ctk-installer/main.go
tariq1890: I have removed the EnableNRI() method. The toolkit will no longer try to edit the config TOML if the NRI is enabled

cmd/nvidia-ctk-installer/main.go
tariq1890: Done

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Done

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Done

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: This is done

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: I have changed the method names, please check now

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: > We could introduce a new constant that is equal to "nvidia.cdi.k8s.io/container." to simplify this quite a bit. Also note that we need not use strings.CutPrefix

Agreed. I have made the changes to directly lookup the necessary keys instead of looping through the entire keyset of the map. Thanks for the suggestion!

cmd/nvidia-ctk-installer/main.go
tariq1890: Done

cmd/nvidia-ctk-installer/container/runtime/runtime.go
tariq1890: Here are the [docs](https://github.com/containerd/nri/tree/main/plugins/device-injector#testing) for the example plugin in the upstream nri repo where they instruct users to provide `10` as the value.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: I've renamed the function

cmd/nvidia-ctk-installer/container/runtime/runtime.go
tariq1890: Thanks, I've implemented this

cmd/nvidia-ctk-installer/main.go
tariq1890: Thanks for catching this; This is now fixed

cmd/nvidia-ctk-installer/main.go
tariq1890: I've removed the code which enables the NRI in the config TOML

tariq1890:
@elezar It looks like `*` is not an allowed character in annotation keys

> The Pod "ubuntu" is invalid: metadata.annotations: Invalid value: "nvidia.cdi.k8s.io/container.&ast;": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')

I am thinking of just using `pod` as the reserved annotation suffix to trigger the injection of the CDI device(s) to every container in the pod.

So examples of supported annotation k-v pairs are as follows:

```
nvidia.cdi.k8s.io/container.nvidia-dcgm-exporter: management.nvidia.com/gpu=all
nvidia.cdi.k8s.io/pod: management.nvidia.com/gpu=all
```

tariq1890:
@elezar Thanks for the thorough review! I have addressed your comments and the PR is ready for another round of review

klihub:
> I am thinking of just using `pod` as the reserved annotation suffix to trigger the injection of the CDI device(s) to every container in the pod.
> 
> So examples of supported annotation k-v pairs are as follows:
> 
> ```
> nvidia.cdi.k8s.io/container.nvidia-dcgm-exporter: management.nvidia.com/gpu=all
> nvidia.cdi.k8s.io/pod: management.nvidia.com/gpu=all
> ```

@tariq1890 :+1: We have a [similar convention]( https://github.com/containerd/nri/blob/2d1ee56b9da5b7e9c2762b6a30cd7f184da43f4c/pkg/plugin/annotations.go#L34) with the addition that an omitted `/pod` suffix  in the key also implies pod scope.

tariq1890:
@klihub Even better! Thank you for this suggestion. Always happy to remove code and reuse library code instead :)

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Done

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,139 @@
...
+			},
+		)
+		pluginLogger.Infof(ctx, "%s: injected CDI device %q...", containerName(pod, ctr), name)
+	}
+
+	return nil
+}
+
+func parseCDIDevices(pod *api.PodSandbox, key, container string) []string {
+	cdiDeviceNames, ok := plugin.GetEffectiveAnnotation(pod, key, container)
```
elezar, line 91: Ok. So we've switched to using the format as implemented in the NRI plugin directly? That's better. We may need to update our documentation / implementation to use `nvidia.cdi.k8s.io/pod` for the cases where all containers need access to the device.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,139 @@
...
+		return pod.Name + "/" + container.Name
+	}
+	return container.Name
+}
+
+// Start starts the NRI plugin
+func (p *Plugin) Start(ctx context.Context, nriSocketPath, nriPluginIdx string) error {
+	pluginOpts := []stub.Option{
+		stub.WithPluginIdx(nriPluginIdx),
+		stub.WithLogger(toNriLogger{p.logger}),
```
elezar, line 112: Not for this PR: Note that the `stub` has a `WithOnClose()` option that allows one to specify a callback when connection is lost. Does it make sense to at least pass a function that logs so that we can collect data on when / how often this happens?

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -213,6 +243,11 @@ func (opts *Options) Validate(logger logger.Interface, c *cli.Command, runtime s
 }
 
 func Setup(c *cli.Command, opts *Options, runtime string) error {
+	// if NRI is enabled, we don't need to modify the container runtime config TOML.
+	if opts.EnableNRI {
+		return nil
+	}
```
elezar: As I have called out, I don't know whether this is something that we want to do in the long term, but I'm OK to have this be the case for the time being. (If we also get the changes from #1521 in then a user can explicitly control this -- or at least opt back in to configuring things).

Not a blocker though.

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -94,6 +97,33 @@ func Flags(opts *Options) []cli.Flag {
...
+			Usage:       "Specify the plugin index to register to NRI",
+			Value:       defaultNRIPluginIdx,
+			Destination: &opts.NRIPluginIndex,
+			Sources:     cli.EnvVars("RUNTIME_NRI_PLUGIN_INDEX"),
+			Action: func(ctx context.Context, c *cli.Command, u uint) error {
+				if u > 99 {
+					return fmt.Errorf("nri-plugin-index must be in the range [0,99]")
+				}
+				return nil
+			},
```
elezar: This should be an argument defined in `main`. It does not affect the runtime config.

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -94,6 +97,33 @@ func Flags(opts *Options) []cli.Flag {
...
+		},
+		&cli.StringFlag{
+			Name:        "nri-socket",
+			Usage:       "Specify the path to the NRI socket file to register the NRI plugin server",
+			Value:       nriapi.DefaultSocketPath,
+			Destination: &opts.NRISocket,
+			Sources:     cli.EnvVars("RUNTIME_NRI_SOCKET"),
+		},
```
elezar: Unless we're using this socket to actually configure the runtime, this should not be defined here but in `main.go`.

cmd/nvidia-ctk-installer/main.go
```diff
@@ -44,6 +49,10 @@ type options struct {
 	sourceRoot  string
 	packageType string
 
+	enableNRIPlugin bool
```
tariq1890, line 52: @elezar Moved the NRI parameters to the top-level  options struct

cmd/nvidia-ctk-installer/main.go
```diff
@@ -216,20 +253,32 @@ func (a *app) Run(c *cli.Command, o *options) error {
...
+
 		err = a.waitForSignal()
 		if err != nil {
 			return fmt.Errorf("unable to wait for signal: %v", err)
 		}
 
-		err = runtime.Cleanup(c, &o.runtimeOptions, o.runtime)
-		if err != nil {
-			return fmt.Errorf("unable to cleanup runtime: %v", err)
+		if !o.enableNRIPlugin {
```
tariq1890, line 277: @elezar This should no longer be required once #1521 gets in. I can make the necessary changes in a follow-up if this PR gets merged before #1521

cmd/nvidia-ctk-installer/container/runtime/runtime.go
```diff
@@ -32,8 +32,9 @@ import (
 const (
 	defaultSetAsDefault = true
 	// defaultRuntimeName specifies the NVIDIA runtime to be use as the default runtime if setting the default runtime is enabled
-	defaultRuntimeName   = "nvidia"
-	defaultHostRootMount = "/host"
+	defaultRuntimeName        = "nvidia"
+	defaultHostRootMount      = "/host"
+	defaultNRIPluginIdx  uint = 10
```
elezar: Is this value still used?

cmd/nvidia-ctk-installer/main.go
elezar: Sure let's keep these separate for now. I can rebase #1521 once we merge this.

elezar (approved):
Thanks @tariq1890 (I just realised that I had a bunch of pending comments here).

This looks good now.

cmd/nvidia-ctk-installer/container/runtime/runtime.go
tariq1890: yes, I've moved this out of the runtimeOptions level

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Sure, I can look into that

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Correct. Yes, I will make sure that this is reflected in our documentation

cmd/nvidia-ctk-installer/container/runtime/runtime.go
tariq1890: I have since moved this conditional out of the method, but agreed that we would no longer need that conditional after #1521 is merged

cmd/nvidia-ctk-installer/container/runtime/runtime.go
tariq1890: Agreed, this has since been moved out of the `runtime` package

cmd/nvidia-ctk-installer/container/runtime/runtime.go
tariq1890: It is not, I will remove this. Thanks!

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: This is done

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,142 @@
...
+		stub.WithOnClose(func() {
+			p.logger.Infof("NRI ttrpc connection to %s down. NRI plugin stopped.", nriSocketPath)
```
cdesiniotis: minor nit:
```suggestion
			p.logger.Infof("NRI ttrpc connection to %s is down. NRI plugin stopped.", nriSocketPath)
```

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -0,0 +1,142 @@
...
+		}),
+	}
+	if len(nriSocketPath) > 0 {
+		_, err := os.Stat(nriSocketPath)
+		if err != nil {
+			return fmt.Errorf("failed to find valid nri socket in %s: %w", nriSocketPath, err)
```
cdesiniotis: minor nit:
```suggestion
			return fmt.Errorf("failed to find valid nri socket at %s: %w", nriSocketPath, err)
```

cdesiniotis (approved):
@tariq1890 thanks for the work on this! This lgtm, just left a few minor typos / nits.

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: Just addressed this

cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
tariq1890: done, thanks

PR #1411, opened 2025-11-03
Update rpm package to have 256bit digests

elezar:
This change uses fpm to update the built rpm packages to have
256bit digests without changing their contents.

This is required in cases where `SHA1` or `MD5` digests are not supported such as FIPS or newer RPM-based distributions. (See #1307).

For the v1.18.0 packages:
```
$ docker run --rm -ti -v $(pwd)/release-v1.18.0-stable:/work -w /work rockylinux:9 bash -c "find . -name '*.rpm' -exec rpm --checksig --verbose {} \;"
./packages/centos7/aarch64/libnvidia-container-static-1.18.0-1.aarch64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
./packages/centos7/aarch64/libnvidia-container-tools-1.18.0-1.aarch64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
./packages/centos7/aarch64/libnvidia-container1-1.18.0-1.aarch64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
./packages/centos7/aarch64/libnvidia-container-libseccomp2-1.18.0-1.aarch64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
./packages/centos7/aarch64/nvidia-container-toolkit-1.18.0-1.aarch64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
./packages/centos7/aarch64/libnvidia-container1-debuginfo-1.18.0-1.aarch64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
./packages/centos7/aarch64/nvidia-container-toolkit-base-1.18.0-1.aarch64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
./packages/centos7/aarch64/nvidia-container-toolkit-operator-extensions-1.18.0-1.aarch64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
./packages/centos7/aarch64/libnvidia-container-devel-1.18.0-1.aarch64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
./packages/centos7/x86_64/libnvidia-container1-1.18.0-1.x86_64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
./packages/centos7/x86_64/libnvidia-container-libseccomp2-1.18.0-1.x86_64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
[...]

PR #1500, opened 2025-11-24
[no-relnote] Ensure that fpm generates sha256 digests

elezar:
Improves #1411

I have verified the following on a FIPS-enabled host:

1. The packages install in a `uibi9` container (after veryfying that the previous packages do not install).
2. The packages still install in a `centos7` container that does not support `sha256` digests by default.


ArangoGutierrez (approved):
LGTM


For reference: https://fpm.readthedocs.io/en/latest/cli-reference.html
```
--rpm-digest md5|sha1|sha256|sha384|sha512
Select a digest algorithm. md5 works on the most platforms.
```

PR #1510, opened 2025-12-03
[no-relnote] Run golang checks on PRs

elezar:
This change enables golang checks on PRs (e.g. by dependabot). This provides basic checks for minimal dependency bumps without requiring an /ok-to-test by a maintainer.

### DIFF
--- a/.github/workflows/golang.yaml
+++ b/.github/workflows/golang.yaml
@@ -16,6 +16,13 @@ name: Golang
 
 on:
   workflow_call: {}
+  pull_request:
+    types:
+      - opened
+      - synchronize
+    branches:
+      - main
+      - release-*
 
 jobs:
   check:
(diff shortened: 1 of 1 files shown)


ArangoGutierrez (approved):

ArangoGutierrez:
We need to set all golang tests to github managed runners as part of this PR

elezar:
> We need to set all golang tests to github managed runners as part of this PR

Could you elaborate? The tests are already using `ubuntu-latest`, but the `build` step uses internal runners. I have added logic to only select the internal runners if the job is not triggered by a PR. (this was after your comment though).

.github/workflows/golang.yaml
```diff
@@ -77,7 +84,7 @@ jobs:
 
   build:
     name: Build
-    runs-on: linux-amd64-cpu4
+    runs-on: ${{ ( github.event_name == 'pull_request' ) && 'ubuntu-latest' || 'linux-amd64-cpu4' }}
```
ArangoGutierrez, line 87: 👍🏽

ArangoGutierrez:
Is it a good idea to back port this to 1.18?

PR #1511, opened 2025-12-03
Add --no-all-device option to nvidia-ctk cd generate

elezar:
Add a `--no-all-device` option to the nvidia-ctk cdi generate command to not generate an `all` device including the specs for all generated devices. This can also be set using the `NVIDIA_CTK_CDI_GENERATE_NO_ALL_DEVICE` environment variable.

### DIFF
--- a/cmd/nvidia-ctk/cdi/generate/generate.go
+++ b/cmd/nvidia-ctk/cdi/generate/generate.go
@@ -73,6 +73,8 @@ type options struct {
 		ignorePatterns []string
 	}
 
+	noAllDevice bool
+
 	// the following are used for dependency injection during spec generation.
 	nvmllib nvml.Interface
 }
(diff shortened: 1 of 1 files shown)


cmd/nvidia-ctk/cdi/generate/generate.go
```diff
@@ -387,13 +395,18 @@ func (m command) generateSpecs(opts *options) ([]generatedSpecs, error) {
...
 		spec.WithEdits(*commonEdits.ContainerEdits),
 		spec.WithFormat(opts.format),
-		spec.WithMergedDeviceOptions(
-			transform.WithName(allDeviceName),
-			transform.WithSkipIfExists(true),
-		),
 		spec.WithPermissions(0644),
 	}
 
+	if !opts.noAllDevice {
```
ArangoGutierrez, line 401: When the "no-all-device" is set the spec will be generated without the option of `transform.WithName(allDeviceName)`

ArangoGutierrez (approved):
LGTM

PR #1512, opened 2025-12-03
Add --device-id flag to nvidia-ctk cdi generate command

elezar:
This change allows device IDs to be specified when generating CDI specs using
the nvidia-ctk cdi generate CDI. The command now accepts a (repeatable) CLI
flag allowing device IDs to be specified. These can be indices or uuids.

### DIFF
--- a/cmd/nvidia-ctk/cdi/generate/generate_test.go
+++ b/cmd/nvidia-ctk/cdi/generate/generate_test.go
@@ -18,6 +18,7 @@ package generate
 
 import (
 	"bytes"
+	"fmt"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


cmd/nvidia-ctk/cdi/generate/generate_test.go
```diff
@@ -452,6 +452,10 @@ containerEdits:
 	for _, tc := range testCases {
 		// Apply overrides for all test cases:
 		tc.options.nvidiaCDIHookPath = "/usr/bin/nvidia-cdi-hook"
+		if tc.options.deviceIDs == nil {
```
ArangoGutierrez, line 477: should we add a test case for a bad/dummy/good device id?

elezar: The validity of a device ID depends on the mode used to generate CDI specs. With that said, I have added a basic test for an invalid NVML id.

PR #1515, opened 2025-12-04
Add enable-cuda-compat on Tegra-based systems

elezar:
This change ensures that tegra-based systems also include a
enable-cuda-compat hook. In the case of Orin devices, this
compat root in the container is set to /usr/local/cuda/compat-orin
to allow these compat libraries to be selected (if present in the
container).

See #1271

### DIFF
--- a/pkg/nvcdi/lib-csv.go
+++ b/pkg/nvcdi/lib-csv.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"slices"
 	"strconv"
+	"strings"
 
 	"tags.cncf.io/container-device-interface/pkg/cdi"
 	"tags.cncf.io/container-device-interface/specs-go"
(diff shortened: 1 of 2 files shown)


cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -44,8 +44,9 @@ type command struct {
 }
 
 type options struct {
-	hostDriverVersion string
-	containerSpec     string
+	cudaCompatContainerRoot string
+	hostDriverVersion       string
+	containerSpec           string
```
ArangoGutierrez: Is there a logical order for these vars? or should we order them alphabetically?

pkg/nvcdi/lib-csv.go
```diff
@@ -57,27 +87,76 @@ func (l *csvlib) GetDeviceSpecs() ([]specs.Device, error) {
...
+		cudaCompatDiscoverer,
+		// The ldcacheUpdateHook is added last to ensure that the created symlinks are included
+		ldcacheUpdateHook,
+	)
+	return d, nil
+}
+
+// cudaCompatDiscoverer returns a discoverer for the CUDA forward compat hook
+// on Tegra-based systems.
+// It the system has NVML available, this is used to determine the driver
```
ArangoGutierrez:
```suggestion
//If the system has NVML available, this is used to determine the driver
```

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
elezar: `containerSpec` is not expected user input and is only used when we want to FORCE some path to a container state. The default for OCI hooks is that this is loaded from stdin.
I can add a comment to `containerSpec` to clarify this, but as it is the options that we expect to change in general are ordered alphabetically.

ArangoGutierrez: so it will look like 

```go
	cudaCompatContainerRoot string
	hostDriverVersion       string
    // comment
	containerSpec           string
```

I'm ok with that

PR #1518, opened 2025-12-05
Fix trigger of CDI refresh service

EdSwarthout:
Allow any module suffix.

Commit 1f5ce738fcc2 ("Fix trigger of CDI refresh service") did not allow for compressed modules with suffixes such as .zst or .gz.

Also fix systemd warning:

/etc/systemd/system/nvidia-cdi-refresh.service:26: Ignoring unknown escape sequences: "/(nvidia|nvidia-current)\.ko[:]"

### DIFF
--- a/deployments/systemd/nvidia-cdi-refresh.service
+++ b/deployments/systemd/nvidia-cdi-refresh.service
@@ -23,7 +23,7 @@ Type=oneshot
 # Values from Environment will be replaced if defined in EnvironmentFile
 Environment=NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/var/run/cdi/nvidia.yaml
 EnvironmentFile=-/etc/nvidia-container-toolkit/nvidia-cdi-refresh.env
-ExecCondition=/usr/bin/grep -qE '/(nvidia|nvidia-current)\.ko[:]' /lib/modules/%v/modules.dep
+ExecCondition=/usr/bin/grep -qE '/(nvidia|nvidia-current)\\.ko' /lib/modules/%v/modules.dep
 ExecStart=/usr/bin/nvidia-ctk cdi generate
 CapabilityBoundingSet=CAP_SYS_MODULE CAP_SYS_ADMIN CAP_MKNOD
 


Lioli7k:
Pattern in the patch looks sketchy, since it would gladly accept something like `updates/nvidia.ko.:` or `updates/nvidia.ko.not!an,extension:`

Here are some alternative pattern suggestions:
```bash
'/(nvidia|nvidia-current)\\.ko(\\.(zst|xz|gz|bz2))?[:]' # Be really specific
'/(nvidia|nvidia-current)\\.ko(\\.[[:alnum:]]+)?[:]'    # Allow any alphanumeric extension
'/(nvidia|nvidia-current)\\.ko'                         # Ignore everything after module name
```

UPD: Double backslashes in the pattern are needed due `systemd` trying to treat single backslash as an escape sequence.

deployments/systemd/nvidia-cdi-refresh.service
```diff
@@ -23,7 +23,7 @@ Type=oneshot
 # Values from Environment will be replaced if defined in EnvironmentFile
 Environment=NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/var/run/cdi/nvidia.yaml
 EnvironmentFile=-/etc/nvidia-container-toolkit/nvidia-cdi-refresh.env
-ExecCondition=/usr/bin/grep -qE '/(nvidia|nvidia-current)\.ko[:]' /lib/modules/%v/modules.dep
+ExecCondition=/usr/bin/grep -qE '/(nvidia|nvidia-current)\\.ko(\\..*)?[:]' /lib/modules/%v/modules.dep
```
elezar: As @Lioli7k points out, would it be simpler to remove the `[:]` instead? Although this was added in #1409 it may be too restrictive.

```suggestion
ExecCondition=/usr/bin/grep -qE '/(nvidia|nvidia-current)\\.ko' /lib/modules/%v/modules.dep
```

ArangoGutierrez: Current patch proposal will match lines in /lib/modules/%v/modules.dep that look like this:

kernel/drivers/video/nvidia.ko:
updates/dkms/nvidia-current.ko:
extra/nvidia.ko.gz: (Compressed module)
weak-updates/nvidia.ko.xz: (Different compression)

Reversing #1409 will also work, as grep will act on the `.ko` even if there is more after it `.ko.eg`, so I am ok with both options

elezar:
> LGTM, properly reverses #1409

@ArangoGutierrez this doesn't reverse #1409. It only removes the check for a trailing `:`. The addition of matching `nvidia-current.ko` in addition to `nvidia.ko` is still valid.

EdSwarthout:
> Pattern in the patch looks sketchy, since it would gladly accept something like `updates/nvidia.ko.:` or `updates/nvidia.ko.not!an,extension:`
> 
> Here are some alternative pattern suggestions:
> 
> ```shell
> '/(nvidia|nvidia-current)\\.ko(\\.(zst|xz|gz|bz2))?[:]' # Be really specific
> ```

I have not seen a specific list of compression methods and this is not future proof.
> ```shell
> '/(nvidia|nvidia-current)\\.ko(\\.[[:alnum:]]+)?[:]'    # Allow any alphanumeric extension
> '/(nvidia|nvidia-current)\\.ko'                         # Ignore everything after module name
> ```

Removing the colon check works, but it is not clear why it was added in https://github.com/NVIDIA/nvidia-container-toolkit/pull/1409. 

Why does this check need to be so specific?  What is wrong with running the refresh service if nvidia is anywhere in the modules.dep?  

What about the day when the open source driver is in the kernel and can be builtin and not a module?

PR #1521, opened 2025-12-08
Add no-runtime-configure option to the toolkit installer

### LINKED ISSUE #1513: Install only mode for the container toolkit
Hi! We've been using the GPU Operator on a self-managed k8s cluster with our own custom AMIs. We are using the Containerd runtime with a custom config that is already embedded in our AMIs. That allows us to avoid having to do the Containerd setup at runtime, which seems to be a bit "brutal" for our use case.

With the migration to `nvidia-ctk-installer` on v1.18, it seems like it is not possible to separate the toolkit installation and the runtime setup. Would it be possible to add a flag to be able to run only the toolkit installation? We're currently using this but we would love to make it available for all nvidia-container-toolkit users: https://github.com/DataDog/nvidia-container-toolkit/commit/fd7c054e98b151ad8bbb02233a01da9c2158aa67
(Note that I'm not opinionated at all on the flag name/implementation specifics)

elezar:
This change adds a `no-runtime-configure` option to the toolkit installer to enable use cases where the NVIDIA Container Toolkit is required without configuring the underlying runtime.

Fixes #1513

### DIFF
--- a/cmd/nvidia-ctk-installer/container/runtime/runtime.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/runtime.go
@@ -228,9 +228,13 @@ type Configurer interface {
 }
 
 type runtime string
+type noopRuntimeConfigurer struct{}
 
 // NewConfigurer is a factory method for creating a runtime configurer.
-func NewConfigurer(name string) Configurer {
+func NewConfigurer(name string, noConfigureRuntime bool, nriEnabled bool) Configurer {
+	if noConfigureRuntime || nriEnabled {
+		return &noopRuntimeConfigurer{}
+	}
 	return runtime(name)
 }
 
@@ -271,4 +275,16 @@ func (r runtime) GetLowlevelRuntimePaths(opts *Options) ([]string, error) {
 	default:
 		return nil, fmt.Errorf("undefined runtime %v", r)
 	}
+}
+
+func (r noopRuntimeConfigurer) Cleanup(_ *cli.Command, _ *Options) error {
+	return nil
+}
+
+func (r noopRuntimeConfigurer) GetLowlevelRuntimePaths(_ *Options) ([]string, error) {
+	return nil, nil
+}
+
+func (r noopRuntimeConfigurer) Setup(_ *cli.Command, _ *Options) error {
+	return nil
 }
--- a/cmd/nvidia-ctk-installer/main.go
+++ b/cmd/nvidia-ctk-installer/main.go
@@ -43,7 +43,6 @@ type options struct {
 	toolkitInstallDir string
 
 	noDaemon    bool
-	runtime     string
 	pidFile     string
 	sourceRoot  string
 	packageType string
@@ -53,7 +52,10 @@ type options struct {
 	nriSocket       string
 
 	toolkitOptions toolkit.Options
-	runtimeOptions runtime.Options
+
+	noRuntimeConfig bool
+	runtime         string
+	runtimeOptions  runtime.Options
 }
 
 func (o options) toolkitRoot() string {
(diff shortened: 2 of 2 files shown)


ArangoGutierrez:
The only comment I had was addressed with a recent commit update.

cdesiniotis:
Note, as currently implemented in the operator, setting `RUNTIME=none` in the toolkit will not work. We process user-supplied envvars [here](https://github.com/NVIDIA/gpu-operator/blob/75d1b966313d4e0fc8c0fdec237e2ca125267619/controllers/object_controls.go#L1315-L1319) but we call [transformForRuntime()](https://github.com/NVIDIA/gpu-operator/blob/75d1b966313d4e0fc8c0fdec237e2ca125267619/controllers/object_controls.go#L1323) right after which unconditionally [sets](https://github.com/NVIDIA/gpu-operator/blob/75d1b966313d4e0fc8c0fdec237e2ca125267619/controllers/object_controls.go#L1332) the `RUNTIME` variable (this will override what the user set). I believe we decided to call transformForRuntime last since any envvars configured in that method are considered to be tightly-coupled with the runtime, and therefore, we do not expect users to set these themselves. I am open to revisiting the behavior here though. 

cc @tariq1890

elezar:
> Note, as currently implemented in the operator, setting `RUNTIME=none` in the toolkit will not work. We process user-supplied envvars [here](https://github.com/NVIDIA/gpu-operator/blob/75d1b966313d4e0fc8c0fdec237e2ca125267619/controllers/object_controls.go#L1315-L1319) but we call [transformForRuntime()](https://github.com/NVIDIA/gpu-operator/blob/75d1b966313d4e0fc8c0fdec237e2ca125267619/controllers/object_controls.go#L1323) right after which unconditionally [sets](https://github.com/NVIDIA/gpu-operator/blob/75d1b966313d4e0fc8c0fdec237e2ca125267619/controllers/object_controls.go#L1332) the `RUNTIME` variable (this will override what the user set). I believe we decided to call transformForRuntime last since any envvars configured in that method are considered to be tightly-coupled with the runtime, and therefore, we do not expect users to set these themselves. I am open to revisiting the behavior here though.
> 
> cc @tariq1890

Thanks for calling this out @cdesiniotis. You're correct then that this is only a piece of the required functionality.

@just1not2 it may be good to expand #1513 with additional requirements / details on how you would like the operator to behave and also how you are currently configuring it.

Converting this to a draft until we have clarified things.

*Update:* Created https://github.com/NVIDIA/gpu-operator/pull/2041 to add basic support for this to the operator. The thinking is that if a user explicitly sets RUNTIME=none in the toolkit enf, this disables the injection of sockets and config files so that no runtime configuration is performed.

tariq1890:
@elezar I lean towards using a flag to trigger a certain behaviour as opposed to relying on a specific value (RUNTIME=none) to trigger said behaviour. Moreover the notion of `RUNTIME=none` is rather ambiguous and could be interpreted in multiple ways by users.

I don't want to block on this if others are okay with the proposed approach, but my preference right is now if we used a flag like `--no-runtime-config` (as discussed offline) instead

elezar:
> @elezar I lean towards using a flag to trigger a certain behaviour as opposed to relying on a specific value (RUNTIME=none) to trigger said behaviour. Moreover the notion of `RUNTIME=none` is rather ambiguous and could be interpreted in multiple ways by users.
> 
> I don't want to block on this if others are okay with the proposed approach, but my preference right is now if we used a flag like `--no-runtime-config` (as discussed offline) instead

I think a dedicated flag makes more sense too. Abusing the `RUNTIME` envvar was motivated by my (incorrect) assumption that this would not require changes to the GPU Operator. Since these are required anyway, let me rework this.

elezar:
I have also updated https://github.com/NVIDIA/gpu-operator/pull/2041 for this specific implementation. (It's not strictly needed to function correctly, but does mean that the mounts are cleaner).

PR #1530, opened 2025-12-09
Fix library driver library resolution in musl-based containers

### LINKED ISSUE #1526: Fix driver injection in alpine linux containers
Since `musl` (used in Alpine) doesn't use ldconfig or the ldcache, the current mechansim to ensure that the injected libraries are detectable does not work in Alpine containers.

We could update the `update-ldcache` and `enable-cuda-compat` hooks to create / modify `/etc/ld-musl-$(arch -m).path` in addition to creating the `ld.so.conf.d` files. 

See https://wiki.musl-libc.org/faq

To validate a possible fix:
1. Start an alpine container requesting gpus:
    ```
    docker run --rm -ti --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all alpine
    ```
2. Install gcompat since we KNOW that nvidia-smi uses glibc:
    ```
    apk add --no-cache gcompat
    ```
    (without this we get the following message when running nvidia-smi: /bin/sh: nvidia-smi: not found)
3. Create a .path file with the expected contents:
    ```
    cat /etc/ld.so.conf.d/* >> /etc/ld-musl-$(arch -m).path
    ```
4. Run nvidia-smi:
    ```
    nvidia-smi -L
    GPU 0: Orin (nvgpu) (UUID: 1833c8b5-9aa0-5382-b784-68b7e77eb185)
    ```

elezar:
Fixes #1526

### DIFF
--- a/internal/ldconfig/ldconfig.go
+++ b/internal/ldconfig/ldconfig.go
@@ -21,6 +21,7 @@ import (
 	"bufio"
 	"flag"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -168,6 +169,7 @@ func (l *Ldconfig) UpdateLDCache() error {
 		return fmt.Errorf("failed to write %s drop-in: %w", ldsoconfdFilenamePattern, err)
 	}
 
+	systemSearchPaths := l.getSystemSearchPaths()
 	// In most cases, the hook will be executing a host ldconfig that may be configured widely
 	// differently from what the container image expects.
 	// The common case is Debian-like (e.g. Debian, Ubuntu) vs non-Debian-like (e.g. RHEL, Fedora).
@@ -176,8 +178,13 @@ func (l *Ldconfig) UpdateLDCache() error {
 	// paths to a drop-in conf file that is likely to be last in lexicographic order. Entries in the
 	// top-level ld.so.conf file may be processed after this drop-in, but this hook does not modify
 	// the top-level file if it exists.
-	if err := createLdsoconfdFile(defaultLdsoconfdDir, ldsoconfdSystemDirsFilenamePattern, l.getSystemSearchPaths()...); err != nil {
+	if err := createLdsoconfdFile(defaultLdsoconfdDir, ldsoconfdSystemDirsFilenamePattern, systemSearchPaths...); err != nil {
 		return fmt.Errorf("failed to write %s drop-in: %w", ldsoconfdSystemDirsFilenamePattern, err)
+	}
+
+	// Also output the folders to the alpine .path file as required.
+	if err := createMuslPathFileIfRequired(append(filteredDirectories, systemSearchPaths...)...); err != nil {
+		return fmt.Errorf("failed to update .path file for musl: %w", err)
 	}
 
 	return SafeExec(ldconfigPath, args, nil)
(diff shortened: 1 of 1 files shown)


rahulait (commented):
LGTM. I'll let folks who have more context to this repo approve the PR.

internal/ldconfig/ldconfig.go
```diff
@@ -176,10 +178,15 @@ func (l *Ldconfig) UpdateLDCache() error {
...
 	// top-level ld.so.conf file may be processed after this drop-in, but this hook does not modify
 	// the top-level file if it exists.
-	if err := createLdsoconfdFile(defaultLdsoconfdDir, ldsoconfdSystemDirsFilenamePattern, l.getSystemSearchPaths()...); err != nil {
+	if err := createLdsoconfdFile(defaultLdsoconfdDir, ldsoconfdSystemDirsFilenamePattern, systemSearchPaths...); err != nil {
 		return fmt.Errorf("failed to write %s drop-in: %w", ldsoconfdSystemDirsFilenamePattern, err)
 	}
 
+	// Also output the folders to the alpine .path file as required.
+	if err := createAlpinePathFileIfRequired(append(filteredDirectories, systemSearchPaths...)...); err != nil {
+		return err
```
cdesiniotis: Let's add a more descriptive error message here.

tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -589,4 +587,30 @@ EOF`)
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(Equal(hostOutput))
+		})
+
+		It("nvidia-smi should succeed when using the nvidia-container-runtime-hook", Label("legacy"), func(ctx context.Context) {
+			output, _, err := runner.Run(`docker run --rm --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all -e NVIDIA_DRIVER_CAPABILITIES=all \
+				alpinecompat nvidia-smi -L`)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(output).To(Equal(hostOutput))
+		})
```
cdesiniotis, line 614: Question -- was this test case passing before the relevant changes in this PR? If so, what is libnvidia-container doing to handle containers that use musl instead of glibc?

internal/ldconfig/ldconfig.go
```diff
@@ -360,6 +374,38 @@ func processLdsoconfFile(ldsoconfFilename string) ([]string, []string, error) {
...
+		return fmt.Errorf("could not open .path file: %w", err)
+	}
+	defer func() {
+		_ = pathFile.Close()
+	}()
+
+	return outputListToFile(pathFile, dirs...)
+}
+
+func isAlpine() bool {
```
cdesiniotis: Potentially out-of-scope -- is there an easy way of detecting if musl is the standard C library used in a distro? If yes, would we be open to expanding this change to be distro-agnostic? Meaning we would create the `/etc/ld-musl-$ARCH.path` file for any container that uses musl?

A quick LLM query suggested the below:
```
ldd --version 2>&1 | grep -q musl && echo "Using musl" || echo "Using glibc"
```

cdesiniotis (approved):
Thanks @elezar. I have one minor suggestion about logging. I am approving this PR assuming that comment gets addressed. My other two comments are questions that should not be viewed as blockers.

internal/ldconfig/ldconfig.go
elezar: I'd be open to a follow-up to make this more generic. Note, however, that we can't run executables from the container as these should be considered untrusted.

tests/e2e/nvidia-container-toolkit_test.go
elezar: Looking at this, I need to update the test case anyway, but it SHOULD have been passing. Will confirm and drop a link here.

In short, the difference between the `nvidia-container-cli` and the CDI-based implementation is that the former mounts the libraries to the standard library paths in the container. These are dependent on the distribution for the container. This also mean that by defaul the libraries can be discovered even if the paths are not set up (either in the ldcache or in the `.path` file).

elezar: See https://github.com/NVIDIA/nvidia-container-toolkit/actions/runs/20752777985/job/59587159749?pr=1530 which runs the tests before the changes here.

internal/ldconfig/ldconfig.go
elezar: Done.

PR #1535, opened 2025-12-11
trim extraneous whitespaces when parsing comma-separated config sources

tariq1890:
This change makes the input parsing of the `RUNTIME_CONFIG_SOURCES` a bit more forgiving. It is not impossible that users may specify a space when specifying comma separated list of container runtime config sources

### DIFF
--- a/cmd/nvidia-ctk-installer/container/container.go
+++ b/cmd/nvidia-ctk-installer/container/container.go
@@ -199,12 +199,12 @@ func (o Options) GetConfigLoaders(commandSourceFunc func(string, string) toml.Lo
 	var loaders []toml.Loader
 	for _, configSource := range o.ConfigSources {
 		parts := strings.SplitN(configSource, "=", 2)
-		source := parts[0]
+		source := strings.TrimSpace(parts[0])
 		switch source {
 		case "file":
 			fileSourcePath := o.TopLevelConfigPath
 			if len(parts) > 1 {
-				fileSourcePath = parts[1]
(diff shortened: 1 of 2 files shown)


cmd/nvidia-ctk-installer/container/container.go
elezar: Question: Should we also trim the argument if supplied?

tariq1890: You mean trim the arg value before we split the string into a slice ?

tariq1890: If yes, I am not sure if that makes much of a difference as those spaces would be trimmed after we split the slice and loop through the elements anyway

elezar: What if the flag has the value: `"file= foo.txt"`? Here there is a leading space in the file path.

elezar: I mean:
```suggestion
				fileSourcePath = strings.TrimSpace(parts[1])
```

tariq1890: Oh ok, I get it now. Yes, I think we can trim those spaces out too

tariq1890: Sure, I can make this change too

tariq1890: This is done, please check now

tariq1890: I've also added a unit test to cover this change

PR #1541, opened 2025-12-12
feat: add sth close to the no-cgroups legacy feature in CDI mode [WIP/Draft]

oOraph:
Related: https://github.com/NVIDIA/nvidia-container-toolkit/issues/1542

Use case: same as the no-cgroups flags in legacy mode (= with the hook) -> create dev nodes, setup environment, mount shared libraries but do not give access to devices yet

### DIFF
--- a/internal/config/runtime.go
+++ b/internal/config/runtime.go
@@ -27,6 +27,8 @@ type RuntimeConfig struct {
 	Runtimes []string    `toml:"runtimes"`
 	Mode     string      `toml:"mode"`
 	Modes    modesConfig `toml:"modes"`
+	// Close to the "no-cgroups" bool when using the legacy mode with the hooks.
+	MknodOnly bool `toml:"mknod-only"`
 }
 
 // modesConfig defines (optional) per-mode configs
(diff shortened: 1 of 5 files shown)


elezar:
@oOraph thinking about this again. I don't know whether we want to expose this through the `nvidia-container-runtime`? Would providing options to generate a CDI spec with the "correct" options be sufficient for your use case?

Something like:
```
nvidia-ctk cdi generate --force-device-node-permissions=m
```?

oOraph:
> @oOraph thinking about this again. I don't know whether we want to expose this through the `nvidia-container-runtime`? Would providing options to generate a CDI spec with the "correct" options be sufficient for your use case?
> 
> Something like:
> 
> ```
> nvidia-ctk cdi generate --force-device-node-permissions=m
> ```
> ?

@elezar I guess so. Actually I'm not 100% fluent with the nvidia-container-runtime modes :) you will probably help me here: if I use the auto mode it won't use the cdi generated specs by the nvidia-ctk will it ? is this the difference between the cdi and cdi-jit modes ? Would I then need to use this command and switch explicitely the runtime to cdi mode for it to use the cdi specs generated outside of the runtime ? If so this would be enough for my case

elezar:
@oOraph sorry for the delay in looking into this. I think that the work in https://github.com/NVIDIA/nvidia-container-toolkit/pull/630 would be useful here. I think it makes more sense to add this as an option to the `edits.Factory` (https://github.com/NVIDIA/nvidia-container-toolkit/pull/630/changes#diff-1cac692fac117aeb25bc481c81413a7b9f4f6ab44640ad3ca96aa4bdda9092fdR51) and handle this there.

Let me revisit this once we've merged that change.

PR #1551, opened 2025-12-19
Add debug-level log option for ctk-installer

jactor-sue:
Enable debug-level log for nvidia-ctk-installer cmd by `--debug` or `-d`.

### DIFF
--- a/cmd/nvidia-ctk-installer/main.go
+++ b/cmd/nvidia-ctk-installer/main.go
@@ -33,10 +33,11 @@ const (
 	defaultNRIPluginIdx uint = 10
 )
 
-var defaultLowLevelRuntimes = []string{"runc", "crun"}
-
-var waitingForSignal = make(chan bool, 1)
-var signalReceived = make(chan bool, 1)
+var (
+	defaultLowLevelRuntimes = []string{"runc", "crun"}
+	waitingForSignal        = make(chan bool, 1)
+	signalReceived          = make(chan bool, 1)
+)
 
 // options stores the command line arguments
 type options struct {


cmd/nvidia-ctk-installer/main.go
```diff
@@ -28,11 +29,15 @@ const (
 	defaultRuntime = "docker"
 )
 
-var availableRuntimes = map[string]struct{}{"docker": {}, "crio": {}, "containerd": {}}
-var defaultLowLevelRuntimes = []string{"runc", "crun"}
+var (
```
elezar: If you're going to make these changes please make them in a separate commit.

cmd/nvidia-ctk-installer/main.go
```diff
@@ -143,6 +150,14 @@ func (a app) build() *cli.Command {
...
 				Sources:     cli.EnvVars("TOOLKIT_PID_FILE", "PID_FILE"),
 			},
+			&cli.BoolFlag{
+				Name:        "debug",
+				Aliases:     []string{"d"},
+				Value:       false,
+				Usage:       "enable debug-level log",
+				Destination: &options.debug,
+				Sources:     cli.EnvVars("TOOLKIT_DEBUG"),
+			},
```
elezar, line 209: One thing that we have been considering is switching to an explicit verbosity-based approach for this component so as to prepare for a switch to `klog` at some point. Could we changes this to:

```suggestion
			&cli.IntFlag{
				Name:        "log-versbosity",
				Aliases:     []string{"v"},
				Destination: &options.logVerbosity,
				Sources:     cli.EnvVars("LOG_VERBOSITY"),
			},
```

cmd/nvidia-ctk-installer/main.go
jactor-sue: Nothing changed here, auto-formated by the [gofumpt](https://github.com/mvdan/gofumpt) formatter.

cmd/nvidia-ctk-installer/main.go
jactor-sue: So you want to change logger from `logrus` to `klog` for whole toolkit stack or only for installer?

elezar: The idea would be to move to something like `logr.Logger` internally, and then use `klog` in the installer and `logrus` for the CLI components. This may be longer term though, so a simpler change may be to ensure that the installer can have debug logging enabled as per your original change.

Deprecating the `--debug` flag, or mapping it to a well-defined `log-verbosity` at some point in the future would also be possible.

elezar: (feel free to rebase and drop my commit so that I can review this).

elezar: I've updated the PR -- removing my proposed commit.

PR #1553, opened 2025-12-27
fix: add device rules for AllowDevice.conf when running in legacy mode

jokestax:
## Description

This PR fixes an issue where containers using the NVIDIA runtime in **legacy mode** fail after a `systemctl daemon-reload` is performed on the host.

The root cause is that the `devices.allow` cgroup file is not being updated with NVIDIA character device rules when containers are created in legacy mode. This causes `nvidia-smi` and other tools to fail with:
```
Failed to initialize NVML: Unknown Error
```

### Changes

- Added cgroup device rules for NVIDIA devices in the legacy mode modifier
- Rules are now added based on `NVIDIA_VISIBLE_DEVICES` environment variable:
  - Common devices: `nvidiactl`, `nvidia-uvm`, `nvidia-uvm-tools`, `nvidia-modeset`
  - GPU-specific devices: `/dev/nvidia0`, `/dev/nvidia1`, etc. (based on requested GPUs)
- Supports `NVIDIA_VISIBLE_DEVICES=all`, specific indices (`0,1,2`), and UUIDs

## Related Issue(s)

Fixes https://github.com/opencontainers/runc/issues/4859

### Reproducing the Issue (Before Fix)

1. Install the latest `nvidia-container-toolkit` version

2. Update `/etc/nvidia-container-runtime/config.toml` to use legacy mode:
```toml
   [nvidia-container-runtime]
   #debug = "/var/log/nvidia-container-runtime.log"
   log-level = "info"
   mode = "legacy"
   runtimes = ["runc", "crun"]
```

3. Restart Docker:
```bash
   sudo systemctl restart docker
```

4. Start a GPU container:
```bash
   docker run -d --runtime=nvidia \
     -e NVIDIA_VISIBLE_DEVICES=0 or --gpus=all \
     --name gpu-test \
     nvidia/cuda:12.2.0-base-ubuntu22.04 sleep 30000
```

5. Verify `nvidia-smi` works:
```bash
   docker exec -it gpu-test nvidia-smi
   # Should work ✓
```

6. In a **second terminal**, run:
```bash
   sudo systemctl daemon-reload
```

7. Test `nvidia-smi` again in the container:
```bash
   docker exec -it gpu-test nvidia-smi
   # ERROR: Failed to initialize NVML: Unknown Error ✗
```

### Testing the Fix

1. Clone this repository and build the new binary:
```bash
   git clone <repository-url>
   cd nvidia-container-toolkit
   go build -o nvidia-container-runtime ./cmd/nvidia-container-runtime
```

2. Backup the original binary:
```bash
   sudo cp /usr/bin/nvidia-container-runtime /usr/bin/nvidia-container-runtime.bak
```

3. Install the new binary:
```bash
   sudo cp nvidia-container-runtime /usr/bin/nvidia-container-runtime
```

4. Restart Docker:
```bash
   sudo systemctl restart docker
```

5. Start a new GPU container:
```bash
   docker run -d --runtime=nvidia \
     ---e NVIDIA_VISIBLE_DEVICES=0 or --gpus=all \
     --name gpu-test-fixed \
     nvidia/cuda:12.2.0-base-ubuntu22.04 sleep 30000
```

6. Verify `nvidia-smi` works:
```bash
   docker exec -it gpu-test-fixed nvidia-smi
   # Should work ✓
```

7. Run `daemon-reload` in another terminal:
```bash
   sudo systemctl daemon-reload
```

8. Test `nvidia-smi` again:
```bash
   docker exec -it gpu-test-fixed nvidia-smi
   # Should still work ✓
```

### Rollback (If Needed)
```bash
sudo cp /usr/bin/nvidia-container-runtime.bak /usr/bin/nvidia-container-runtime
sudo systemctl restart docker
```

### DIFF
--- a/internal/modifier/stable.go
+++ b/internal/modifier/stable.go
@@ -62,26 +62,30 @@ type stableRuntimeModifier struct {
 // as a prestart hook.
 func (m stableRuntimeModifier) Modify(spec *specs.Spec) error {
 	// If an NVIDIA Container Runtime Hook already exists, we don't make any modifications to the spec.
+	hookExists := false
 	if spec.Hooks != nil {
 		for _, hook := range spec.Hooks.Prestart {
 			hook := hook
 			if isNVIDIAContainerRuntimeHook(&hook) {
 				m.logger.Infof("Existing nvidia prestart hook (%v) found in OCI spec", hook.Path)
-				return nil
+				hookExists = true
+				break
 			}
 		}
 	}
 
-	path := m.nvidiaContainerRuntimeHookPath
-	m.logger.Infof("Using prestart hook path: %v", path)
-	args := []string{filepath.Base(path)}
-	if spec.Hooks == nil {
-		spec.Hooks = &specs.Hooks{}
+	if !hookExists {
+		path := m.nvidiaContainerRuntimeHookPath
+		m.logger.Infof("Using prestart hook path: %v", path)
+		args := []string{filepath.Base(path)}
+		if spec.Hooks == nil {
+			spec.Hooks = &specs.Hooks{}
+		}
+		spec.Hooks.Prestart = append(spec.Hooks.Prestart, specs.Hook{
+			Path: path,
+			Args: append(args, "prestart"),
+		})
 	}
-	spec.Hooks.Prestart = append(spec.Hooks.Prestart, specs.Hook{
-		Path: path,
-		Args: append(args, "prestart"),
-	})
 
 	if err := m.AddDeviceCgroupRules(spec); err != nil {
 		return err


elezar:
@jokestax could you comment on the use cases where you are expecting to use `legacy` mode and the new default `jit-cdi` mode is not sufficient?

jokestax:
Hi @elezar,in older versions of containerd where cdi is not supported like versions <1.6 ig

elezar:
When container does not support CDI, the `nvidia` runtime can be used. This does not depend on the mode in which the `nvidia-container-runtime` is configured. This means that even when running in the default mode (i.e. `jit-cdi` as of v1.18.0), the `nvidia` runtime can be used to perform device injection.

Note that the `cdi` mode is generally used when some other component is responsible for generating CDI specifications for requested devices.

If you are having problems running containerd <= 1.6 with the `nvidia` runtime configured in the default `jit-cdi` mode, this should be reported separately.

jokestax:
Thanks for the clarification @elezar I've confirmed the issue doesn't occur in jit-cdi mode.
what would you recommend as the next step here — should I close this PR, or is there anything else needed?

elezar:
> Thanks for the clarification @elezar I've confirmed the issue doesn't occur in jit-cdi mode. what would you recommend as the next step here — should I close this PR, or is there anything else needed?

Thanks for confirming that things work with the new default mode. The complexity that this adds when running in `legacy` mode is not something that we're looking to support at the moment -- especially since we are actively trying to move away from legacy mode. There are some use cases (e.g. ONLY specifying the Docker `--gpus` flag) that still invoke the `nvidia-container-runtime-hook` through other mechanisms that still need to be migrated.

With this in mind, I would recommend closing the PR.

jokestax:
Thanks so much @elezar for taking the time to review this,I'll go ahead and close this PR,If there are any open issues where I could contribute, I'd be happy to help out. Thanks again!

PR #1561, opened 2026-01-07
fix: Include /dev/nvidia2 in container requesting iGPU if required

elezar:
On Tegra-based systems where both an iGPU (e.g. Thor) and a dGPU are present, the rendering devices for the iGPU are associated with `/dev/nvidia2`. This means that this device node should be included if present.

Depends on #1556


pkg/nvcdi/lib-csv.go
```diff
@@ -108,9 +105,9 @@ func (l *csvlib) mixedDeviceSpecGenerators(ids ...string) (DeviceSpecGenerator,
 // platform-specific CSV files.
 type csvDeviceGenerator struct {
 	*csvlib
-	index               int
-	uuid                string
-	noFilterDeviceNodes bool
+	index int
+	uuid  string
+	mode  string
```
ArangoGutierrez: Should we consider 

```go
type gpuMode string

const (
    gpuModeDGPU gpuMode = "dgpu"
    gpuModeIGPU gpuMode = "igpu"
)
```

internal/platform-support/tegra/filter.go
```diff
@@ -125,6 +128,14 @@ func (p filterByMountSpecType) Apply(input MountSpecPathsByTyper) MountSpecPaths
 	return ms
 }
 
+func (p filterByMountSpecPathsByTyper) Apply(input MountSpecPathsByTyper) MountSpecPathsByTyper {
+	f := make(filterByMountSpecType)
+	for t, p := range p.MountSpecPathsByType() {
+		f[t] = &matcherAsFilter{pathPatterns(p)}
```
ArangoGutierrez: readability NIT
```suggestion
	for t, paths := range p.MountSpecPathsByType() {
		f[t] = &matcherAsFilter{pathPatterns(paths)}
```

elezar: Updated.

elezar: Updated.

pkg/nvcdi/lib-csv.go
elezar: Done.

PR #1581, opened 2026-01-15
elezar comments for NRI plugin server

elezar:
Here is a proposal for a cleanup of #1498 

It includes some of the intial commits from #1521 and builds on that to ensure that the runtime is not configured if NRI is enabled. Note that the config options are moved to top-level flags since we don't propagate the NRI config to the runtime.

### DIFF
--- a/cmd/nvidia-ctk-installer/main.go
+++ b/cmd/nvidia-ctk-installer/main.go
@@ -9,6 +9,7 @@ import (
 	"syscall"
 	"time"
 
+	nriapi "github.com/containerd/nri/pkg/api"
 	"github.com/urfave/cli/v3"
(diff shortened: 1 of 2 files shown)


cmd/nvidia-ctk-installer/main.go
```diff
@@ -143,6 +151,33 @@ func (a app) build() *cli.Command {
 				Destination: &options.pidFile,
 				Sources:     cli.EnvVars("TOOLKIT_PID_FILE", "PID_FILE"),
 			},
+			&cli.BoolFlag{
+				Name:        "enable-nri-in-runtime",
```
tariq1890, line 155: I prefer `--enable-nri-plugin` as it's more explicit in what it's trying to do

PR #1597, opened 2026-01-21
[no-relnote] Set golang version when building libnvidia-container

elezar:
This change ensures that we use the same golang version to build the NVIDIA Container Toolkit packages and the packages in libnvidia-container.

This requires https://github.com/NVIDIA/libnvidia-container/pull/334 and also bumps `libnvidia-container` to https://github.com/NVIDIA/libnvidia-container/commit/a83ddc087e255fa88612f7d6afa43acdd71ddd96


elezar:
We may want to consider backporting this and https://github.com/NVIDIA/libnvidia-container/pull/334

PR #1598, opened 2026-01-21
Query CUDA forward compatibility elf note if available

elezar:
This change queries an ELF note section in libcuda in the container to determine whether the forward compat libraries in the container should be used over the host drivers.

If the elf note section is not available, we fall back to the heuristics using the major numbers of the host and compat driver libraries.

### DIFF
--- a/cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
+++ b/cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
@@ -169,15 +169,21 @@ func (m command) getContainerForwardCompatDir(containerRoot containerRoot, o *op
 }
 
 func (m command) useCompatLibraries(libcudaCompatPath string, hostDriverVersion string) (bool, error) {
+	driverMajor, err := extractMajorVersion(hostDriverVersion)
+	if err != nil {
+		return false, fmt.Errorf("failed to extract major version from %q: %v", hostDriverVersion, err)
+	}
+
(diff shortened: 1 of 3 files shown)


cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -154,24 +154,44 @@ func (m command) getContainerForwardCompatDir(containerRoot containerRoot, cudaC
...
 	if err != nil {
-		return "", fmt.Errorf("failed to extract major version from %q: %v", hostDriverVersion, err)
+		return false, fmt.Errorf("failed to extract major version from %q: %v", hostDriverVersion, err)
 	}
 
-	if driverMajor >= compatMajor {
-		m.logger.Debugf("Compat major version is not greater than the host driver major version (%v >= %v)", hostDriverVersion, compatDriverVersion)
-		return "", nil
+	// First check the elf header.
+	cudaCompatHeader, _ := GetCUDACompatElfHeader(libcudaCompatPath)
```
ArangoGutierrez, line 178: Nit: Could we at least log the error?

elezar: Firstly, this is in a hook and the logs are not captured and as such logs are not that helpful. Secondly, I want to explicitly ignore errors at this stage since I want to fall back to the heuristics that we already have in place.

PR #1614, opened 2026-01-27
Allow Orin CUDA forward compat root to be specified

elezar:
This change allows the CUDA forward compat root used for Orin-based systems to be specified as a config option or as a flag to the `nvidia-ctk cdi generate` command.

### DIFF
--- a/cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
+++ b/cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
@@ -25,11 +25,14 @@ import (
 	"fmt"
 	"os"
 	"slices"
+	"strings"
+
+	"golang.org/x/mod/semver"
 )
 
 type compatElfHeader struct {
 	Format      int
-	CUDAVersion string `json:"CUDA Version"`
+	CUDAVersion cudaVersion `json:"CUDA Version"`
 	Driver      []int
 	Device      []int
 }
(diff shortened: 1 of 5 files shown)


pkg/nvcdi/lib-csv.go
```diff
@@ -52,6 +57,9 @@ func (l *nvcdilib) asCSVLib() *csvlib {
 	if len(l.csv.Files) == 0 {
 		l.csv.Files = csv.DefaultFileList()
 	}
+	if l.csv.CompatContainerRoot == "" {
+		l.csv.CompatContainerRoot = defaultOrinCompatContainerRoot
```
cdesiniotis, line 61: Question -- will the compat folder path change on future architectures, like Thor? How do you envision us handling that?

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
```diff
@@ -114,10 +117,35 @@ func getCUDAFwdCompatibilitySection(lib *elf.File) *elf.Section {
...
+// driver verions, this is checked, otherwise the CUDA version specified in the
+// ELF section is checked.
+func (h *compatElfHeader) UseCompat(hostDriverMajor int, hostCUDAVersion string) bool {
 	if h == nil {
 		return false
 	}
 
-	return slices.Contains(h.Driver, driverMajor)
+	if hostDriverMajor != 0 && len(h.Driver) > 0 {
+		return slices.Contains(h.Driver, hostDriverMajor)
```
cdesiniotis, line 131: Question -- I thought we only use the compat libs when the major version is greater than the host driver's major version. Am I missing something here? As currently implemented it appears as if `UseCompat()` returns true if the major versions are equal.

pkg/nvcdi/lib-csv.go
elezar: With Thor systems (and beyond) the same compat libraries that are used for dGPU-based systems are used. Thus the only edge case we have to deal with is Orin system that are currently under support and require different compat packages.

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
elezar: That was the initial basic check, yes. This change extends the basic heuristics to actually check the versions specified in the ELF header of the `libcuda.so` included in the compat folder. The header includes a list of driver major versions that the library can provide forward compatibility for. That is what this check tries to implement.

In the case of Orin-based systems there is no equivalent for the host driver version, and the CUDA version is compared.

cdesiniotis: Does this mean that our `enable-cuda-compat` hook now supports both CUDA forward compatibility and CUDA minor version (aka enhanced) compatibility?

cdesiniotis: I am assuming this does allow us to support CUDA enhanced compatibility. Below is an example exemplifying this:

```
Host driver: 590.44.01 / CUDA 13.1.0
CUDA compat in container: 590.48.01 / CUDA 13.1.1
```

Assuming that `590` is present in the list of driver major versions in the ELF header of the `libcuda.so` in the container, the compat libraries would be used and this exercises CUDA enhanced compatibility (right?).

Taking another example, what is the expected / current behavior for the below?

```
Host driver: 590.48.01 (CUDA version: 13.1.1)
CUDA compat version in container: 590.44.01 / CUDA 13.1.0
```

Should the host driver libraries or the compat libraries be used? Does it even matter in this example as both the CUDA major and minor versions are the same? IIUC our current `enable-cuda-compat` hook would decide that the compat libraries should be used (for the same reasons as the first example).

elezar: I have found a t least one example where the logic as implemented here does not work as expected (see https://github.com/NVIDIA/nvidia-container-toolkit/pull/1697). For that reason (and as discussed offline), I think it best to revert this until we have a better defintion of the expectations.

PR #1615, opened 2026-01-27
test: Adjust e2e test for Docker 29.2.0

elezar:
As of Docker 29.2.0, the `--gpus=all` flag is converted to a `--device=nvidia.com/gpu=all` request. This means that tests that expect DIFFERENT behaviour when using the `--gpus=all` flag and other flags are not applicable.

This change updates the firmware injection test for the new behaviour.

### DIFF
--- a/tests/e2e/nvidia-container-toolkit_test.go
+++ b/tests/e2e/nvidia-container-toolkit_test.go
@@ -24,16 +24,26 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+
+	"golang.org/x/mod/semver"
 )
 
 // Integration tests for Docker runtime
 var _ = Describe("docker", Ordered, ContinueOnFailure, func() {
+	var dockerVersion string
 	var hostDriverVersion string
 	var hostDriverMajor string
 	var hostOutput string
 
 	// Install the NVIDIA Container Toolkit
 	BeforeAll(func(ctx context.Context) {
+		// Get the docker version output in the form:
+		// Docker version 29.1.3, build f52814d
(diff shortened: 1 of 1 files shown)


tests/e2e/nvidia-container-toolkit_test.go
```diff
@@ -430,14 +430,16 @@ EOF`)
 		})
 
 		It("should not fail when using the nvidia-container-runtime", func(ctx context.Context) {
-			_, _, err := runner.Run("docker run --rm --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all -e NVIDIA_DRIVER_CAPABILITIES=all firmware-test")
+			output, _, err := runner.Run("docker run --rm --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all -e NVIDIA_DRIVER_CAPABILITIES=all firmware-test")
```
jgehrcke, line 443: If I read correctly, this might be silently dropping error detail that may matter at some point to someone :-) We drop `stderr` of the child process here, right?

"Silently dropping or even changing error detail that _may_ matter at some point to someone" is I think chapter one, sin 1 in my imaginary book about error handling :)

But you probably know better here!

elezar: Yes, we do drop `stderr` here. I understand that we should be more dilegent about checking these, but that's a bigger change than I want o make in this PR. I would like to revisit our testing strategy and organisation for this repo in general in the near future.

(Let me at least update this PR to explicitly check `stderr`).

elezar: For the time being I updated the tests to consisder the docker version on the host and check the applicable conditions based on that.

PR #1621, opened 2026-01-30
add Talos glibc library search paths

hydazz:
This is a required change to get the k8s-dra-driver to work on Talos (or other OS's that use `/usr/local/glibc/`)

as discussed https://github.com/NVIDIA/k8s-dra-driver-gpu/pull/695

### DIFF
--- a/internal/lookup/library.go
+++ b/internal/lookup/library.go
@@ -32,6 +32,9 @@ func NewLibraryLocator(opts ...Option) Locator {
 	opts = append(opts,
 		WithSearchPaths([]string{
 			"/",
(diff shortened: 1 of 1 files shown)


elezar (commented):
My one question would be whether we want to add the paths to the end of the list instead of at the front.

Also, in https://github.com/NVIDIA/k8s-dra-driver-gpu/pull/695 it was also required to update the path when searching for `nvidia-smi`. Is that change also applicable here?

hydazz:
fixed on the talos side, natively works as is on 1.13.0+
This PR is no longer needed

PR #1637, opened 2026-02-06
Refactor device node tests

elezar:
In order to make changes related to device nodes more testable, these changes refactor the way they are discovered in tests to allow for tests to specify mocks used in discovery.

We also replace the logic to check the `__NVCT_TESTING_DEVICES_ARE_FILES` envvar when discovering device nodes on our test file systems with stubs that can be overridden per test.

This was motivated by #1631  and #630.


jgehrcke (approved):
> to allow for tests to specify mocks used in discovery.

Nice! LGTM

PR #1645, opened 2026-02-10
refactor: Add AsOptional wrapper for lookup.Locator

elezar:
This change moves from a field based check for an optional locator to using a wrapper. This enables simpler composition and makes it simpler to make new locators explicitly optional.

This includes the changes from #1648 which should be reviewed first.

### DIFF
--- a/pkg/lookup/ldcache.go
+++ b/pkg/lookup/ldcache.go
@@ -35,13 +35,9 @@ type ldcacheLocator struct {
 
 var _ Locator = (*ldcacheLocator)(nil)
 
-// NewLdcacheLocator creates a locator that allows libraries to be found using
+// newLdcacheLocator creates a locator that allows libraries to be found using
 // the ldcache.
-func NewLdcacheLocator(opts ...Option) Locator {
-	return NewFactory(opts...).NewLdcacheLocator()
-}
-
-func (f *Factory) NewLdcacheLocator() Locator {
+func (f *Factory) newLdcacheLocator() Locator {
 	cache, err := ldcache.New(f.logger, f.root)
 	if err != nil {
 		f.logger.Warningf("Failed to load ldcache: %v", err)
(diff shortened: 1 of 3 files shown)


pkg/lookup/file.go
```diff
@@ -27,6 +27,8 @@ import (
 // file can be used to locate file (or file-like elements) at a specified set of
 // prefixes. The validity of a file is determined by a filter function.
 type file struct {
+	// logger logger.Interface
+	// filter func(string) error
```
cdesiniotis: Was this change included by accident?

pkg/lookup/ldcache.go
```diff
@@ -35,13 +35,9 @@ type ldcacheLocator struct {
...
 
-// NewLdcacheLocator creates a locator that allows libraries to be found using
+// newLdcacheLocator creates a locator that allows libraries to be found using
 // the ldcache.
-func NewLdcacheLocator(opts ...Option) Locator {
-	return NewFactory(opts...).NewLdcacheLocator()
-}
-
-func (f *Factory) NewLdcacheLocator() Locator {
+func (f *Factory) newLdcacheLocator() Locator {
```
cdesiniotis, line 40: Question -- I am assuming this was done since none of our code today (outside of this package) use NewLdcacheLocator? I am assuming they use `NewLibraryLocator` instead?

pkg/lookup/file.go
cdesiniotis: I see this was uncommented in a future commit.

pkg/lookup/ldcache.go
elezar: Yes. This was an attempt to reduce the API. If I recall correctly, we used to instantate the ldcache locator directly, but then encapsulated it in a library locator later, but never changed the visibility of the function.

pkg/lookup/file.go
elezar: Yes. I rebased and reworked this a couple of times. Thanks for picking it up though.

PR #1646, opened 2026-02-10
Add dlopen library locator

elezar:
This attepts to add a DLOpen locator as required by nix packages: https://github.com/NixOS/nixpkgs/blob/1c415018b24020442ee94e6b84a8a4a24078a61e/pkgs/by-name/nv/nvidia-container-toolkit/0001-Add-dlopen-discoverer.patch#L50

Also requires https://github.com/NVIDIA/go-nvml/pull/171


pkg/lookup/library.go
```diff
@@ -17,6 +17,13 @@
 package lookup
 
 // NewLibraryLocator creates a library locator using the specified options.
+// If search paths (WithSearchPaths(path1, path2, ...)) are explicitly specified
+// a library locator using these as absolute paths are used. Otherwise the the
```
cdesiniotis: nit: 
```suggestion
// a library locator using these as absolute paths are used. Otherwise the
```

pkg/lookup/dlopen.go
```diff
@@ -0,0 +1,75 @@
...
+
+// dlopenLocator can be used to locate libraries given a system's dynamic
+// linker.
+type dlopenLocator struct {
+	logger logger.Interface
+}
+
+// NewDlopenLocator creats a locator that can be used for locating libraries
+// through the dlopen mechanism.
+func (f *Factory) NewDlopenLocator() Locator {
```
cdesiniotis: Question -- do we want to make this method private initially? We recently made the constructor for the ldcache locator private in https://github.com/NVIDIA/nvidia-container-toolkit/commit/8e35c22864884ec4809834fe938c49cf47311084

pkg/lookup/dlopen.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+)
+
+// dlopenLocator can be used to locate libraries given a system's dynamic
+// linker.
+type dlopenLocator struct {
+	logger logger.Interface
+}
+
+// NewDlopenLocator creats a locator that can be used for locating libraries
```
cdesiniotis: nit:
```suggestion
// NewDlopenLocator creates a locator that can be used for locating libraries
```

pkg/lookup/dlopen.go
elezar: Yes, that's a good point. The initial implementation of this predates the refactor (It was one of the motivators for the refactor). I didn't make it private after the rebase.

cdesiniotis (approved):

PR #1648, opened 2026-02-13
Refactor ldcache locator

elezar:
This change simplifies the lookup of libraries in the Ldcache. Instead of storing a list of cache entries to the libraries they resolve to the cache entries and any symlinks in the chain to the final libraries are stored. This simplifies the lookup of libraries to a list traversal.

### DIFF
--- a/pkg/lookup/ldcache.go
+++ b/pkg/lookup/ldcache.go
@@ -19,18 +19,25 @@ package lookup
 import (
 	"fmt"
 	"path/filepath"
-	"slices"
+	"strings"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/ldcache"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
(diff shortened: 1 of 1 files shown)


pkg/lookup/ldcache.go
```diff
@@ -43,76 +50,70 @@ func NewLdcacheLocator(opts ...Option) Locator {
...
+	var libraries []string
 	_, libs64 := cache.List()
 	for _, library := range libs64 {
-		if _, processed := resolvesTo[library]; processed {
-			continue
-		}
-		candidates, err := chain.Locate(library)
+		chain, err := symlinks.ResolveChain(library)
 		if err != nil {
-			b.logger.Errorf("error processing library %s from ldcache: %v", library, err)
```
cdesiniotis, line 56: nit: may be valuable to add a warning message here:
```
b.logger.Warningf("Failed to resolve symlink chain for library %q: %v", library, err)
```

PR #1652, opened 2026-02-16
Add dev root to driver

elezar:
These changes add a `DevRoot` member to the driver type. This means that the `driverRoot` (for libraries and executables) and the `devRoot` need not be tracked separately.

### DIFF
--- a/internal/platform-support/dgpu/dgpu.go
+++ b/internal/platform-support/dgpu/dgpu.go
@@ -18,6 +18,7 @@ package dgpu
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/NVIDIA/go-nvlib/pkg/nvlib/device"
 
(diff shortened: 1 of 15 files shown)


internal/discover/gdrcopy.go
```diff
@@ -16,12 +16,15 @@
...
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/lookup/root"
+)
 
-func NewGDRCopyDiscoverer(logger logger.Interface, devRoot string) (Discover, error) {
+func NewGDRCopyDiscoverer(logger logger.Interface, driver *root.Driver) (Discover, error) {
 	return NewCharDeviceDiscoverer(
 		logger,
-		devRoot,
+		driver.Root,
```
cdesiniotis: Did you mean `driver.DevRoot`?

elezar: Yes. Good catch. Let me add a test too.

elezar: I've added tests for some fo the basic discoverers that at least caught this bug.

PR #1655, opened 2026-02-17
Add a modifier factory

elezar:
This change refactors modifier construction. It adds a `Factory` type which is used to encapsulate modifier construction. This means that it is not required to pass the same arguments (e.g. `logger`, `driver`, etc.) to each function for creating a specific modifier.

### DIFF
--- a/internal/modifier/csv_test.go
+++ b/internal/modifier/csv_test.go
@@ -58,7 +58,7 @@ func TestNewCSVModifier(t *testing.T) {
 			image, _ := image.New(
 				image.WithEnvMap(tc.envmap),
 			)
-			f := NewFactory(
+			f := createFactory(
 				WithLogger(logger),
 				WithConfig(tc.cfg),
 				WithImage(&image),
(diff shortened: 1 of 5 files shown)


jgehrcke (approved):
I scrolled through the changes, and then I saw

Image

Is that new? Approved!

(the CI failure is certainly something you know how to interpret...)

elezar:
> Is that new? Approved!

Relatively, yes.

> (the CI failure is certainly something you know how to interpret...)

The issue is that we're installing the `rpmrebuild` package from the `epel` repos and as of this morning there seems to be a glibc mismatch. I've added a commit on top to switch to a `fedora:43` image to rebuild the packages to at least unblock the CI. (Created https://github.com/NVIDIA/nvidia-container-toolkit/pull/1657 to track the CI fix separately).

jgehrcke:
> and as of this morning there seems to be a glibc mismatch

I see. Complex, real-world CI has so many moving targets.

elezar:
> > and as of this morning there seems to be a glibc mismatch
> 
> I see. Complex, real-world CI has so many moving targets.

Looking into this a bit more deeply, I think it's related to the package posting for https://access.redhat.com/errata/RHSA-2026:2786 it seems as if the `.i686` package is available on the repos but the `.x86_64` packages are not.

PR #1667, opened 2026-02-19
Use automatic CDI spec generation to generate CDI specs for other modifiers

elezar:
This change ensures that we reuse the instatiation of the nvcdi library to generate CDI specs for CSV mode, feature gated modes mode, as well as jit-cdi mode. This reduces the amount of code that needs to be kept in sync when making modifications.

### DIFF
--- a/internal/modifier/gated.go
+++ b/internal/modifier/gated.go
@@ -21,8 +21,6 @@ import (
 
 	"github.com/NVIDIA/nvidia-container-toolkit/api/config/v1"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/discover"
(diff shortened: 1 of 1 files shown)


internal/modifier/factory.go
```diff
@@ -71,6 +70,9 @@ func createFactory(opts ...Option) *Factory {
 }
 
 func (f *Factory) validate() error {
+	if f.driver == nil {
+		return fmt.Errorf("a driver muse be specified")
```
cdesiniotis:
```suggestion
		return fmt.Errorf("a driver must be specified")
```

PR #1675, opened 2026-02-23
stop marking PRs as stale

rahulait:
We don't want to close PRs due to inactivity. We would like to disable/remove labeling PRs as stale.


Sarangsnambiar:
awesome

PR #1690, opened 2026-03-03
fix: Ensure that CUDA Compat Container path is set by default

elezar:
This change ensures that the CUDA Compat Container path is set on orin-based systems. Without this, the CUDA Compat libraries for SBSA are (incorrectly) used in a container.

### DIFF
--- a/pkg/nvcdi/options.go
+++ b/pkg/nvcdi/options.go
@@ -69,9 +69,6 @@ func populateOptions(opts ...Option) *options {
 		mode:              ModeAuto,
 		driverRoot:        "/",
 		nvidiaCDIHookPath: "/usr/bin/nvidia-cdi-hook",
-		csv: csvOptions{
-			CompatContainerRoot: defaultOrinCompatContainerRoot,
-		},
 	}
 	for _, opt := range opts {
 		opt(o)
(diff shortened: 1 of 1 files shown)


pkg/nvcdi/options.go
```diff
@@ -69,9 +69,6 @@ func populateOptions(opts ...Option) *options {
 		mode:              ModeAuto,
 		driverRoot:        "/",
 		nvidiaCDIHookPath: "/usr/bin/nvidia-cdi-hook",
-		csv: csvOptions{
-			CompatContainerRoot: defaultOrinCompatContainerRoot,
-		},
```
jgehrcke, line 74: by removing this, `o.csv` is now initialized as `nil`, right?

jgehrcke: or maybe not because of this type definition (not a pointer):
```
csv csvOptions
```

pkg/nvcdi/options.go
```diff
@@ -109,6 +106,10 @@ func populateOptions(opts ...Option) *options {
 		o.csv.Files = csv.DefaultFileList()
 	}
 
+	if o.mode == ModeCSV && o.csv.CompatContainerRoot == "" {
+		o.csv.CompatContainerRoot = defaultOrinCompatContainerRoot
```
jgehrcke, line 110: so `o.csv` is never a nil pointer dereference, yes?

pkg/nvcdi/options.go
elezar: Yes, it's just an embeded struct, it is not a pointer. This means that the default value for the `o.csv.CompatContainerRoot` field is now `""`.

pkg/nvcdi/options.go
elezar: Correct. The `csv` member is NOT a pointer type. It is an embeded struct.

PR #1694, opened 2026-03-04
fix: Prevent config option from reverting modified configurations

sunjq1:
## Summary
This PR fixes an issue where uncommented default configuration options (such as `nvidia-container-runtime.debug`) were being automatically re-commented when running `nvidia-ctk config`(without the `--set` flag), even if the user had explicitly enabled them in the configuration file.

## Problem
The current implementation of `commentDefaults()` relies on a `valuesSet` map to track which configuration keys have been explicitly modified. However, when loading an existing configuration file from disk, the `valuesSet` map remained empty.

As a result, if a user manually uncommented a default setting without changing its value(e.g., `nvidia-container-runtime.debug = "/var/log/nvidia-container-runtime.log"`), or uncommented with the `--set` flag, the toolkit would:

- Fail to recognize the key as "user-set" during the loading phase.
- Match the value against the internal hardcoded defaults.
- Re-apply the `#` comment prefix during the `Save()` operation, effectively reverting the user's manual configuration.

## Changes
- Modified the configuration loading logic to differentiate between loading an existing file and generating a fresh default configuration.
- Updated `loadConfigToml()` to traverse the TOML tree after successfully reading from a file, populating the `valuesSet` map with all keys present in the file.

## Testing Performed
- Verified that running `nvidia-ctk config --in-place` no longer re-comments uncommented lines in `/etc/nvidia-container-runtime/config.toml`.
- Verified that `nvidia-ctk config default` still generates a standard template with the correct comments and default values.

### Before Fix

```bash
$ ./nvidia-ctk config default -o /etc/nvidia-container-runtime/config.toml 
$ ./nvidia-ctk config |grep debug
#debug = "/var/log/nvidia-container-toolkit.log"
#debug = "/var/log/nvidia-container-runtime.log"
$ ./nvidia-ctk config --set nvidia-container-runtime.debug=/var/log/nvidia-container-runtime.log --in-place
$ ./nvidia-ctk config |grep debug
#debug = "/var/log/nvidia-container-toolkit.log"
#debug = "/var/log/nvidia-container-runtime.log"
```
### After Fix
```bash
$ ./nvidia-ctk config default -o /etc/nvidia-container-runtime/config.toml 
$ ./nvidia-ctk config |grep debug
#debug = "/var/log/nvidia-container-toolkit.log"
#debug = "/var/log/nvidia-container-runtime.log"
$ ./nvidia-ctk config --set nvidia-container-runtime.debug=/var/log/nvidia-container-runtime.log --in-place
$ ./nvidia-ctk config |grep debug
#debug = "/var/log/nvidia-container-toolkit.log"
debug = "/var/log/nvidia-container-runtime.log"
```

### DIFF
--- a/api/config/v1/toml.go
+++ b/api/config/v1/toml.go
@@ -98,8 +98,31 @@ func (o options) loadConfigToml() (*Toml, error) {
 	}
 	defer tomlFile.Close()
 
-	return loadConfigTomlFrom(tomlFile)
+	t, err := loadConfigTomlFrom(tomlFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load specified config file: %w", err)
+	}
 
+	for _, key := range t.tree.Keys() {
+		t.markKeysSet(key, "")
+	}
+	return t, nil
+}
+
+func (t *Toml) markKeysSet(key string, prefix string) {
+	fullKey := key
+	if prefix != "" {
+		fullKey = prefix + "." + key
+	}
+
+	t.valuesSet[fullKey] = true
+
+	subTree := t.tree.Get(fullKey)
+	if nextTree, ok := subTree.(*toml.Tree); ok {
+		for _, nextKey := range nextTree.Keys() {
+			t.markKeysSet(nextKey, fullKey)
+		}
+	}
 }
 
 func defaultToml() (*Toml, error) {
--- a/api/config/v1/toml_test.go
+++ b/api/config/v1/toml_test.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"bytes"
+	"os"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


elezar (commented):
As a general comment, could we add unit tests to validate the original behaviour and then verify that the fix addresses this.

sunjq1:
> As a general comment, could we add unit tests to validate the original behaviour and then verify that the fix addresses this.

I have updated the PR with a unit test: `TestLoadConfigToml` in `api/config/v1/toml_test.go`.

### Before Fix
```bash
$ go test
--- FAIL: TestLoadConfigToml (0.00s)
    --- FAIL: TestLoadConfigToml/save_uncommented_configuration_items_to_file (0.00s)
        toml_test.go:388: 
                Error Trace:    nvidia-container-toolkit/api/config/v1/toml_test.go:388
                Error:          Not equal: 
                                ...
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -7,3 +7,3 @@
                                 [nvidia-container-cli]
                                -debug = "/var/log/nvidia-container-toolkit.log"
                                +#debug = "/var/log/nvidia-container-toolkit.log"
                                 environment = []
                                @@ -18,3 +18,3 @@
                                 [nvidia-container-runtime]
                                -debug = "/var/log/nvidia-container-runtime.log"
                                +#debug = "/var/log/nvidia-container-runtime.log"
                                 log-level = "info"
                Test:           TestLoadConfigToml/save_uncommented_configuration_items_to_file
FAIL
exit status 1
FAIL    github.com/NVIDIA/nvidia-container-toolkit/api/config/v1        0.010s
```
### After Fix 
```bash
$ go test
PASS
ok      github.com/NVIDIA/nvidia-container-toolkit/api/config/v1        0.009s
```

sunjq1:
@elezar hi, could you PTAL

cdesiniotis (approved):
@sunjq1 apologies for the delay. Thanks for your contribution. Would you mind [signing your commits](2fe4baf403c3b3c078a14ade11bb826c9f5308f0) and squashing them?

sunjq1:
> @sunjq1 apologies for the delay. Thanks for your contribution. Would you mind [signing your commits](2fe4baf403c3b3c078a14ade11bb826c9f5308f0) and squashing them?

@cdesiniotis Addressed. I squashed the changes into a single signed commit with DCO sign-off. I also fixed the errcheck findings in `api/config/v1/toml_test.go` by checking the `cfgToml.Save` return values.

sunjq1:
@cdesiniotis hi, I rebased the PR onto the latest main, and I verified locally with:
```
go test ./api/config/v1
golangci-lint run -v --timeout 5m
```
Both commands pass locally. Could you please rerun the checks for the latest commit?

cdesiniotis (approved):

PR #1726, opened 2026-03-12
Bump version to 1.20.0~dev post-release

cdesiniotis:


versions.mk
```diff
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 LIB_NAME := nvidia-container-toolkit
-LIB_VERSION := 1.19.0
-LIB_TAG :=
+LIB_VERSION := 1.20.0
+LIB_TAG := dev
```
elezar, line 17: As a note. This is what fills up our Shipit view with versions that we don't want to release. Does it make sense to use something like `rc.0` here instead so that ordering at least remains consistent?

cdesiniotis: Could you clarify how adding `rc.0` here would help with the ordering?

cdesiniotis: Updated to `rc.0` as you suggested.

elezar: To be clear, this could be solved elsewhere in the stack (e.g. where we construct the kitmaker build version), but see the following example:

Screenshot 2026-04-13 at 11 01 42


Note that the `1.19.0~dev` releases ALWAYS come before the `1.19.0` releases.

Looking at the how we construct the kitmaker version, we could use `dev` here and then change the following line https://github.com/elezar/nvidia-container-toolkit/blob/5eee5ce7b8ab19012f5aabdafd525f21b3a54496/scripts/release-kitmaker-artifactory.sh#L196 too:

```
diff --git a/scripts/release-kitmaker-artifactory.sh b/scripts/release-kitmaker-artifactory.sh
index 58ea8a61..0f4780fe 100755
--- a/scripts/release-kitmaker-artifactory.sh
+++ b/scripts/release-kitmaker-artifactory.sh
@@ -193,7 +193,7 @@ function upload_archive() {
 component="nvidia_container_toolkit"
 version="${VERSION%~rc.*}"
 version_suffix=$(date -r "${IMAGE_EPOCH}" '+%Y.%m.%d.%s' || date -d @"${IMAGE_EPOCH}" '+%Y.%m.%d.%s')
-kitmaker_version="${VERSION%~rc.*}.${version_suffix}"
+kitmaker_version="${VERSION%~*}.${version_suffix}"
 kitmaker_os="linux"
 
 # create_and_upload creates a kitmaker archive for the specified component, os, and arch and uploads it.
```

That might be a cleaner long-term change.

cdesiniotis: Got it. Your proposal makes sense. I have changed `LIB_TAG` back to `dev` to align with what was done previously. I have opened https://github.com/NVIDIA/nvidia-container-toolkit/pull/1764 to hopefully clean up our future builds in ShipIt.

elezar: Note that we should only merge this AFTER we have rebased on #1764 otherwise we will produce at least one build that has a `1.20.0~dev.*` kitmaker version.

PR #1745, opened 2026-03-26
fix: Additional GIDs are dropped due to file mode mask

elezar:
The filemode returned from DeviceFromPath clears the bits that are track whether the device is a character device. This means that the check that ensures that the mode includes the character device bits always fails and no additional GIDs are detected.

This change removes the check. In practice this should have no effect since we only ever detect char devices.

### DIFF
--- a/internal/edits/device.go
+++ b/internal/edits/device.go
@@ -130,9 +130,6 @@ func (d *device) getAdditionalGIDs(dn *specs.DeviceNode) []uint32 {
 	if dn.FileMode == nil {
 		return nil
 	}
-	if dn.FileMode.Type()&os.ModeCharDevice == 0 {
-		return nil
-	}
 	if permission := dn.FileMode.Perm(); isWorldReadable(permission) && isWorldWriteable(permission) {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


internal/edits/device.go
cdesiniotis, line 76: Question -- I do see that `dn` returned here does have a member `Type`. What if we used that to set the corresponding `Type` field in the CDI spec for this device? Or is there a reason why do do not set the `Type` field?

cdesiniotis (approved):
One non-blocking question, otherwise this lgtm.

I did add one unit test to your commit, which fails without the change.

internal/edits/device.go
elezar: I think we could set the type field too, and probably should, but I didn't want to make the change as part of this PR to make backporting the fix clearer. We would also need to check the interaction with this and the code in the CDI package for filling the missing device information.


For reference see:
* https://github.com/cncf-tags/container-device-interface/blob/db8fd53e4587ebcb5cd240a0054e75e6f6d55799/pkg/cdi/container-edits_unix.go#L80
* https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#devices

PR #1754, opened 2026-04-07
[NRI Plugin] get low-level runtime bin paths from previous install

tariq1890:
This PR fixes the "nvidia-operator-validator pod stuck indefinitely in Terminating" issue observed in clusters that use `cri-o + crun`

### Problem

The toolkit by default sources a list of the absolute paths of low-level runtime binaries from the containerd/crio runtime config TOML file. This list is then prepended to the `runtimes= ["runc", "crun"]` list in the toolkit's config TOML file. Through this method, the toolkit binds the `nvidia-container-runtime` wrapper script to the correct low-level runtime binary. With the switch to the NRI Plugin mode, the toolkit no longer reads from the containerd/crio runtime config TOML file, so it does not have a way to inject the correct low-level runtime binary to the `nvidia-container-runtime` wrapper script. As a result, pods that were created with the `nvidia-container-runtime` bound to the low-level-runtime binary (sourced from the container runtime config TOML) before NRI Plugin enablement cannot be deleted afterwards as the wrapper-script just binds to `runc` instead (taken from the default list `runtimes = ["runc", "crun"]`. This discrepancy in `nvidia-container-runtime` before and after NRI Plugin enablement causes the `KillContainer` functions performed by the container runtime to fail leaving the pod stuck in Terminating permanently

### Solution

When NRI Plugin is enabled, we turn to the toolkit config toml file installed prior to the NRI plugin enablement. This way, we retrieve the list of low-level runtime binary paths used previously. This way, the toolkit is able to successfully terminate the containers that were created with`nvidia-container-runtime` and pointing to the correct low-level runtime binary in doing so. This unblocks migration from non-NRI to NRI plugin mode on an existing cluster

### DIFF
--- a/cmd/nvidia-ctk-installer/main.go
+++ b/cmd/nvidia-ctk-installer/main.go
@@ -273,7 +273,19 @@ func (a *app) Run(ctx context.Context, c *cli.Command, o *options) error {
 
 	runtimeConfigurer := runtime.NewConfigurer(o.runtime, o.noRuntimeConfig, o.enableNRIPlugin)
 
-	if len(o.toolkitOptions.ContainerRuntimeRuntimes) == 0 {
+	if o.enableNRIPlugin {
+		// When NRI Plugin is enabled, the toolkit no longer interacts with the cri-o/containerd runtime config TOML and
+		// therefore can no longer source the list of low-level runtime binary paths used by the container runtime.
+		// This becomes an issue when migrating from non-NRI to NRI where the low-level runtime binary path previously
+		// referenced is still needed after enabling the NRI Plugin mode. To do this, we retrieve the toolkit config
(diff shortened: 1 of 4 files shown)


tariq1890:
Scenarios tested:

- OpenShift cluster (cri-o + crun)
- Vanilla Kubernetes on Ubuntu (cri-o + runc)
- Vanilla Kubernetes on RHEL (cri-o + crun)
- Vanilla Kubernetes on Ubuntu (containerd + runc)
- Vanilla Kubernetes upgrade from v25.10.1 to v26.3.1 + toolkit version built from this PR (containerd + runc)

cdesiniotis (approved):
LGTM. This aligns with what we discussed over slack.

PR #1755, opened 2026-04-08
[ci] build arch-specific images natively

tariq1890:


.github/workflows/image.yaml
```diff
@@ -114,10 +187,41 @@ jobs:
...
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build Manifest
+        env:
+          MULTIARCH_IMAGE: ghcr.io/nvidia/container-toolkit:${{ inputs.version }}
```
cdesiniotis: Question -- should we parameterize the repository name so that this can work (potentially) on forks? E.g. use `${{ github.repository }}` instead of hardcoding `ghcr.io/nvidia/container-toolkit` as the image name?

tariq1890: Yeah, that makes sense. We could do that

cdesiniotis (approved):

.github/workflows/image.yaml
tariq1890: Can you check now?

cdesiniotis (approved):

.github/workflows/image.yaml
cdesiniotis: Thanks!

PR #1782, opened 2026-04-16
fix systemd unit conditions to work on WSL2

elezar:
On WSL2, nvidia-smi lives at /usr/lib/wsl/lib/nvidia-smi rather than /usr/bin or /usr/sbin, and the NVIDIA kernel modules (nvidia.ko / nvidia-current.ko) are not present since GPU access goes through the DirectX Graphics infrastructure (/dev/dxg).

Add /usr/lib/wsl/lib/nvidia-smi as a third OR ConditionPathExists and extend ExecCondition to succeed when /dev/dxg exists so the service runs on WSL2 without breaking the check on standard driver installs.


cdesiniotis (approved):
Thanks @elezar. LGTM.

PR #1784, opened 2026-04-17
Expand cudacompat hook to always inspect libcuda.so ELF header if available

cdesiniotis:
This commit reverts  https://github.com/NVIDIA/nvidia-container-toolkit/commit/08bf5839aeeb0e940fd6dcb53a570d808406efbc which made it so that the libcuda.so ELF header was only ever inspected on certain Tegra systems, like Orin. We now leverage the libcuda.so ELF header in all cases (if available) to determine if the CUDA compat libraries bundled in the container should be used in favor of the host driver libraries. This allows us to support CUDA minor version compatibility.


cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
```diff
@@ -140,6 +158,36 @@ func (containerVersion cudaVersion) UseCompat(hostVersion string) bool {
 	return semver.Compare(normalizeVersion(containerVersion), normalizeVersion(hostVersion)) > 0
 }
 
+// normalizeVersion converts the given version into a valid semantic version.
+// This function will always return a string in the format of vMAJOR.MINOR.PATCH
+// It accounts for version strings that have leading zeros, which is common
+// in NVIDIA driver version strings. For example, 570.211.01 will be converted to
+// v570.22.1
```
elezar, line 169: Could we comment on why dropping leading zeros is important here? My assumption is that we need to ensure that `570.211.1` and `570.211.01` compare as equal, meaning that we may want to add a `compareVersions` function which calls `normalizeVersion` and has the relevant tests:
```
func compareVersions[T string | cudaVersion, O string | cudaVersion](this T, other O) int {
    return semver.Compare(normalizeVersion(this), normalizeVersion(other))
}
```

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
```diff
@@ -140,6 +158,36 @@ func (containerVersion cudaVersion) UseCompat(hostVersion string) bool {
...
+		major = "0"
+	}
+	if minor == "" {
+		minor = "0"
+	}
+	if patch == "" {
+		patch = "0"
+	}
+
+	return "v" + fmt.Sprintf("%s.%s.%s", major, minor, patch)
```
elezar: What about this as an option:

```suggestion
	majorMinorPatch := []string{"0", "0", "0"}

	versionParts := strings.SplitN(strings.TrimPrefix(string(v), "v"), ".", 3)
	for i, versionPart := range versionParts {
		majorMinorPatch[i] = strings.TrimLeft(versionPart, "0")
	}

	return "v" + strings.Join(majorMinorPatch, ".")
```

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
cdesiniotis: Version strings with leading zeros are not valid semantic versions. And from https://pkg.go.dev/golang.org/x/mod/semver#Compare:
```
An invalid semantic version string is considered less than a valid one. All invalid semantic version strings compare equal to each other.
```
As a result, calling `semver.Compare("575.57.08", "575.10.10")` would incorrectly return `-1` because the first argument was not a valid semantic version. I wanted to continue leveraging `semver.Compare()` so I decided to expand the logic in our `normalizeVersion()` method.

cdesiniotis: I have added a `compareVersions()` method as you have suggested and added some unit tests.

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
cdesiniotis: Updated!

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
elezar: One question / comment from my side. Does it make sense to add this fix for incorrect version comparisons as a separate commit / PR independently of the program flow changes?

cdesiniotis: Makes sense. I have separated the commits in this PR.

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
```diff
@@ -137,9 +155,28 @@ func (containerVersion cudaVersion) UseCompat(hostVersion string) bool {
...
 
-	return semver.Compare(normalizeVersion(containerVersion), normalizeVersion(hostVersion)) > 0
+	return compareVersions(containerVersion, hostVersion) > 0
+}
+
+func compareVersions[T string | cudaVersion, O string | cudaVersion](this T, other O) int {
+	return semver.Compare(normalizeVersion(this), normalizeVersion(other))
 }
 
+// normalizeVersion converts the given version into a valid semantic version.
```
tariq1890: Shouldn't this be a separate commit/PR? It's not directly related to the change as suggested by the PR description, correct?

cdesiniotis: I pulled this change into a separate commit. There are two commits introduced in this PR.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -196,12 +199,14 @@ func (m command) useCompatLibraries(libcudaCompatPath string, hostDriverVersion
 		return false, nil
 	}
 
+	// If we reach this point, it means we could not read the ELf header but
```
tariq1890:
```suggestion
	// If we reach this point, it means we could not read the ELF header but
```

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -177,16 +177,19 @@ func (m command) getContainerForwardCompatDir(containerRoot containerRoot, o *op
 }
 
 func (m command) useCompatLibraries(libcudaCompatPath string, hostDriverVersion string, hostCUDAVersion string) (bool, error) {
+	// First check the ELF header of the libcuda.so included in the compat directory.
+	// If this is present, we use the ELF header to determine whether the CUDA compat
+	// libraries in the container should be used over the host driver libraries.
+	compatDriverVersion := strings.TrimPrefix(filepath.Base(libcudaCompatPath), "libcuda.so.")
+	cudaCompatHeader, _ := GetCUDACompatElfHeader(libcudaCompatPath)
```
tariq1890: let's not ignore the error please. If we consider the error to be non-blocking, we should atleast log it

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
```diff
@@ -118,14 +119,31 @@ func getCUDAFwdCompatibilitySection(lib *elf.File) *elf.Section {
...
 
-	return h.CUDAVersion.UseCompat(hostCUDAVersion)
+	if compatDriverVersion == "" || hostDriverVersion == "" {
+		if hostCUDAVersion != "" {
+			return h.CUDAVersion.UseCompat(hostCUDAVersion)
+		}
+		return false
+	}
+
+	hostDriverMajor, err := extractMajorVersion(hostDriverVersion)
```
tariq1890: Instead of validating the driver versions in this method, why not pass in SemVer structs as arguments to the `UseCompat` method? This way, you don't have to parse driver versions in this method and do the error handling for it. The parsing  and validating of driver versions can be done prior to calling `UseCompat`.

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
```diff
@@ -137,9 +155,28 @@ func (containerVersion cudaVersion) UseCompat(hostVersion string) bool {
...
 }
 
+// normalizeVersion converts the given version into a valid semantic version.
+// This function will always return a string in the format of vMAJOR.MINOR.PATCH
+// It accounts for version strings that have leading zeros, which is common
+// in NVIDIA driver version strings. For example, 570.211.01 will be converted to
+// v570.22.1
 func normalizeVersion[T string | cudaVersion](v T) string {
-	return "v" + strings.TrimPrefix(string(v), "v")
+	majorMinorPatch := []string{"0", "0", "0"}
```
tariq1890: Consider using `Masterminds/semver` to handle the leading zeros. Pasting sample code below:

```
package main

import (
	"fmt"
	"github.com/Masterminds/semver/v3"
)

func main() {
	// Normal parsing throws an error
	_, err := semver.NewVersion("1.02.3")
	fmt.Println(err) // Output: ...request version string... literal leading zeros not allowed

	// Coerce relaxes constraints and auto-strips leading zeros
	v, _ := semver.NewCoercedVersion("1.02.3")
	fmt.Println(v.String()) // Output: 1.2.3
}
```

tariq1890: [Masterminds/semver](https://github.com/Masterminds/semver/blob/master/go.mod) also doesn't import any other deps, so importing it should not affect our dep tree.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
cdesiniotis: Updated.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
cdesiniotis: Updated. We log a warning message now if an error is returned.

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
cdesiniotis: As suggested, I have updated to using `Masterminds/semver` to parse and compare version strings throughout.

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
cdesiniotis: As suggested, I have updated to using `Masterminds/semver` to parse and compare version strings throughout. I have removed the `compareVersions()` and `normalizeVersion()` helper methods as they are no longer needed.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
```diff
@@ -190,13 +190,36 @@ func (m command) getContainerForwardCompatDir(containerRoot *root, o *options) (
...
+	if err != nil {
+		m.logger.Warningf("failed to parse host CUDA version %q: %v", hostCUDAVersion, err)
+	}
+
 	// First check the ELF header of the libcuda.so included in the compat directory.
 	// If this is present, we use the ELF header to determine whether the CUDA compat
 	// libraries in the container should be used over the host driver libraries.
-	compatDriverVersion := strings.TrimPrefix(filepath.Base(libcudaCompatFile.Name()), "libcuda.so.")
 	cudaCompatHeader, _ := GetCUDACompatElfHeaderFromReader(libcudaCompatFile)
+	if err != nil {
```
tariq1890, line 218: we need to capture the error in L217

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
tariq1890: Nice, looks great!

cmd/nvidia-cdi-hook/cudacompat/cuda-elf-header.go
tariq1890: So `Masterminds/semver`'s Parse handles the leading zeroes out of the box?

tariq1890:
Thanks @cdesiniotis, added one more comment.

cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
cdesiniotis: Great catch! Should be fixed now.

tariq1890 (approved):

PR #1790, opened 2026-04-21
feat(nvcdi): Allow IPC sockets to not be discovered.

LandonTClipp:
There are cases we are dealing with where the containers should not have access to any NVIDIA IPC sockets like fabricmanager/persistenced. We want the ability to disable this.

This commit adds a feature flag `disable-ipc-discoverer` that if provided will cause the IPC discoverer.Discoverer to be nil. Ultimately, this is needed so that the CDI spec file generated by k8s-device-plugin does not include these sockets.

### DIFF
--- a/pkg/nvcdi/api.go
+++ b/pkg/nvcdi/api.go
@@ -96,4 +96,8 @@ const (
 	// FeatureNoAdditionalGIDsForDeviceNodes disables the injection of additional GIDs
 	// for a device node when the node is not readable and writable by the user.
 	FeatureNoAdditionalGIDsForDeviceNodes = FeatureFlag("no-additional-gids-for-device-nodes")
+
+	// FeatureDisableIPCDiscoverer disables the inclusion of IPC sockets
+	// (nvidia-persistenced, nvidia-fabricmanager, MPS) in the CDI spec.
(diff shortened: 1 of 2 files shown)


cdesiniotis (approved):
Thanks for the contribution @LandonTClipp. 

I don't have any objections to introducing a feature flag to opt-out of this behavior. My one question would be -- are there any use cases where application containers need access to either the nvidia-persistenced or nvidia-fabricmanager sockets? IIUC excluding the nvidia-persistenced socket would only lead to cosmetic differences (e.g. running nvidia-smi in the container would report persistence mode as off even when it is on) and should not have any functional impact. I am not entirely certain about nvidia-fabricmanager.

LandonTClipp:
> I don't have any objections to introducing a feature flag to opt-out of this behavior. My one question would be -- are there any use cases where application containers need access to either the nvidia-persistenced or nvidia-fabricmanager sockets?

Are you asking whether containers would need access to one but not the other? Perhaps. In that case I think a more granular feature flag could be used, but I think this change doesn't prevent us from adding that additional granularity if we wanted to in the future. For our particular workload sandboxing use case, _any_ host NVIDIA IPC socket needs to be totally blocked.

cdesiniotis:
>Are you asking whether containers would need access to one but not the other?

Yes. But as you said, a more granular flag is out-of-scope for this PR.

Issue #1470, opened 2025-11-19
Add support for requirements checks to CDI


elezar:
At present CDI mode does not support checking of the `NVIDIA_REQUIRE_` envvars. There is some logic implemented for CSV mode, but this is only triggered when the `nvidia` runtime is invoked in `csv` mode.

This should be extended to support the general CDI case.

hiSandog:
CDI support should preserve the same requirement semantics and diagnostics as CSV mode, but the source and timing of `NVIDIA_REQUIRE_*` evaluation need to be explicit. Tests should cover satisfied/unsatisfied driver, CUDA, architecture, and brand constraints for both modes, including multiple devices. A failure should identify the exact unsatisfied expression before container startup rather than silently changing device selection.

PR #1795, opened 2026-04-29
Add support for requirements checks to CDI

JunAr7112:
This PR is aimed at addressing this [issue](https://github.com/NVIDIA/nvidia-container-toolkit/issues/1470). Essentially, at present only the legacy mode will check for NVIDIA_REQUIRE_ envvars. This PR will create a common checkRequirements function (with helpers to convert to semver format and get brand reqs with NVML) for both csv mode and CDI mode to go through the NVIDIA_REQUIRE_ envvars.

This was tested by deploying a pod with an invalid envvar value and verifying that the checks would stop deployment:

// CDI-only negative test: container create should fail when NVIDIA_REQUIRE_* cannot be satisfied (e.g. cuda>=99.0 on any real host). Uses RuntimeClass `nvidia` (CDI / toolkit mode), not `nvidia-legacy`.

apiVersion: v1
kind: Pod
metadata:
  name: require-cuda-fail
spec:
  runtimeClassName: nvidia
  restartPolicy: Never
  containers:
    - name: c
      image: ubuntu:22.04
      command: ["sleep", "3600"]
      resources:
        limits:
          nvidia.com/gpu: "1"
      env:
        - name: NVIDIA_VISIBLE_DEVICES
          value: "all"
        - name: NVIDIA_DRIVER_CAPABILITIES
          value: "all"
        - name: NVIDIA_REQUIRE_CUDA
          value: "cuda>=99.0"
        - name: NVIDIA_REQUIRE_DRIVER
          value: "driver>=9999.0.0"
..................................
 Events:
  Type     Reason     Age   From               Message
  
  Normal   Scheduled  12m   default-scheduler  Successfully assigned default/require-cuda-fail to ipp1-3167
  Normal   Pulled     12m   kubelet            Container image "ubuntu:22.04" already present on machine
  Normal   Created    12m   kubelet            Created container c
  Warning  Failed     12m   kubelet            Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: could not apply required modification to OCI specification: error modifying OCI spec: requirements not met: unsatisfied condition: driver>=9999.0.0 (driver=595.58.3)

### DIFF
--- a/internal/modifier/image_requirements.go
+++ b/internal/modifier/image_requirements.go
@@ -1,0 +1,200 @@
+/**
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package modifier
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+	"golang.org/x/mod/semver"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/config/image"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/cuda"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/lookup/root"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/requirements"
+)
+
+// checkRequirements evaluates NVIDIA_REQUIRE_* constraints using the host
+// CUDA driver API version from libcuda, the NVIDIA display driver version from
+// the driver root (libcuda / libnvidia-ml soname), the compute capability of
+// CUDA device 0, and (when requirements reference brand) the GPU product brand
+// from NVML. It is used for CSV and CDI / JIT-CDI modes.
+func checkRequirements(logger logger.Interface, image *image.CUDA, driver *root.Driver) error {
+	if image == nil || image.HasDisableRequire() {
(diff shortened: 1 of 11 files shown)


elezar (changes requested):
One note here. It is not sufficient to run the check in the modifier, we would have to ensure that we generate a hook that implements the check in some form. The driver version etc. are known at the point of spec generation, and one would have to inspect the envvars in the container.

As an additional note, this would be an ideal candidate to move to a `createRuntime` hook.

internal/modifier/image_requirements.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/lookup/root"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/requirements"
+)
+
+// checkRequirements evaluates NVIDIA_REQUIRE_* constraints using the host
+// CUDA driver API version from libcuda, the NVIDIA display driver version from
+// the driver root (libcuda / libnvidia-ml soname), the compute capability of
+// CUDA device 0, and (when requirements reference brand) the GPU product brand
+// from NVML. It is used for CSV and CDI / JIT-CDI modes.
```
elezar, line 38: Note that there are cases where `libcuda.so` is not applicable (if we're not injecting actuall GPU devices, for example).

internal/modifier/image_requirements.go
```diff
@@ -0,0 +1,200 @@
...
+	brand, ok := brandTypeToRequirementString(brandType)
+	if !ok {
+		return "", fmt.Errorf("unknown NVML brand type %v", brandType)
+	}
+	return brand, nil
+}
+
+// brandTypeToRequirementString maps NVML brand enums to lowercase tokens
+// consistent with typical NVIDIA_REQUIRE_* image constraints.
+func brandTypeToRequirementString(b nvml.BrandType) (string, bool) {
```
elezar, line 161: Question: is this something that we already have access to in `go-nvlib`?

internal/modifier/image_requirements.go
```diff
@@ -0,0 +1,200 @@
...
+	r := requirements.New(logger, imageRequirements)
+
+	cudaVersion, err := cuda.Version()
+	if err != nil {
+		logger.Warningf("Failed to get CUDA version: %v", err)
+	} else {
+		r.AddVersionProperty(requirements.CUDA, cudaVersion)
+	}
+
+	compteCapability, err := cuda.ComputeCapability(0)
```
elezar, line 62: Here we're always using the first device (which was fine for older Tegra-based systems), but this does not map to multi-device systems especially if they're heterogeneous.

PR #1804, opened 2026-05-07
Fix handling of CUDA Compat header on Orin systems

elezar:
On Orin systems, the CUDA compat header included in libcuda.so in the container has two problems:

1. The padding is different to that of the standard compat libraries which would cause an out-of-bounds panic when processing the relevant note.
2. The JSON data for the CUDA 13.2 compat libraries is malformed meaning that no CUDA compat information can be extracted.

This PR makes the following changes:
1. Updates the code to be resillient to out of bounds errors.
2. Adds a test case including the malformed JSON and an example including the corrected library.

Note that the libraries used as test data have been truncated to only include the relevant notes.


cdesiniotis (approved):
Thanks @elezar! CI failure is known and is unrelated to this PR (is fixed on main already).

PR #1818, opened 2026-05-12
add support for injecting /dev/dri* devices nodes for gfx MIGs

tariq1890:
Currently, the container toolkit skips adding the `/dev/dri` device nodes to a device if it's found to be a MIG device. Today, with the availability of GPUs like [RTX Pro 6000D Blackwell](https://docs.nvidia.com/datacenter/tesla/mig-user-guide/supported-mig-profiles.html#rtx-pro-6000-blackwell-mig-profiles), MIG slices with graphics capabilities can be created. 

In this PR, we update the NVSandboxutils and NVML discoverers to check if the detected MIG device supports graphics and inject the `/dev/dri` device nodes if true.

### DIFF
--- a/internal/platform-support/dgpu/dgpu.go
+++ b/internal/platform-support/dgpu/dgpu.go
@@ -72,6 +72,11 @@ func NewForMigDevice(d device.Device, mig device.MigDevice, opts ...Option) (dis
 		return nil, err
 	}
 	o.isMigDevice = true
+	migProfile, err := mig.GetProfile()
+	if err != nil {
+		return nil, fmt.Errorf("error getting MIG Profile attributes: %w", err)
+	}
+	o.migAttributes = migProfile.GetInfo().Attributes
 
 	var discoverers []discover.Discover
 	var errs error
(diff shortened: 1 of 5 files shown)


elezar (commented):
This looks reasonable @tariq1890. Could we link the related `go-nvlib` PRs that will be required for this in the description?

One other thing to note. This handles the injection of these devices when nvidiasandbox-utils is used, but not when *pure* nvml is used. Is this a concern?

tariq1890:
Thanks for the review @elezar . There are no go-nvlib or go-nvml changes required for this PR. These toolkit changes just work as they are.

Regarding the NVML discoverer, I thought the changes would just work looking at the code. I can try testing that with the nvsandboxutils discoverer disabled

tariq1890:
I stand corrected :). Changes are indeed needed in the NVML MIG discoverer. I'll make those changes

tariq1890:
> This handles the injection of these devices when nvidiasandbox-utils is used, but not when pure nvml is used. Is this a concern?

This is now addressed. I was able to confirm through local testing that the pure NVML discoverer will inject the `/dev/dri*` device nodes if the MIG slices support graphics.

cdesiniotis (approved):

Issue #1624, opened 2026-02-03
[Question]: Avoid retrying `nvidia-ctk cdi generate` on platforms without Nvidia GPUs


fbq:
I have a scenario where we build the same Linux image for both platforms with and without Nvidia GPUs, because we want to avoid duplication. But since https://github.com/NVIDIA/nvidia-container-toolkit/commit/4335df0d423dc74ef0952246f86df794e9adc879, IIUC, the `nvidia-cdi-refresh.service` will keep loading Nvidia drivers if there is no Nvidia GPU and generate a lot of error logs that noise out meaningful information in the logs. Would there be a way for this "Restart" behavior to be configurable? Via a config or ENV? Thanks!

elezar:
@fbq does the image also have the NVIDIA driver modules available and `nvidia-smi` installed on non-NIVDIA platforms?

With that said, given #1611 using retries may have been the incorrect mechanism to ensure that we're able to generate CDI specs.

Does adding `After=multi-user.target` to the `nvidia-cdi-refresh.service` (and removing the restart configuration):
```
diff --git a/deployments/systemd/nvidia-cdi-refresh.service b/deployments/systemd/nvidia-cdi-refresh.service
index ece65e5f..eff939a6 100644
--- a/deployments/systemd/nvidia-cdi-refresh.service
+++ b/deployments/systemd/nvidia-cdi-refresh.service
@@ -17,9 +17,7 @@ Description=Refresh NVIDIA CDI specification file
 ConditionPathExists=|/usr/bin/nvidia-smi
 ConditionPathExists=|/usr/sbin/nvidia-smi
 ConditionPathExists=/usr/bin/nvidia-ctk
-# Limit the number of successive restarts to 5 in 10 seconds.
-StartLimitBurst=5
-StartLimitIntervalSec=10s
+After=multi-user.target
 
 [Service]
 Type=oneshot
@@ -29,10 +27,6 @@ EnvironmentFile=-/etc/nvidia-container-toolkit/nvidia-cdi-refresh.env
 ExecCondition=/usr/bin/grep -qE '/(nvidia|nvidia-current)\\.ko' /lib/modules/%v/modules.dep
 ExecStart=/usr/bin/nvidia-ctk cdi generate
 CapabilityBoundingSet=CAP_SYS_MODULE CAP_SYS_ADMIN CAP_MKNOD
-# We set the service to restart on failure to ensure that a CDI spec is
-# eventually generated.
[...]

PR #1825, opened 2026-05-14
Update systemd unit conditions to execute nvidia-smi and re-add restart logic

cdesiniotis:
This ensures that the nvidia-cdi-refresh.service does not start until after the driver is loaded and available. The added benefit of running nvidia-smi is that it will load the module (if not already loaded) and create the device nodes (if not already created).

This PR also re-introduces the restart logic which was removed in https://github.com/NVIDIA/nvidia-container-toolkit/commit/5fe6b42eab11f144866de07df4f755845c4b94cf. Without the restart logic, it is possible that our service gets started too early in the boot process and executing "nvidia-smi" fails. With retries, the "nvidia-smi" invocation should eventually succeed, allowing us to generate a CDI spec.

Note, re-introducing the restart logic in the systemd service means we should re-open https://github.com/NVIDIA/nvidia-container-toolkit/issues/1624. I believe there are ways to address the use case in #1624 while still keeping the restart logic -- 1) Reducing the logs emitted by `nvidia-ctk cdi generate` and/or 2) Updating `nvidia-ctk cdi generate` to exit early with a zero exit code if no GPUs are detected on the system.

### DIFF
--- a/deployments/systemd/nvidia-cdi-refresh.service
+++ b/deployments/systemd/nvidia-cdi-refresh.service
@@ -18,6 +18,9 @@ ConditionPathExists=|/usr/bin/nvidia-smi
 ConditionPathExists=|/usr/sbin/nvidia-smi
 ConditionPathExists=|/usr/lib/wsl/lib/nvidia-smi
 ConditionPathExists=/usr/bin/nvidia-ctk
+# Limit the number of successive restarts to 5 in 10 seconds.
+StartLimitBurst=5
+StartLimitIntervalSec=10s
 
 [Service]
 Type=oneshot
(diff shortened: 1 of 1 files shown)


deployments/systemd/nvidia-cdi-refresh.service
```diff
@@ -25,6 +25,7 @@ Type=oneshot
 Environment=NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/var/run/cdi/nvidia.yaml
 EnvironmentFile=-/etc/nvidia-container-toolkit/nvidia-cdi-refresh.env
 ExecCondition=/bin/sh -c '/usr/bin/grep -qE "/(nvidia|nvidia-current)[.]ko" /lib/modules/%v/modules.dep || [ -e /dev/dxg ]'
+ExecCondition=/bin/sh -c '/usr/bin/nvidia-smi -L || /usr/sbin/nvidia-smi -L || /usr/lib/wsl/lib/nvidia-smi -L'
```
tariq1890, line 31: The `ConditionPathExists` statements above seems to check if `nvidia-smi` is path resolveable, so wouldn't it suffice to just use the binary name?

cdesiniotis: > The ConditionPathExists statements above seems to check if nvidia-smi is path resolveable

AFAIK it is checking whether those paths exist, not whether they are present in the system's `PATH`. I would expect `/usr/bin` and `/usr/sbin` to be in the `PATH` by default, not sure about `/usr/lib/wsl/lib` for WSL systems.

tariq1890 (approved):

tariq1890 (approved):

tariq1890 (approved):

PR #1827, opened 2026-05-15
Fix nvidia-cdi-refresh systemd packaging

elezar:
This PR updates the installation and configuration of the `nvidia-cdi-refresh` systemd units. This should address issues when installing the packages on non-running systems (e.g. during machine image builds).

Changes made using Codex.

### DIFF
--- a/packaging/debian/nvidia-container-toolkit-base.maintscript
+++ b/packaging/debian/nvidia-container-toolkit-base.maintscript
@@ -1,0 +1,2 @@
+rm_conffile /etc/systemd/system/nvidia-cdi-refresh.service @VERSION@~ nvidia-container-toolkit-base
+rm_conffile /etc/systemd/system/nvidia-cdi-refresh.path @VERSION@~ nvidia-container-toolkit-base
--- a/packaging/rpm/SPECS/nvidia-container-toolkit.spec
+++ b/packaging/rpm/SPECS/nvidia-container-toolkit.spec
@@ -20,6 +20,13 @@ Source6: nvidia-cdi-hook
 Source7: nvidia-cdi-refresh.service
 Source8: nvidia-cdi-refresh.path
 Source9: nvidia-cdi-refresh.env
+Source10: 90-nvidia-container-toolkit.preset
+
+%if 0%{?rhel} == 7 || 0%{?amzn} == 2
+BuildRequires: systemd
+%else
+BuildRequires: systemd-rpm-macros
+%endif
 
 Obsoletes: nvidia-container-runtime <= 3.5.0-1, nvidia-container-runtime-hook <= 1.4.0-2
 Provides: nvidia-container-runtime
@@ -31,11 +38,12 @@ Requires: nvidia-container-toolkit-base == %{version}-%{release}
 Provides tools and utilities to enable GPU support in containers.
 
 %prep
-cp %{SOURCE0} %{SOURCE1} %{SOURCE2} %{SOURCE3} %{SOURCE4} %{SOURCE5} %{SOURCE6} %{SOURCE7} %{SOURCE8} %{SOURCE9} .
+cp %{SOURCE0} %{SOURCE1} %{SOURCE2} %{SOURCE3} %{SOURCE4} %{SOURCE5} %{SOURCE6} %{SOURCE7} %{SOURCE8} %{SOURCE9} %{SOURCE10} .
 
 %install
 mkdir -p %{buildroot}%{_bindir}
-mkdir -p %{buildroot}%{_sysconfdir}/systemd/system/
+mkdir -p %{buildroot}%{_unitdir}
+mkdir -p %{buildroot}%{_presetdir}
 mkdir -p %{buildroot}%{_sysconfdir}/nvidia-container-toolkit
 
 install -m 755 -t %{buildroot}%{_bindir} nvidia-container-runtime-hook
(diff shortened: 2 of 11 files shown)


packaging/rpm/SPECS/nvidia-container-toolkit.spec
```diff
@@ -20,6 +20,13 @@ Source6: nvidia-cdi-hook
 Source7: nvidia-cdi-refresh.service
 Source8: nvidia-cdi-refresh.path
 Source9: nvidia-cdi-refresh.env
+Source10: 90-nvidia-container-toolkit.preset
+
+%if 0%{?rhel} == 7 || 0%{?amzn} == 2
+BuildRequires: systemd
+%else
+BuildRequires: systemd-rpm-macros
+%endif
```
cdesiniotis, line 29: Question -- these are build-time constraints that only apply when building the RPM packages, correct?

packaging/rpm/SPECS/nvidia-container-toolkit.spec
```diff
@@ -94,29 +103,29 @@ Provides tools such as the NVIDIA Container Runtime and NVIDIA Container Toolkit
...
-      systemctl enable --now nvidia-cdi-refresh.path || echo "Warning: Failed to enable nvidia-cdi-refresh.path" >&2
-      systemctl enable --now nvidia-cdi-refresh.service || echo "Warning: Failed to enable nvidia-cdi-refresh.service" >&2
-      
-      # Trigger CDI spec regeneration immediately after install/upgrade
-      echo "Regenerating NVIDIA CDI specification..."
-      systemctl start nvidia-cdi-refresh.service || echo "Warning: Failed to trigger CDI refresh" >&2
-      ;;
-  esac
+  systemctl start nvidia-cdi-refresh.path >/dev/null 2>&1 || :
+  systemctl start nvidia-cdi-refresh.service >/dev/null 2>&1 || :
```
cdesiniotis, line 140: Question -- what's the rationale behind not running `systemctl enabled --now` anymore before this? Is it because our systemd service always gets enabled upon installation (due to the newly added `.preset` file)?

cdesiniotis (approved):
Thanks @elezar. Overall lgtm, left some general questions.

packaging/rpm/SPECS/nvidia-container-toolkit.spec
elezar: Yes, that is my understanding. We build the packages on centos7, so the `systemd-rpm-macros` packages are not available.

packaging/rpm/SPECS/nvidia-container-toolkit.spec
elezar: The way I understand it, this does not always work as expected in cases where `systemd` is not yet in a valid state. For example, when building a system image -- which is why this is being triggered for DGX Spark.

packaging/debian/nvidia-container-toolkit-base.maintscript
```diff
@@ -0,0 +1,2 @@
+rm_conffile /etc/systemd/system/nvidia-cdi-refresh.service @VERSION@~ nvidia-container-toolkit-base
+rm_conffile /etc/systemd/system/nvidia-cdi-refresh.path @VERSION@~ nvidia-container-toolkit-base
```
cdesiniotis, line 2: IIUC this will remove the old `nvidia-cdi-refresh.service` / `nvidia-cdi-refresh.path` files if they happen to still exist in `/etc/systemd/system/` from a prior installation of `nvidia-container-toolkit-base`. Should we also do this in our rpm packaging (I don't see any equivalent logic added there)?

elezar: You're right. Let me follow-up on this.

packaging/rpm/SPECS/nvidia-container-toolkit.spec
elezar: Correct. These are build-time requirements for `rpmbuild` only. The systemd macros need to be available when building the RPM; on CentOS 7 / Amazon Linux 2 this comes via `systemd`, while newer RPM build environments can use `systemd-rpm-macros`.

packaging/rpm/SPECS/nvidia-container-toolkit.spec
elezar: Yes. The default enablement path is now the preset file plus `%systemd_post`, rather than a direct `systemctl enable --now` during install. The direct call was fragile for image-build cases where systemd is present but not in a running state. We still start the path/service best-effort on live systems, but package install success no longer depends on current systemd state. The only direct `systemctl enable` left is the `%posttrans` migration for previously enabled RPM units whose old enablement symlinks point at the removed `/etc/systemd/system` unit files.

packaging/debian/nvidia-container-toolkit-base.maintscript
elezar: Good catch. I added an RPM migration block in `%posttrans base`. The old `/etc/systemd/system/nvidia-cdi-refresh.{service,path}` files were package-owned RPM files, so RPM should remove those on upgrade. The missing piece was the unowned `systemctl enable` symlinks under `multi-user.target.wants/`, which can still point at the old `/etc/systemd/system` unit files. The new block removes only those stale old-location symlinks and re-enables the affected units against the new packaged unit location.

cdesiniotis (approved):
Thanks @elezar. I'll leave it to you to merge this PR,  just in case you were still planning to run additional testing.

elezar:
> Thanks @elezar. I'll leave it to you to merge this PR, just in case you were still planning to run additional testing.

I was going to, but didnt' get around to it. Would you have an objection to merging it and then cutting a v1.20.0-rc.1 for early integration testing?

PR #1829, opened 2026-05-15
[cri-o config] fix drop-in config directory used for config overrides

tariq1890:
As per the official CRI-O [documentation](https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.d.5.md#configuration-precedence), the cri-o runtime configuration overrides via drop-in files only work if they are placed under the `/etc/crio/crio.conf.d` directory. The current default drop-in config directory `/etc/crio/conf.d` will not be recognised by CRI-O and therefore, the configs will not be applied

### DIFF
--- a/cmd/nvidia-ctk-installer/container/runtime/crio/crio.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/crio/crio.go
@@ -47,7 +47,7 @@ const (
 
 	// Config-based settings
 	DefaultConfig       = "/etc/crio/crio.conf"
-	DefaultDropInConfig = "/etc/crio/conf.d/99-nvidia.toml"
+	DefaultDropInConfig = "/etc/crio/crio.conf.d/99-nvidia.toml"
 
 	DefaultSocket      = "/var/run/crio/crio.sock"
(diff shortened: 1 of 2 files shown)


elezar (approved):
Thanks @tariq1890. We obviously carried the value over from the initial implementation in https://github.com/NVIDIA/nvidia-container-toolkit/pull/1272 and never validated it.

As a follow-up, what are our ideas about testing drop-in files in crio? Also, I suppose that a user can always override this, so we have a workaround that can be applied. Is there a need to backport this?

cdesiniotis:
Thanks @tariq1890. We do not rely on the default value in the GPU Operator, which explains why we haven't run into any issues there: https://github.com/NVIDIA/gpu-operator/blob/6888f3b7c079719f472fa25d91c955dbff54deb8/controllers/object_controls.go#L81

cdesiniotis (approved):

tariq1890:
@elezar I think we do need to backport this. As @cdesiniotis pointed out, the gpu-operator deployments are not affected by this. However, users who setup toolkit directly on hosts with CRI-O as the container runtime will need this fix

PR #1854, opened 2026-05-27
[nri] only restrict management CDI devices to the toolkit namespace

tariq1890:
When we had first implemented the NRI Plugin, we only expected it to be used to inject GPU management CDI devices. The current namespace check worked fine with that assumption. However, we have received requests to allow for the injection of CDI devices into gpu workload pods that don't need GPU management access. As this namespace check is not really required for workload pods, we update the namespace check to only apply to pods that explicitly request the GPU management CDI device.

### DIFF
--- a/cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
+++ b/cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
@@ -101,14 +101,17 @@ func (p *Plugin) injectCDIDevices(pod *api.PodSandbox, ctr *api.Container, a *ap
 
 // parseCDIDevices processes the podSpec and determines which containers which need CDI devices injected to them
 func (p *Plugin) parseCDIDevices(pod *api.PodSandbox, key, container string) []string {
-	if p.namespace != pod.Namespace {
-		p.logger.Debugf("pod %s/%s is not in the toolkit's namespace %s. Skipping CDI device injection...", pod.Namespace, pod.Name, p.namespace)
-		return nil
-	}
-
 	cdiDeviceNames, ok := plugin.GetEffectiveAnnotation(pod, key, container)
 	if !ok {
 		return nil
+	}
+
+	if strings.Contains(cdiDeviceNames, "management.nvidia.com/gpu") {
+		if p.namespace != pod.Namespace {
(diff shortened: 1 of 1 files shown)


cmd/nvidia-ctk-installer/container/runtime/nri/plugin.go
```diff
@@ -101,16 +101,19 @@ func (p *Plugin) injectCDIDevices(pod *api.PodSandbox, ctr *api.Container, a *ap
...
-		p.logger.Debugf("pod %s/%s is not in the toolkit's namespace %s. Skipping CDI device injection...", pod.Namespace, pod.Name, p.namespace)
-		return nil
-	}
-
 	cdiDeviceNames, ok := plugin.GetEffectiveAnnotation(pod, key, container)
 	if !ok {
 		return nil
 	}
 
+	if strings.Contains(cdiDeviceNames, "management.nvidia.com/gpu") {
```
cdesiniotis, line 109: What happens if the pod annotation is not a fully-qualified CDI device name? For example, `all` is specified instead of `management.nvidia.com/gpu=all`? How do we handle that case?

tariq1890: If it's not a valid CDI device name, the container create would fail right?

cdesiniotis: Yes the container create would fail. 

In NVIDIA Container Runtime, we account for this by prepending a default CDI kind whenever a device name is not fully qualified. See https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/internal/modifier/cdi.go#L117-L119. I believe the reasoning behind this was so that the existing `NVIDIA_VISIBLE_DEVICES` interface (where one could specify `all`, `index`, `uuid`, etc) would continue to work. I am just wondering if we want to have the same behavior with NRI.

cdesiniotis: > I am just wondering if we want to have the same behavior with NRI.

This may not be as important as this is an entirely new interface -- NRI requires the use of a pod annotation and not an existing envvar like `NVIDIA_VISIBLE_DEVICES`.

cdesiniotis (approved):

PR #1846, opened 2026-05-22
fix(cudacompat): TestGlobFiles/multiple_matches determinism

damaestro:
I've had this test pass and had it fail when building for the Fedora copr. This should fix it up.

PR #1855, opened 2026-05-27
fix failing test in TestGlobFiles

tariq1890:
This commit addresses a unit test failure observed in the "multiple matches" scenario where the expected and actual arrays match but don't have the same order. We fix this by only requiring that the elements in the slice match regardless of the order.


tariq1890:
Closing in favour of #1846

PR #1861, opened 2026-06-03
feat: drop `nvidia-cdi-hook` shell shim

frezbo:
This PR drops the `nvidia-cdi-hook` shell shim that exec's `nvidia-cdi-hook.real`, the shim was not doing anything extra in this case as like the legacy shims.

This allows using toolkit installation on hosts that don't have a shell at all, when using NRI mode runc exec's `nvidia-cdi-hook`.

This is part of fixing upstream discussion at https://github.com/NVIDIA/gpu-operator/discussions/2385

### DIFF
--- a/cmd/nvidia-ctk-installer/toolkit/installer/executables.go
+++ b/cmd/nvidia-ctk-installer/toolkit/installer/executables.go
@@ -88,6 +88,13 @@ func (t *ToolkitInstaller) collectExecutables(destDir string) ([]Installer, erro
 			return nil, err
 		}
 
+		// do not create a shell wraper for nvidia-cdi-hook executable
+		if executable.path == "nvidia-cdi-hook" {
+			installers = append(installers, executableFile(executablePath))
+
+			continue
+		}
+
 		wrappedExecutableFilename := filepath.Base(executablePath)
 		dotRealFilename := wrappedExecutableFilename + ".real"
 
@@ -122,7 +129,17 @@ func (t *ToolkitInstaller) collectExecutables(destDir string) ([]Installer, erro
 	}
 
 	return installers, nil
+}
 
+// executableFile installs an executable directly, without a shell wrapper,
+// preserving its basename and mode.
+type executableFile string
+
+func (e executableFile) Install(destDir string) error {
+	dest := filepath.Join(destDir, filepath.Base(string(e)))
+	_, err := installFile(string(e), dest)
+
+	return err
 }
 
 type wrapper struct {
@@ -155,7 +172,7 @@ func (w *wrapper) Install(destDir string) error {
 		return fmt.Errorf("failed to render wrapper: %w", err)
 	}
 	wrapperFile := filepath.Join(destDir, filepath.Base(w.Source))
-	return installContent(content, wrapperFile, mode|0111)
+	return installContent(content, wrapperFile, mode|0o111)
 }
 
 func (w *render) render() (io.Reader, error) {
--- a/cmd/nvidia-ctk-installer/toolkit/installer/installer_test.go
+++ b/cmd/nvidia-ctk-installer/toolkit/installer/installer_test.go
@@ -44,7 +44,7 @@ func TestToolkitInstaller(t *testing.T) {
 
 	installer := &fileInstallerMock{
 		installFileFunc: func(s1, s2 string) (os.FileMode, error) {
-			return 0666, nil
+			return 0o666, nil
 		},
 		installContentFunc: func(reader io.Reader, s string, fileMode fs.FileMode) error {
 			var b bytes.Buffer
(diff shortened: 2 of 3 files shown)


elezar:
I know that @jfroy had a related proposal in #700. How does this proposal compare to that approach?

UPDATE: OK. I see now. Since we're not applying a specific config to the hook in terms of envvars etc there is no need for a wrapper.

I'll leave @cdesiniotis and @tariq1890 to have the final say, but I think the approach is reasonable.

frezbo:
> I know that @jfroy had a related proposal in #700. How does this proposal compare to that approach?
> 
> UPDATE: OK. I see now. Since we're not applying a specific config to the hook in terms of envvars etc there is no need for a wrapper.
> 
> I'll leave @cdesiniotis and @tariq1890 to have the final say, but I think the approach is reasonable.

I'm okay with #700 if that's going to be something that's considered, I'll wait for further feedback

frezbo:
> > I know that @jfroy had a related proposal in #700. How does this proposal compare to that approach?
> > UPDATE: OK. I see now. Since we're not applying a specific config to the hook in terms of envvars etc there is no need for a wrapper.
> > I'll leave @cdesiniotis and @tariq1890 to have the final say, but I think the approach is reasonable.
> 
> I'm okay with #700 if that's going to be something that's considered, I'll wait for further feedback

On second thoughts with moving towards NRI, and only really needing nvidia-cdi-hook binary, it think #700 is an overkill

jfroy:
I agree. #700 has lingered for 2 years with no traction.

cdesiniotis:
@frezbo could you provide the entire procedure / set of patches required to get the GPU Operator installation to work on Talos? Were there any changes required besides setting `cdi.nriPluginEnabled=true` and the changes in this PR? https://github.com/NVIDIA/gpu-operator/discussions/2385#discussioncomment-16657621 seems to indicate you hit some volume mount issues with the NVIDIA Container Toolkit daemonset.

frezbo:
> @frezbo could you provide the entire procedure / set of patches required to get the GPU Operator installation to work on Talos? Were there any changes required besides setting `cdi.nriPluginEnabled=true` and the changes in this PR? [NVIDIA/gpu-operator#2385 (comment)](https://github.com/NVIDIA/gpu-operator/discussions/2385#discussioncomment-16657621) seems to indicate you hit some volume mount issues with the NVIDIA Container Toolkit daemonset.

Yes, first this patch for Talos:

```yaml
machine:
  files:
    - content: |
        [plugins]
          [plugins."io.containerd.nri.v1.nri"]
             disable = false
      path: /etc/cri/conf.d/20-customization.part
      op: create
```

There will be change in talos to enable NRI by default, so this patch will not be needed for future releases

Then I installed with:

```yaml
helm upgrade --install -n gpu-operator gpu-operator nvidia/gpu-operator \
  --set driver.enabled=false \
  --set toolkit.enabled=true \
  --set toolkit.installDir=/opt/nvidia \
  --set toolkit.repository=ghcr.io/frezbo \
  --set toolkit.image=container-toolkit \
  --set toolkit.version=v1.19.1-nri-nowrap \
  --set hostPaths.driverInstallDir=/usr/local \
  --set cdi.enabled=true \
  --set cdi.nriPluginEnabled=true \
  --set 'toolkit.env[0].name=CONTAINERD_CONFIG' \
  --set 'toolkit.env[0].value=/run/nvidia-containerd-noop/config.toml' \
  --set 'toolkit.env[1].name=RUNTIME_DROP_IN_CONFIG' \
  --set 'toolkit.env[1].value=/run/nvidia-containerd-noop/conf.d/99-nvidia.toml'
```

The fixes in the GPU operator at https://github.com/NVIDIA/gpu-operator/pull/2514 means the toolkit.env is not neeed

I build container-toolkit and used it, I also tested with nvidia-container-toolkit extension dropping building nvidia-con tainer-toolkit binaires (but the official ones would work as is, we're just slimming it down)

frezbo:
The plan is to make NRI the official Talos supported NVIDIA installation method, this also reduces our overhead of maintaining extensions (currently we have to generate cdi specs, so with NRI that also goes away and we align more with upstream)

cmd/nvidia-ctk-installer/toolkit/installer/executables.go
```diff
@@ -88,6 +90,12 @@ func (t *ToolkitInstaller) collectExecutables(destDir string) ([]Installer, erro
 			return nil, err
 		}
 
+		if executable.installUnwrapped {
+			installers = append(installers, executableFile(executablePath))
+
+			continue
+		}
```
cdesiniotis: I slightly prefer removing the additional `installUnwrapped` boolean, and instead, making an exception for the `nvidia-cdi-hook` explicit here:
```suggestion
		# do not create a shell wrapper for the nvidia-cdi-hook executable
		if executable.path == "nvidia-cdi-hook" {
			installers = append(installers, executableFile(executablePath))
			continue
		}
```

cdesiniotis:
Thank you @frezbo for the contribution and providing additional details! From a maintainer perspective, I would prefer this PR over #700 as it is simpler and easier to maintain going forward. 

> The plan is to make NRI the official Talos supported NVIDIA installation method

I don't have any objections to this. Just note that when the NRI plugin is enabled, we no longer create the `nvidia` runtime class. So any existing pod spec that still has `runtimeClasName: nvidia` set will stop working. This might be worth mentioning in https://docs.siderolabs.com/talos/v1.13/configure-your-talos-cluster/hardware-and-drivers/nvidia-gpu#deploying-nvidia-gpu-operator after updating the install command with `--set cdi.nriPluginEnabled=true`

frezbo:
> Thank you @frezbo for the contribution and providing additional details! From a maintainer perspective, I would prefer this PR over #700 as it is simpler and easier to maintain going forward.
> 
> > The plan is to make NRI the official Talos supported NVIDIA installation method
> 
> I don't have any objections to this. Just note that when the NRI plugin is enabled, we no longer create the `nvidia` runtime class. So any existing pod spec that still has `runtimeClasName: nvidia` set will stop working. This might be worth mentioning in https://docs.siderolabs.com/talos/v1.13/configure-your-talos-cluster/hardware-and-drivers/nvidia-gpu#deploying-nvidia-gpu-operator after updating the install command with `--set cdi.nriPluginEnabled=true`

Yes, this will be for v1.14 of Talos only, so docs will be updated accordingly, The suggestions makes it simpler, I;ll update the PR

cmd/nvidia-ctk-installer/toolkit/installer/executables.go
frezbo: fixed

cdesiniotis (approved):

PR #1866, opened 2026-06-05
import text/template instead of html/template

tariq1890:
This commit replaces the "html/template" library with "text/template" as the former is not the right fit for the container-toolkit's use case. With "html/template" we face the risk of encountering bugs from HTML-escaping performed by "html/template"


elezar (approved):
I'm assuming that there are no other places where we're also using the http/template package.

PR #1876, opened 2026-06-10
Add libnvidia-nvvm70.so library to CDI spec

### LINKED ISSUE #1875: [Bug]: libnvidia-nvvm70.so missing from generated CDI spec
`libnvidia-nvvm70.so` was introduced in R575 to enable legacy Just-in-time Link Time Optimization (JIT LTO) Driver APIs for pre-CUDA 12.0 applications where PTX needs to be compiled to SASS. It was reported that a pre-CUDA 12.0 application container running on a host with the R610 driver failed with a `PTX JIT compiler library not found` error when attempting to load `libnvidia-nvvm70.so` as this library was not present in the container.

cdesiniotis:
fixes https://github.com/NVIDIA/nvidia-container-toolkit/issues/1875

### Testing

```
$ ls -ltr /usr/lib/x86_64-linux-gnu/libnvidia-nvvm*
-rw-r--r-- 1 root root 24829456 Dec  8  2025 /usr/lib/x86_64-linux-gnu/libnvidia-nvvm70.so.4
-rw-r--r-- 1 root root 78117032 Dec  8  2025 /usr/lib/x86_64-linux-gnu/libnvidia-nvvm.so.590.48.01
lrwxrwxrwx 1 root root       27 Dec  8  2025 /usr/lib/x86_64-linux-gnu/libnvidia-nvvm.so.4 -> libnvidia-nvvm.so.590.48.01

$ ./nvidia-ctk --version
NVIDIA Container Toolkit CLI version 1.20.0-dev
commit: d8d7fecdc5d83b8b897188c0bd1f2c4bae5220bb-dirty

$ ./nvidia-ctk cdi generate --output ./nvidia.yaml 2>/dev/null

$ grep -A6 "libnvidia-nvvm70" nvidia.yaml
        - hostPath: /usr/lib/x86_64-linux-gnu/libnvidia-nvvm70.so.4
          containerPath: /usr/lib/x86_64-linux-gnu/libnvidia-nvvm70.so.4
          options:
            - ro
            - nosuid
            - nodev
            - rbind
            - rprivate

$ grep -A6 "libnvidia-nvvm.so.590.48.01$" nvidia.yaml
        - hostPath: /usr/lib/x86_64-linux-gnu/libnvidia-nvvm.so.590.48.01
          containerPath: /usr/lib/x86_64-linux-gnu/libnvidia-nvvm.so.590.48.01
          options:
            - ro
            - nosuid
            - nodev
            - rbind
            - rprivate
```

### DIFF
--- a/pkg/nvcdi/driver-nvml.go
+++ b/pkg/nvcdi/driver-nvml.go
@@ -81,6 +81,10 @@ func (l *nvcdilib) NewDriverLibraryDiscoverer(version string) (discover.Discover
 	if err != nil {
 		return nil, err
 	}
+	legacyNVVMLibraryMounts, err := l.getLegacyNVVMLibraryMounts()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get legacy nvvm library mounts: %w", err)
+	}
 	explicitLibraryMounts, err := l.getExplicitDriverLibraryMounts()
 	if err != nil {
 		return nil, err
@@ -88,6 +92,7 @@ func (l *nvcdilib) NewDriverLibraryDiscoverer(version string) (discover.Discover
 
 	libraries := discover.Merge(
 		versionSuffixLibraryMounts,
+		legacyNVVMLibraryMounts,
 		explicitLibraryMounts,
 	)
 
@@ -181,6 +186,25 @@ func (l *nvcdilib) getExplicitDriverLibraryMounts() (discover.Discover, error) {
 
 	return mounts, nil
 
+}
+
+func (l *nvcdilib) getLegacyNVVMLibraryMounts() (discover.Discover, error) {
+	legacyNVMMLibrary := []string{
+		"libnvidia-nvvm70.so.*",
+	}
+
+	driverLibraryLocator, err := l.driver.DriverLibraryLocator()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get driver library locator: %w", err)
+	}
+	mounts := discover.NewMounts(
+		l.logger,
+		driverLibraryLocator,
+		l.driver.Root,
+		legacyNVMMLibrary,
+	)
+
+	return mounts, nil
 }
 
 func getUTSRelease() (string, error) {


pkg/nvcdi/driver-nvml.go
```diff
@@ -81,13 +81,18 @@ func (l *nvcdilib) NewDriverLibraryDiscoverer(version string) (discover.Discover
...
+		return nil, err
+	}
 	explicitLibraryMounts, err := l.getExplicitDriverLibraryMounts()
 	if err != nil {
 		return nil, err
 	}
 
 	libraries := discover.Merge(
 		versionSuffixLibraryMounts,
+		legacyNVVMLibraryMounts,
```
cdesiniotis, line 95: I initially added `libnvidia-nvvm70.so` to `getExplicitDriverLibraryMounts()` (which was a one line change), but I noticed that the explicit driver libraries only get added if one opts-in to the `enable-explicit-driver-libraries` feature flag (this behavior was introduced in https://github.com/NVIDIA/nvidia-container-toolkit/commit/5a3dcbe15ab65acdb0a7940e8961dab6ed400816).

tariq1890 (approved):

pkg/nvcdi/driver-nvml.go
```diff
@@ -81,13 +81,18 @@ func (l *nvcdilib) NewDriverLibraryDiscoverer(version string) (discover.Discover
 	if err != nil {
 		return nil, err
 	}
+	legacyNVVMLibraryMounts, err := l.getLegacyNVVMLibraryMounts()
+	if err != nil {
+		return nil, err
```
tariq1890: nit: can you wrap the error here?

pkg/nvcdi/driver-nvml.go
cdesiniotis: @elezar in the commit message for https://github.com/NVIDIA/nvidia-container-toolkit/commit/5a3dcbe15ab65acdb0a7940e8961dab6ed400816 you mention:
```
Instead of always injecting the list of explict libraries, we
switch these to opt-in instead. This means that they can be
generated in use cases such as GKE where they are expected,
but don't affect other use cases where they are not.
```

Do you recall what use cases were affected by having the explicit driver libraries mounted by default? I am curious as to why this was made an opt-in to begin with.

pkg/nvcdi/driver-nvml.go
cdesiniotis: done

PR #1889, opened 2026-06-15
drop the usage of the E2E_SSH_USER secret

tariq1890:
"ubuntu" being the SSH user of AWS Ubuntu EC2 instances is a well-known fact. There is not much to gain from persisting this value as a secret.

Source: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/managing-users.html


cdesiniotis (approved):

.github/workflows/e2e.yaml
Edgarruiz8585, line 116: > ~~279256497+Edgarruiz8585@users.noreply.github.com~~

PR #1905, opened 2026-07-04
docs: fix nvidia-container-runtime usage example

### LINKED ISSUE #781: nvidia-container-runtime usage example is broken/incorrect
The nvidia-container-runtime usage example is broke/incorrect: https://github.com/NVIDIA/nvidia-container-toolkit/tree/main/cmd/nvidia-container-runtime#usage-example

It doesn't work and when you look in the logs you'll find: 

time="2024-11-08T12:50:03-05:00" level=trace msg="Skipping modifier for non-create subcommand"

This is caused by the this function:

[func HasCreateSubcommand(args []string) bool {](https://github.com/NVIDIA/nvidia-container-toolkit/blob/b14209123455c1defced8318bba2d52a3506c3ed/internal/oci/args.go#L96)

Here's the code
```
// HasCreateSubcommand checks the supplied arguments for a 'create' subcommand
func HasCreateSubcommand(args []string) bool {
	var previousWasBundle bool
	for _, a := range args {
		// We check for '--bundle create' explicitly to ensure that we
		// don't inadvertently trigger a modification if the bundle directory
		// is specified as `create`
		if !previousWasBundle && IsBundleFlag(a) {
			previousWasBundle = true
			continue
		}

		if !previousWasBundle && a == "create" {
			return true
		}

		previousWasBundle = false
	}

	return false
}
```
As far as I can tell, the only way this function will ever return true is if you run `sudo nvidia-container-runtime create` first. I think the intentioon was that if you specific --bundle explicitly it would also work? Not sure, it's a bit hard understand what the bundle flag stuff actually accomplishes. 

In any event, I thought I'd file an issue incase someone else runs apon this and wonders why it doesn't work. Once you run create the spec gets modified properly and you can use it with 'run' if you want.

iacker:
This updates the `nvidia-container-runtime` usage example so it follows the runtime path described earlier in the README: the NVIDIA runtime modifies the OCI spec when a `create` command is detected, then the container can be started.

The old example used `run`, which skips the modifier path and leaves the generated `config.json` unchanged, as reported in #781.

Fixes #781

Validation:
- Confirmed the old `nvidia-container-runtime run nvidia_smi` example is no longer present.
- Confirmed the replacement example uses `create`, `start`, and cleanup with `delete`.

AI assistance disclosure:
- I used AI assistance to identify the stale example and prepare this small documentation update.

### DIFF
--- a/cmd/nvidia-container-runtime/README.md
+++ b/cmd/nvidia-container-runtime/README.md
@@ -204,7 +204,15 @@ curl -sS http://cdimage.ubuntu.com/ubuntu-base/releases/16.04/release/ubuntu-bas
 nvidia-container-runtime spec
 sed -i 's;"sh";"nvidia-smi";' config.json
 sed -i 's;\("TERM=xterm"\);\1, "NVIDIA_VISIBLE_DEVICES=0";' config.json
+sed -i 's;"terminal":true;"terminal":false;' config.json
 
-# Run the container
-sudo nvidia-container-runtime run nvidia_smi
+# Create and start the container
+sudo nvidia-container-runtime create nvidia_smi > nvidia_smi.log 2>&1
+sudo nvidia-container-runtime start nvidia_smi
+
+# View the nvidia-smi output
+cat nvidia_smi.log
+
+# Delete the container
+sudo nvidia-container-runtime delete nvidia_smi
 ```


henry118 (approved):
LGTM

cdesiniotis:
I tried out the new procedure, and hit the below error:
```
$ sudo nvidia-container-runtime create nvidia_smi
ERRO[0000] runc create failed: cannot allocate tty if runc will detach without setting console socket
```

I have modified the commit in this PR so that the procedure is functional. 

@henry118 ptal

henry118 (approved):
LGTM. Thanks for pointing out the TTY issue.

cdesiniotis:
Thank you @iacker for the contribution.

PR #1907, opened 2026-07-05
[CDI Hooks] add ability to specify OCI hook type

tariq1890:
This PR makes it possible to specify a desired OCI hook type should we want the CDI spec generator to create OCI hooks that run in stages other than `createContainer`.

### DIFF
--- a/internal/discover/hooks.go
+++ b/internal/discover/hooks.go
@@ -26,6 +26,18 @@ import (
 // A HookName represents a supported CDI hooks.
 type HookName string
 
+// An OCIHookType is the OCI lifecycle stage at which a CDI hook runs.
+type OCIHookType string
+
+const (
+	// OCIHookTypeCreateRuntime runs in the runtime namespace before the container is created.
+	OCIHookTypeCreateRuntime = OCIHookType(cdi.CreateRuntimeHook)
+	// OCIHookTypeCreateContainer runs in the container mount namespace before pivot_root.
+	OCIHookTypeCreateContainer = OCIHookType(cdi.CreateContainerHook)
+	// OCIHookTypeStartContainer runs in the container namespace after the container is started.
+	OCIHookTypeStartContainer = OCIHookType(cdi.StartContainerHook)
+)
+
 const (
 	// AllHooks is a special hook name that allows all hooks to be matched.
 	AllHooks = HookName("all")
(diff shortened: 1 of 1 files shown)


internal/discover/compat_libs.go
```diff
@@ -17,7 +17,7 @@ type EnableCUDACompatHookOptions struct {
 // NewCUDACompatHookDiscoverer creates a discoverer for a enable-cuda-compat hook.
 // This hook is responsible for setting up CUDA compatibility in the container and depends on the host driver version.
 func NewCUDACompatHookDiscoverer(logger logger.Interface, hookCreator HookCreator, o *EnableCUDACompatHookOptions) Discover {
-	return hookCreator.Create(EnableCudaCompatHook, o.args()...)
+	return hookCreator.Create(OCIHookTypeCreateContainer, EnableCudaCompatHook, o.args()...)
```
elezar: Since we're changing the API anyway, what about rather changing the type of the first argument to accept a `Hook` type that would give us the lifecycle AND the name?

```
type Hook interface {
    Type() HookType
    Name() HookName
}
```

For the default hooks we could have `HookName` implement this interface:
```
func (h HookName) Type() HookType {
    return OCIHookTypeCreateContainer
}

func (h HookName) Name() HookName {
    return h
}
```

internal/platform-support/dgpu/nvsandboxutils.go
```diff
@@ -119,7 +119,7 @@ func (d *nvsandboxutilsDGPU) Hooks() ([]discover.Hook, error) {
 		return nil, nil
 	}
 
-	hook := d.hookCreator.Create("create-symlinks", d.deviceLinks...)
+	hook := d.hookCreator.Create(discover.OCIHookTypeCreateContainer, "create-symlinks", d.deviceLinks...)
```
elezar: My concern is that the `"create-symlink"` hook (and others) MUST -- as it's currently designed -- run as a `createContainer` hook. Splitting these arguments seems to indicate that there is more flexibility than we actually have.

tariq1890: My motivation for this change is to make possible the addition of another CDI hook which will run as `createRuntime`.

The alternative to this that I can think of is adding a switch-case block within the method to select the OCI hook type based on the CDI hook name. Right now, only `createcontainer` will get selected of course. 

The drawback here is that this approach would not work well for CDI hooks which could be run at different stages. But I am not sure if this is a legitimate scenario that we even want to address.

cdesiniotis: I share Evan's concern as well. Having the caller set the hook type when currently there is only one "right" choice doesn't feel right to me.

> The alternative to this that I can think of is adding a switch-case block within the method to select the OCI hook type based on the CDI hook name. Right now, only createcontainer will get selected of course.

I would prefer this. We could always make this more flexible in the future if a use case arises for it.

internal/discover/hooks.go
```diff
@@ -195,13 +207,22 @@ func (c cdiHookCreator) Create(name HookName, args ...string) *Hook {
...
+		Lifecycle: string(c.getOCIHookType(name)),
 		Path:      c.nvidiaCDIHookPath,
 		Args:      append(c.requiredArgs(name), c.transformArgs(name, args...)...),
 		Env:       []string{fmt.Sprintf("NVIDIA_CTK_DEBUG=%v", c.debugLogging)},
 	}
 }
 
+func (c cdiHookCreator) getOCIHookType(name HookName) OCIHookType {
+	switch name {
+	case CreateSymlinksHook, ChmodHook, DisableDeviceNodeModificationHook, EnableCudaCompatHook, UpdateLDCacheHook:
```
tariq1890, line 219: I've created the method intentionally this way for readability's sake

internal/platform-support/dgpu/nvsandboxutils.go
tariq1890: @cdesiniotis @elezar  I've made the changes without modifying the interface

internal/discover/compat_libs.go
tariq1890: I've updated the PR without the API changes now

cdesiniotis (approved):

tariq1890:
Thanks for the review @elezar @cdesiniotis !

PR #1913, opened 2026-07-07
Validate imex channels for CDI/JIT-CDI mode

### LINKED ISSUE #1309: Check that IMEX Channels requested in NVIDIA_IMEX_CHANNELS are valid channel indices
With #1267 support for the `NVIDIA_IMEX_CHANNELS` was added to JIT CDI spec generation. One issue is that the validity of the channel index is not checked explicitly. This means that passing `NVIDIA_IMEX_CHANNELS=bogus` will succeed the spec generation (although it *should* fail when trying to inject the non-existent device).

We should update the logic to return a reasonable error message if:
* The channel specified is not in the range `[0, 2047]`
* The underlying device node does NOT exist.

henry118:
Fixes: #1309

This patch updates the logic to return an error if:

* The channel specified is not in the range [0, 1<<20)
* The underlying device node does NOT exist on the host

### DIFF
--- a/pkg/nvcdi/lib-imex.go
+++ b/pkg/nvcdi/lib-imex.go
@@ -18,6 +18,7 @@ package nvcdi
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -39,6 +40,11 @@ var _ deviceSpecGeneratorFactory = (*imexlib)(nil)
 
 const (
 	classImexChannel = "imex-channel"
+
+	// maxImexChannelID is the maximum valid IMEX channel ID.  Channel IDs must fit
+	// in the minor number of a dev_t (20 bits), matching the bound enforced by
+	// nvidia-container-cli in the legacy code path.
+	maxImexChannelID = (1 << 20) - 1
 )
 
 // GetCommonEdits returns an empty set of edits for IMEX devices.
@@ -73,9 +79,12 @@ func (l *imexlib) getChannelIDs(ids ...string) ([]string, error) {
 		if trimmed == "all" {
 			return l.getAllChannelIDs()
 		}
-		_, err := strconv.ParseUint(trimmed, 10, 64)
+		channelID, err := strconv.ParseUint(trimmed, 10, 64)
 		if err != nil {
-			return nil, fmt.Errorf("invalid channel ID %v: %w", id, err)
+			return nil, fmt.Errorf("invalid channel ID %s: %w", id, err)
+		}
+		if channelID > maxImexChannelID {
+			return nil, fmt.Errorf("invalid channel ID %s: must be in the range [0, %d]", id, maxImexChannelID)
 		}
 		channelIDs = append(channelIDs, trimmed)
 	}
@@ -107,13 +116,17 @@ func (l *imexlib) getAllChannelIDs() ([]string, error) {
 // GetDeviceSpecs returns the CDI device specs the specified IMEX channel.
 func (l *imexChannel) GetDeviceSpecs() ([]specs.Device, error) {
 	path := "/dev/nvidia-caps-imex-channels/channel" + l.id
+	hostPath := filepath.Join(l.devRoot, path)
+	if _, err := os.Stat(hostPath); err != nil {
+		return nil, fmt.Errorf("IMEX channel %s not found at %s: %w", l.id, hostPath, err)
+	}
 	deviceSpec := specs.Device{
 		Name: l.id,
 		ContainerEdits: specs.ContainerEdits{
 			DeviceNodes: []*specs.DeviceNode{
 				{
 					Path:     path,
-					HostPath: filepath.Join(l.devRoot, path),
+					HostPath: hostPath,
 				},
 			},
 		},
--- a/pkg/nvcdi/lib-imex_test.go
+++ b/pkg/nvcdi/lib-imex_test.go
@@ -79,3 +79,60 @@ containerEdits:
 	require.NoError(t, err)
 	require.Equal(t, expectedSpec, b.String())
 }
+
+func TestImexChannelValidation(t *testing.T) {
+	logger, _ := testlog.NewNullLogger()
+
+	moduleRoot, err := test.GetModuleRoot()
+	require.NoError(t, err)
+	hostRoot := filepath.Join(moduleRoot, "testdata", "lookup", "rootfs-1")
+
+	testCases := []struct {
+		description string
+		ids         []string
+		expectedErr string
+	}{
+		{
+			description: "valid channel",
+			ids:         []string{"0"},
+		},
+		{
+			description: "valid channel with prefix",
+			ids:         []string{"channel0"},
+		},
+		{
+			description: "non-numeric channel is rejected",
+			ids:         []string{"bogus"},
+			expectedErr: "invalid channel ID",
+		},
+		{
+			description: "channel above maximum is rejected",
+			ids:         []string{"1048576"},
+			expectedErr: "must be in the range [0, 1048575]",
+		},
+		{
+			description: "non-existent channel node is rejected",
+			ids:         []string{"5"},
+			expectedErr: "not found",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			lib, err := New(
+				WithLogger(logger),
+				WithMode(ModeImex),
+				WithDriverRoot(hostRoot),
+			)
+			require.NoError(t, err)
+
+			_, err = lib.GetDeviceSpecsByID(tc.ids...)
+
+			if tc.expectedErr != "" {
+				require.ErrorContains(t, err, tc.expectedErr)
+				return
+			}
+			require.NoError(t, err)
+		})
+	}
+}


pkg/nvcdi/lib-imex.go
```diff
@@ -73,10 +77,13 @@ func (l *imexlib) getChannelIDs(ids ...string) ([]string, error) {
 		if trimmed == "all" {
 			return l.getAllChannelIDs()
 		}
-		_, err := strconv.ParseUint(trimmed, 10, 64)
+		channelID, err := strconv.ParseUint(trimmed, 10, 64)
 		if err != nil {
 			return nil, fmt.Errorf("invalid channel ID %v: %w", id, err)
 		}
+		if channelID > maxImexChannelID {
+			return nil, fmt.Errorf("invalid channel ID %v: must be in the range [0, %d]", id, maxImexChannelID)
```
tariq1890: nit: slight preference towards specifying the formatting directive that expresses what data type is used here
```suggestion
			return nil, fmt.Errorf("invalid channel ID %s: must be in the range [0, %d]", id, maxImexChannelID)
```

tariq1890: This will be consistent with what you've done in L119

tariq1890 (approved):

cdesiniotis:
> The channel specified is not in the range [0, 2047]

I actually don't think we should enforce this. While 2048 is the default number of IMEX channels that can be created, one can configure this via the `NVreg_ImexChannelCount` kernel module parameter. I think we should just return an error if the device node does not exist.

I gave our latest release, 1.19.1, a spin and observed that this behavior already exists with our `jit-cdi` mode. If I request for an IMEX channel that does not exist, the container fails during container creation:

```
$ docker run --rm -it --runtime=nvidia --env NVIDIA_VISIBLE_DEVICES=all --env NVIDIA_IMEX_CHANNELS=99999 ubuntu:latest ls -ltr /dev/nvidia*
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: could not apply required modification to OCI specification: error modifying OCI spec: failed to apply edits for device "runtime.nvidia.com/imex-channel=99999": failed to stat CDI host device "/dev/nvidia-caps-imex-channels/channel99999": no such file or directory

Run 'docker run --help' for more information
```

If I create a dummy IMEX channel, it successfully gets injected. 
```
$ sudo mknod /dev/nvidia-caps-imex-channels/channel99999 c 240 0
$ sudo chmod 666 /dev/nvidia-caps-imex-channels/channel99999
$ docker run --rm -it --runtime=nvidia --env NVIDIA_VISIBLE_DEVICES=all --env NVIDIA_IMEX_CHANNELS=99999 ubuntu:latest ls -ltr /dev/nvidia-caps-imex-channels/*
crw-rw-rw- 1 root root 240, 0 Jul  7 23:52 /dev/nvidia-caps-imex-channels/channel99999
```

IMO I think we can close https://github.com/NVIDIA/nvidia-container-toolkit/issues/1309 without any code changes, unless we think the error message can be improved. @henry118 @tariq1890 WDYT?

henry118:
> > The channel specified is not in the range [0, 2047]
> 
> I actually don't think we should enforce this. While 2048 is the default number of IMEX channels that can be created, one can configure this via the `NVreg_ImexChannelCount` kernel module parameter. I think we should just return an error if the device node does not exist.
> 
> I gave our latest release, 1.19.1, a spin and observed that this behavior already exists with our `jit-cdi` mode. If I request for an IMEX channel that does not exist, the container fails during container creation:
> 
> ```
> $ docker run --rm -it --runtime=nvidia --env NVIDIA_VISIBLE_DEVICES=all --env NVIDIA_IMEX_CHANNELS=99999 ubuntu:latest ls -ltr /dev/nvidia*
> docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: could not apply required modification to OCI specification: error modifying OCI spec: failed to apply edits for device "runtime.nvidia.com/imex-channel=99999": failed to stat CDI host device "/dev/nvidia-caps-imex-channels/channel99999": no such file or directory
> 
> Run 'docker run --help' for more information
> ```
> 
> If I create a dummy IMEX channel, it successfully gets injected.
> 
> ```
> $ sudo mknod /dev/nvidia-caps-imex-channels/channel99999 c 240 0
> $ sudo chmod 666 /dev/nvidia-caps-imex-channels/channel99999
> $ docker run --rm -it --runtime=nvidia --env NVIDIA_VISIBLE_DEVICES=all --env NVIDIA_IMEX_CHANNELS=99999 ubuntu:latest ls -ltr /dev/nvidia-caps-imex-channels/*
> crw-rw-rw- 1 root root 240, 0 Jul  7 23:52 /dev/nvidia-caps-imex-channels/channel99999
> ```
> 
> IMO I think we can close #1309 without any code changes, unless we think the error message can be improved. @henry118 @tariq1890 WDYT?

@cdesiniotis 

I checked the legacy code path and the range check did exist in the legacy hook, although it has a much bigger range than `[0, 2047]`. I'm not sure how is 2047 determined.

https://github.com/NVIDIA/libnvidia-container/blob/d85b63168a0da1d35e09a74f2f37e5679a0c2caf/src/cli/common.c#L477-L483

As to the existence check, agreed that it's already validated when the edit is applied. But I think the spec generation case would still require this check to validate the generated devices?

cdesiniotis:
> But I think the spec generation case would still require this check to validate the generated devices?

Could you clarify what you mean by "spec generation case"? 

Today, one can generate a CDI spec for the IMEX channels by running:

```
nvidia-ctk cdi generate --mode=imex --class=imex-channel --output=/var/run/cdi/nvidia-imex-channels.yaml
```

The resulting CDI spec will only contain IMEX channels that were discovered on the system (IMEX channels need to be created before spec generation). So for example, on my system where I created a dummy IMEX channel `99999`, my CDI spec looks like:
```
cdiVersion: 0.5.0
kind: nvidia.com/imex-channel
devices:
    - name: "99999"
      containerEdits:
        deviceNodes:
            - path: /dev/nvidia-caps-imex-channels/channel99999
              hostPath: /dev/nvidia-caps-imex-channels/channel99999
    - name: all
      containerEdits:
        deviceNodes:
            - path: /dev/nvidia-caps-imex-channels/channel99999
              hostPath: /dev/nvidia-caps-imex-channels/channel99999
containerEdits:
    env:
        - NVIDIA_VISIBLE_DEVICES=void
```

To request IMEX channel 99999, you can pass `--device nvidia.com/imex-channel=99999` when launching a container.

henry118:
I found the CLI also allows us to explicitly specify device IDs, e.g.

```
$ nvidia-ctk cdi generate --mode=imex --class=imex-channel --device-id 9999 2>/dev/null
---
cdiVersion: 0.5.0
kind: nvidia.com/imex-channel
devices:
    - name: "9999"
      containerEdits:
        deviceNodes:
            - path: /dev/nvidia-caps-imex-channels/channel9999
              hostPath: /dev/nvidia-caps-imex-channels/channel9999
    - name: all
      containerEdits:
        deviceNodes:
            - path: /dev/nvidia-caps-imex-channels/channel9999
              hostPath: /dev/nvidia-caps-imex-channels/channel9999
containerEdits:
    env:
        - NVIDIA_VISIBLE_DEVICES=void
```

However, without the patch, this command runs well on everywhere. But it would fail if the running host does not have the device. Not sure if that's expected behavior.

henry118:
Also found the source of `2047` value:

https://github.com/NVIDIA/open-gpu-kernel-modules/blob/452cec62d827034798072827d3866d1881662b77/kernel-open/nvidia/nv-reg.h#L1071

It happens to be the default setting for the driver kmod. However it's adjustable when the kmod is loaded. So I think the `libnvidia-container`'s range is more correct. It's the upper bound of the minor number in Linux kernel (20 bits int).

henry118:
Looking more closely into the CDI spec, non-existence device on the host is only considered a fatal error if the device is not minimally specified.

https://github.com/cncf-tags/container-device-interface/blob/7a214f2940751725e4bb6c860381b45fd9795fbe/pkg/cdi/container-edits_unix.go#L94-L96

So I tested running a container with a manually modified CDI spec as below (by adding `type` and `major` fields):

```
cdiVersion: 0.5.0
kind: nvidia.com/imex-channel
devices:
    - name: "9999"
      containerEdits:
        deviceNodes:
            - path: /dev/nvidia-caps-imex-channels/channel9999
              hostPath: /dev/nvidia-caps-imex-channels/channel9999
              type: "c"
              major: 195
    - name: all
      containerEdits:
        deviceNodes:
            - path: /dev/nvidia-caps-imex-channels/channel9999
              hostPath: /dev/nvidia-caps-imex-channels/channel9999
              type: "c"
              major: 195
containerEdits:
    env:
        - NVIDIA_VISIBLE_DEVICES=void
```

The container will successfully start even without the imex device on the host:

```
docker run --rm -it --runtime nvidia --device=nvidia.com/imex-channel=9999 ubuntu ls -ltr /dev/nvidia-caps-imex-channels/
total 0
crw-rw-rw- 1 root root 195, 0 Jul  8 15:30 channel9999
```

Is this expected behavior?

cdesiniotis:
I am not as familiar with that code path, but it appears `runc` is creating the device node in the container (even though it does not exist on the host). 

Let's proceed with this PR, but let's update it so that we do not enforce 2047 as the maximum channel number.

cdesiniotis (approved):

cdesiniotis:
@henry118 can you update the PR description since we have changed the upper bound on the channel size?

henry118:
> @henry118 can you update the PR description since we have changed the upper bound on the channel size?

Sure. updated.

PR #1939, opened 2026-07-13
Generate app profile for EGL/Vulkan use case

### LINKED ISSUE #1899: [Bug]: JIT-CDI graphics visibility is inconsistent on multi-GPU hosts: Vulkan works for one single-GPU pod but not another
Describe the bug

On a 2-GPU host using the CDI/JIT-CDI path, two pods are created with one GPU allocated to each pod. Vulkan graphics works in the pod assigned one GPU, but fails in the pod assigned the other GPU.

From checking the Toolkit source, CDI does inject graphics files and per-GPU device nodes, including EGL/GLVND/Vulkan configs and matching /dev/dri/* nodes. However, the CDI/JIT-CDI path does not appear to create the legacy NVIDIA application profile:

/etc/nvidia/nvidia-application-profiles-rc.d/10-container.conf

That legacy file sets EGLVisibleDGPUDevices for the allocated GPU minor. Without it, NVIDIA graphics userspace appears to still make some presentation/WSI decisions from a host-global view. This can make the GPU that matches the default/global graphics view work, while the other allocated GPU fails.

A downstream patch that restores this per-container application profile for the allocated GPU minor fixes the issue.

To Reproduce

Detailed steps to reproduce the issue.

1. Use a host with 2 NVIDIA GPUs.
2. Enable Toolkit CDI/JIT-CDI mode.
3. Create two Kubernetes pods, each requesting one NVIDIA GPU.
4. Confirm each pod is assigned a different GPU.
5. In each pod, run a Vulkan WSI/presentation test, for example an app or probe that checks queue presentSupport and attempts swapchain creation.
6. Compare the two pods.

Observed result:

- The pod assigned one GPU works with Vulkan.
- The pod assigned the other GPU fails Vulkan presentation/swapchain support.
- The failing pod has a selected GPU device set, but no /etc/nvidia/nvidia-application-profiles-rc.d/10-container.conf.

Expected behavior

Each pod allocated one full GPU through CDI/JIT-CDI should get a coherent one-GPU graphics view. Vulkan presentation and swapchain creation should work regardless of which physical GPU/minor the pod receives.

Environment
nvidia-container-toolkit version: [v1.19.1]
NVIDIA Driver Version: [580.82.07]
Host OS: [Ubuntu 24.04]
Kernel Version: [6.8.0-generic]
CPU Architecture: [x86_64]
GPU Model(s): [GeForce-RTX-5070-Ti]

henry118:
Fixes #1899

The current CDI/CDI-JIT implementation does not create nvidia application profile in the CDI/JIT-CDI code path. 

In the legacy mode, the profile is created when the container has the "graphics/display" capability.

Adding a hook for CDI mode to create the nvidia application profile.

### DIFF
--- a/cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
+++ b/cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
@@ -1,0 +1,88 @@
+/**
+# SPDX-FileCopyrightText: Copyright (c) NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package updateapplicationprofile
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+)
+
+// readDirBatchSize is the number of directory entries read from the
+// container's /dev directory per ReadDir call.
+const readDirBatchSize = 100
+
+// listDeviceNodes enumerates the entries under the container's /dev directory.
+// The char-device check uses the cheap readdir type; the minor number is
+// resolved for char-device entries.
+func listDeviceNodes(containerRoot *os.Root) ([]deviceNode, error) {
+	devDir, err := containerRoot.Open("dev")
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to open /dev in container: %w", err)
+	}
+	defer devDir.Close()
+
+	var nodes []deviceNode
+	for {
+		entries, readErr := devDir.ReadDir(readDirBatchSize)
+		for _, entry := range entries {
+			node := deviceNode{
+				name:         entry.Name(),
+				isCharDevice: entry.Type()&os.ModeCharDevice != 0,
+			}
+			if node.isCharDevice {
+				minor, err := deviceNodeMinor(containerRoot, entry.Name())
+				if err != nil {
+					return nil, err
+				}
+				node.minor = minor
+			}
+			nodes = append(nodes, node)
+		}
+		if readErr == io.EOF {
+			break
+		}
+		if readErr != nil {
+			return nil, fmt.Errorf("failed to read /dev in container: %w", readErr)
+		}
+	}
+
+	return nodes, nil
+}
+
+// deviceNodeMinor returns the device minor number for the named entry under the
+// container's /dev directory.
+func deviceNodeMinor(containerRoot *os.Root, name string) (int, error) {
+	info, err := containerRoot.Lstat(filepath.Join("dev", name))
+	if err != nil {
+		return 0, fmt.Errorf("failed to stat %s: %w", name, err)
+	}
+	stat, ok := info.Sys().(*syscall.Stat_t)
+	if !ok {
+		return 0, fmt.Errorf("unexpected stat type for %s", name)
+	}
+	return int(unix.Minor(stat.Rdev)), nil
+}
--- a/cmd/nvidia-cdi-hook/update-application-profile/update-application-profile.go
+++ b/cmd/nvidia-cdi-hook/update-application-profile/update-application-profile.go
@@ -1,0 +1,145 @@
+/**
+# SPDX-FileCopyrightText: Copyright (c) NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package updateapplicationprofile
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"regexp"
+
+	"github.com/urfave/cli/v3"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/oci"
+)
+
+const (
+	applicationProfileDir  = "etc/nvidia/nvidia-application-profiles-rc.d"
+	applicationProfileFile = applicationProfileDir + "/10-container.conf"
+)
+
+// nvidiaGPUDeviceNode matches the device node names for physical NVIDIA GPUs,
+// e.g. nvidia0, nvidia12.
+var nvidiaGPUDeviceNode = regexp.MustCompile(`^nvidia[0-9]+$`)
+
+// A deviceNode describes a single entry discovered under the container's /dev
+// directory. minor is only meaningful when isCharDevice is true.
+type deviceNode struct {
+	name         string
+	isCharDevice bool
+	minor        int
+}
+
+// selectGPUMinors returns the minor numbers of the entries that are physical
+// NVIDIA GPU device nodes (/dev/nvidiaN char devices).
+func selectGPUMinors(nodes []deviceNode) []int {
+	var minors []int
+	for _, n := range nodes {
+		if n.isCharDevice && nvidiaGPUDeviceNode.MatchString(n.name) {
+			minors = append(minors, n.minor)
+		}
+	}
+	return minors
+}
+
+type options struct {
+	containerSpec string
+}
+
+// NewCommand constructs an update-application-profile subcommand with the specified logger
+func NewCommand(logger logger.Interface) *cli.Command {
+	cfg := options{}
+
+	c := cli.Command{
+		Name:  "update-application-profile",
+		Usage: `Updates driver settings through "application profiles". Currently, this hook sets EGLVisibleDGPUDevices to restrict EGL/Vulkan GPU visibility inside the container`,
+		Action: func(ctx context.Context, cmd *cli.Command) error {
+			return run(ctx, cmd, &cfg, logger)
+		},
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:        "container-spec",
+				Hidden:      true,
+				Usage:       "Specify the path to the OCI container spec. If empty or '-' the spec will be read from STDIN",
+				Destination: &cfg.containerSpec,
+			},
+		},
+	}
+
+	return &c
+}
+
+func run(_ context.Context, _ *cli.Command, cfg *options, logger logger.Interface) error {
+	s, err := oci.LoadContainerState(cfg.containerSpec)
+	if err != nil {
+		return fmt.Errorf("failed to load container state: %w", err)
+	}
+
+	containerRootDirPath, err := s.GetContainerRoot()
+	if err != nil {
+		return fmt.Errorf("failed to determine container root: %w", err)
+	}
+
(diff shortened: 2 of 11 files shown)


cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
```diff
@@ -0,0 +1,88 @@
...
+				}
+				node.minor = minor
+			}
+			nodes = append(nodes, node)
+		}
+		if readErr == io.EOF {
+			break
+		}
+		if readErr != nil {
+			return nil, fmt.Errorf("failed to read /dev in container: %w", readErr)
```
tariq1890, line 69: can we move this up so that it is executed immediately after the ReadDir call?

cmd/nvidia-cdi-hook/update-app-profile/minors_linux.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+// readDirBatchSize is the number of directory entries read from the
+// container's /dev directory per ReadDir call.
+const readDirBatchSize = 100
+
+// deviceNodesInDev enumerates the entries under the container's /dev directory.
+// The char-device check uses the cheap readdir type; the minor number is
+// resolved for char-device entries.
+func deviceNodesInDev(containerRoot *os.Root) ([]deviceNode, error) {
```
tariq1890: how about `enumerateLinuxDeviceNodes?` or `listLinuxDeviceNodes` ?

cmd/nvidia-cdi-hook/update-application-profile/update-application-profile.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	minors := selectGPUMinors(nodes)
+	if len(minors) == 0 {
+		return nil
+	}
+
+	var mask uint64
+	for _, minor := range minors {
+		mask |= uint64(1) << uint(minor)
```
tariq1890, line 121: This assumes that minor numbers will always be a number between 0 and 63. While this is a fair assumption to make, can we have the code throw an error if the number is greater than or equal to 64?

cmd/nvidia-cdi-hook/update-app-profile/minors_linux.go
henry118: updated func name

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
henry118: in case of error, `ReadDir` returns the entries it was able to read before the error, **along with** the error. The current logic is correct.

cmd/nvidia-cdi-hook/update-application-profile/update-application-profile.go
henry118: Returning an error would cause container start failure. But I agree that silently dropping the bit is also not ideal (although `libnvidia-container` does that today). 

I've added a warning message when such case is detected. I think that would suffice.

cmd/nvidia-cdi-hook/update-app-profile/update-app-profile.go
```diff
@@ -0,0 +1,152 @@
...
+
+func run(_ context.Context, _ *cli.Command, cfg *options, logger logger.Interface) error {
+	s, err := oci.LoadContainerState(cfg.containerSpec)
+	if err != nil {
+		return fmt.Errorf("failed to load container state: %w", err)
+	}
+
+	containerRootDirPath, err := s.GetContainerRoot()
+	if err != nil {
+		return fmt.Errorf("failed to determined container root: %w", err)
```
tariq1890:
```suggestion
		return fmt.Errorf("failed to determine container root: %w", err)
```

internal/discover/hooks.go
tariq1890: can we mention `UpdateAppProfile` here?

cmd/nvidia-cdi-hook/update-app-profile/update-app-profile.go
```diff
@@ -0,0 +1,152 @@
...
+				Usage:       "Specify the path to the OCI container spec. If empty or '-' the spec will be read from STDIN",
+				Destination: &cfg.containerSpec,
+			},
+		},
+	}
+
+	return &c
+}
+
+func validateFlags(_ *cli.Command, _ *options) error {
```
tariq1890: I know that this has been followed in other CDI hook files, but we should probably just remove this since it's code doesn't do anything

internal/discover/app_profile.go
```diff
@@ -0,0 +1,26 @@
...
+**/
+
+package discover
+
+import (
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+)
+
+// NewAppProfileHookDiscoverer creates a discoverer for the update-app-profile hook.
+func NewAppProfileHookDiscoverer(logger logger.Interface, hookCreator HookCreator) Discover {
```
tariq1890:
```suggestion
func NewAppProfileHookDiscoverer(hookCreator HookCreator) Discover {
```

cmd/nvidia-cdi-hook/update-app-profile/update-app-profile.go
```diff
@@ -0,0 +1,152 @@
...
+			continue
+		}
+		mask |= uint64(1) << uint(minor)
+	}
+
+	if mask == 0 {
+		return nil
+	}
+
+	if err := containerRoot.MkdirAll(appProfileDir, 0555); err != nil {
```
tariq1890: any specific reason why it's `555` here? Why not `755` like we do everywhere else?

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to open /dev in container: %w", err)
+	}
+	defer devDir.Close()
+
+	var nodes []deviceNode
+	for {
+		entries, readErr := devDir.ReadDir(readDirBatchSize)
```
tariq1890, line 50:
```suggestion
		entries, readErr := devDir.ReadDir(readDirBatchSize)
		if readErr != nil && readErr != io.EOF {
		    return nil, fmt.Errorf("failed to read /dev in container: %w", readErr)
		}
```

tariq1890: Why not return early here if we know it's a non-EOF error?

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
```diff
@@ -0,0 +1,88 @@
...
+	"os"
+	"path"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+)
+
+// readDirBatchSize is the number of directory entries read from the
+// container's /dev directory per ReadDir call.
+const readDirBatchSize = 100
```
tariq1890, line 33: I ask out of curiosity - any particular reason for choosing `100` as the batch size?

cmd/nvidia-cdi-hook/update-app-profile/minors_linux.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+
+	return nodes, nil
+}
+
+// deviceNodeMinor returns the device minor number for the named entry under the
+// container's /dev directory.
+func deviceNodeMinor(containerRoot *os.Root, name string) (int, error) {
+	info, err := containerRoot.Lstat(path.Join("dev", name))
```
tariq1890: The convention we follow in the toolkit code in `filepath.Join()`, so let's use the `path/filepath` stdlib import here.

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
tariq1890: Thanks, this is good to know.

cmd/nvidia-cdi-hook/update-app-profile/update-app-profile.go
henry118: it's just mirroring the behavior of `libnvidia-container`:
https://github.com/NVIDIA/libnvidia-container/blob/5f3d3abb381e4cc78cb29f94d21ede845cb3b374/src/nvc_mount.c#L286

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
henry118: I don't think this is necessary. Any particular reason you feel strongly about this change?

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
henry118: No particular reason. just a random chosen number which is not too big / too small

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
tariq1890: - Handling an error after the method call is the normalised practice. So, I find this more readable and intuitive
- Currently, we loop through the entries, build the `nodes` list and then return `nil` if an error other than `io.EOF` is thrown. With this suggestion, we short-circuit it by returning the `nil` and error in case of a non-EOF error

cmd/nvidia-cdi-hook/update-app-profile/minors_linux.go
```diff
@@ -0,0 +1,88 @@
...
+// The char-device check uses the cheap readdir type; the minor number is
+// resolved for char-device entries.
+func listDeviceNodes(containerRoot *os.Root) ([]deviceNode, error) {
+	devDir, err := containerRoot.Open("dev")
+	if errors.Is(err, os.ErrNotExist) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to open /dev in container: %w", err)
+	}
```
tariq1890: What are your thoughts on this?
```suggestion
	if err != nil {
		if errors.Is(err, os.ErrNotExist) {
		    return nil, nil
		}
		return nil, fmt.Errorf("failed to open /dev in container: %w", err)
	}
```

tariq1890: This is a nit comment, I don't feel too strongly about this. Just curious what style you'd prefer here

tariq1890 (approved):

tariq1890 (approved):
Thanks @henry118 !

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
```diff
@@ -0,0 +1,88 @@
+/**
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cdesiniotis: let's remove the year here (and throughout other files in this PR).
```suggestion
# SPDX-FileCopyrightText: Copyright (c) NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+// readDirBatchSize is the number of directory entries read from the
+// container's /dev directory per ReadDir call.
+const readDirBatchSize = 100
+
+// listDeviceNodes enumerates the entries under the container's /dev directory.
+// The char-device check uses the cheap readdir type; the minor number is
+// resolved for char-device entries.
+func listDeviceNodes(containerRoot *os.Root) ([]deviceNode, error) {
```
cdesiniotis, line 38: Potentially for a follow-up -- There may be an opportunity for code reuse here. We are essentially globbing for files in a root. We do this as well in the `cudacompat` hook, see https://github.com/NVIDIA/nvidia-container-toolkit/blob/1cddfb0dc179136cd720090f0a13e6ce0de611ed/cmd/nvidia-cdi-hook/cudacompat/container-root_linux.go#L101-L138. We may benefit from a `internal/root` package that can be reused by all of our hooks. It should provide the ability to perform file operations "safely" in a root fs. 

I was going to suggest we reuse the `CharDeviceLocator`, from https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/pkg/lookup/device.go, but this code operates on file paths and therefore is not suitable for a hook where we don't trust the root fs we are operating on.

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
henry118: Updated.

cmd/nvidia-cdi-hook/update-application-profile/minors_linux.go
henry118: Good suggestion. Let me create another PR for this refactor work to limit its scope.

cdesiniotis (approved):

PR #1942, opened 2026-07-14
fix: disable chmod hook

rajatchopra:
Chmod hook was deprecated earlier. We can disable it now. Any code requiring correct access to /dev/dri does not need this backward compatible hook as crun has a fix now. For more details, see #1221

Existing CDI specs invoking chmod will now use the unsupported-hook handler: they emit a warning, perform no chmod, and do not prevent container startup.

### DIFF
--- a/cmd/nvidia-cdi-hook/commands/commands_test.go
+++ b/cmd/nvidia-cdi-hook/commands/commands_test.go
@@ -1,0 +1,33 @@
+/**
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package commands
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/urfave/cli/v3"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+)
+
+func TestConfigureCDIHookCommandDisablesChmod(t *testing.T) {
+	command := ConfigureCDIHookCommand(&logger.NullLogger{}, &cli.Command{})
+
+	require.Nil(t, command.Command("chmod"))
+	require.NotNil(t, command.Command("create-symlinks"))
+}
--- a/cmd/nvidia-ctk/cdi/generate/generate_test.go
+++ b/cmd/nvidia-ctk/cdi/generate/generate_test.go
@@ -388,7 +388,7 @@ containerEdits:
 `,
 		},
 		{
-			description: "enableChmodHook",
+			description: "cannotEnableChmodHook",
 			options: options{
 				format:        "yaml",
 				mode:          "management",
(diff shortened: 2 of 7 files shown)


internal/discover/hooks.go
cdesiniotis: Shouldn't we remove all references to the `chmod` hook as part of this commit?

internal/discover/hooks.go
```diff
@@ -63,10 +63,9 @@ const (
...
 )
 
-// defaultDisabledHooks defines hooks that are disabled by default.
-// These hooks can be explicitly enabled using the WithEnabledHooks option.
+// defaultDisabledHooks defines hooks that cannot be enabled.
 var defaultDisabledHooks = []HookName{
-	// ChmodHook is disabled by default as it was a workaround for older
+	// ChmodHook is disabled as it was a workaround for older
 	// versions of crun that has since been fixed.
 	ChmodHook,
```
cdesiniotis: Why do we need to list the chmod hook here if it is not supported / is a no-op anyways?

cmd/nvidia-ctk/cdi/generate/generate_test.go
```diff
@@ -388,7 +388,7 @@ containerEdits:
 `,
 		},
 		{
-			description: "enableChmodHook",
+			description: "cannotEnableChmodHook",
```
cdesiniotis: nit: `chmodIsNoLongerASupportedHook`

cmd/nvidia-cdi-hook/commands/commands_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/urfave/cli/v3"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+)
+
+func TestConfigureCDIHookCommandDisablesChmod(t *testing.T) {
+	command := ConfigureCDIHookCommand(&logger.NullLogger{}, &cli.Command{})
+
+	require.Nil(t, command.Command("chmod"))
+	require.NotNil(t, command.Command("create-symlinks"))
```
cdesiniotis: Question -- Is this returning an error because the `create-symlinks` hook fails in this context (e.g. the right set of arguments are not provided)?

internal/discover/hooks.go
```diff
@@ -185,6 +185,9 @@ func NewHookCreator(opts ...Option) HookCreator {
 	}
 
 	for _, h := range o.enabledHooks {
+		if h == ChmodHook {
+			continue
+		}
```
cdesiniotis: Since the chmod hook is a no-op now, I would prefer if we did not special case for it. I would be in favor of completely removing all references to it instead (aka code removal following its deprecation since the v1.18.0 relesae).

cdesiniotis (commented):
Since this hook has been deprecated and off by default (since the 1.18.0 release, see [commit](https://github.com/NVIDIA/nvidia-container-toolkit/commit/c350d1335507f86304b8e5daef3c5391386abe12)), I think the next step would be to remove it entirely (instead of adding extra logic to prevent one from enabling it).

internal/discover/hooks_test.go
```diff
@@ -71,29 +67,9 @@ func TestNewHookCreator(t *testing.T) {
 				disabledHooks: map[HookName]bool{
 					AllHooks:          true,
 					UpdateLDCacheHook: false,
-					ChmodHook:         true,
 				},
 			},
-		},
-		{
-			name: "multiple hooks disabled and enabled",
```
cdesiniotis, line 79: Can we keep this test case, but just remove the ChmodHook?

internal/discover/hooks_test.go
```diff
@@ -349,24 +284,9 @@ func TestCDIHookCreator_isDisabled(t *testing.T) {
...
-			expectedResult: true, // ChmodHook is disabled by default and explicitly disabled
-		},
-		{
-			name:           "hook explicitly enabled overrides disabled",
-			disabledHooks:  []HookName{ChmodHook},
-			enabledHooks:   []HookName{ChmodHook},
-			hookName:       ChmodHook,
-			args:           []string{"/path/to/file"},
-			expectedResult: false,
-		},
```
cdesiniotis, line 366: Instead of removing these test cases, can we keep them but replace `ChmodHook` with a valid hook please?

internal/discover/hooks.go
rajatchopra: removed the code

internal/discover/hooks.go
rajatchopra: Done

internal/discover/hooks_test.go
rajatchopra: Done.

internal/discover/hooks_test.go
rajatchopra: Done.

pkg/nvcdi/management.go
```diff
@@ -108,13 +108,8 @@ func (l *managementlib) newManagementDeviceDiscoverer() (discover.Discover, erro
 		},
 	)
 
-	deviceFolderPermissionHooks := (*nvcdilib)(l).newDeviceFolderPermissionHookDiscoverer(
-		deviceNodes,
-	)
-
 	d := discover.Merge(
```
henry118: nit: the merge here is not needed given only one element remaining after the change.

rajatchopra: fixed.

henry118 (approved):

PR #1947, opened 2026-07-21
Enable JIT-CDI to inject MIG management capability devices

### LINKED ISSUE #1740: [Bug]: Non-privileged `MIG_STRATEGY=mixed` fails due to missing device-cgroup access for `nvidia-cap1`/`nvidia-cap2` (nvidia-container-toolkit 1.18+ / 1.19.0)
**Describe the bug**
 with `MIG_STRATEGY=mixed` on `nvidia-container-toolkit` 1.18+ (also reproducible on 1.19.0).

`MIG_STRATEGY=mixed` fails in a non-privileged NVIDIA device-plugin pod with:

```text
error getting device memory info: Insufficient Permissions
```

Using `privileged: true` mitigates the issue, but the NVIDIA device-plugin is typically deployed non-privileged and privileged mode is not documented as a requirement for mixed MIG discovery.

**To Reproduce**
## Repro Matrix

1. Non-privileged + `MIG_STRATEGY=mixed` => fails
2. Privileged + `MIG_STRATEGY=mixed` => succeeds, MIG resources register
3. Non-privileged + `MIG_STRATEGY=none` => succeeds
4. Non-privileged + container-toolkit mode = "legacy" => succeeds
5. Non-privileged + 1.17.x container-toolkit = => succeeds

## Observed Host State

`/dev/nvidia-caps` exists on the node and includes `cap1` and `cap2`:

```text
# ls -l /dev/nvidia-caps/
total 0
cr-------- 1 root root 237,   1 Mar 17 17:55 nvidia-cap1
cr--r--r-- 1 root root 237, 102 Mar 17 17:55 nvidia-cap102
cr--r--r-- 1 root root 237, 103 Mar 17 17:55 nvidia-cap103
cr--r--r-- 1 root root 237, 111 Mar 17 17:55 nvidia-cap111
cr--r--r-- 1 root root 237, 112 Mar 17 17:55 nvidia-cap112
cr--r--r-- 1 root root 237, 120 Mar 17 17:55 nvidia-cap120
cr--r--r-- 1 root root 237, 121 Mar 17 17:55 nvidia-cap121
cr--r--r-- 1 root root 237,   2 Mar 17 17:55 nvidia-cap2
cr--r--r-- 1 root root 237,  66 Mar 17 17:55 nvidia-cap66
cr--r--r-- 1 root root 237,  67 Mar 17 17:55 nvidia-cap67
cr--r--r-- 1 root root 237,  75 Mar 17 17:55 nvidia-cap75
cr--r--r-- 1 root root 237,  76 Mar 17 17:55 nvidia-cap76
cr--r--r-- 1 root root 237,  84 Mar 17 17:55 nvidia-cap84
cr--r--r-- 1 root root 237,  85 Mar 17 17:55 nvidia-cap85
cr--r--r-- 1 root root 237,  93 Mar 17 17:55 nvidia-cap93
cr--r--r-- 1 root root 237,  94 Mar 17 17:55 nvidia-cap94
```

But in the non-privileged pod, device-cgroup access appears not granted for `cap1`/`cap2` (EPERM in `devcgroup_check_permission`), while privileged pod bypasses those restrictions and works.

## CDI State

`nvidia-ctk cdi list` on node shows only 15 devices and does not include MIG management classes:

```text
# nvidia-ctk cdi list
INFO[0000] Found 15 CDI devices
nvidia.com/gpu=0:0
nvidia.com/gpu=0:1
nvidia.com/gpu=0:2
nvidia.com/gpu=0:3
nvidia.com/gpu=0:4
nvidia.com/gpu=0:5
nvidia.com/gpu=0:6
nvidia.com/gpu=MIG-07ad8e7c-0197-5973-bd05-7fcd18ae754e
nvidia.com/gpu=MIG-30171940-9037-5504-89d2-7b5acd71bf5d
nvidia.com/gpu=MIG-90825510-8f47-5052-8e15-ead851500769
nvidia.com/gpu=MIG-90f01f31-b6fe-54f6-acc8-0e228a8ca6e2
nvidia.com/gpu=MIG-9207551c-4419-5285-9621-19e99aec2afc
nvidia.com/gpu=MIG-93b1626d-9793-506e-81de-1b00f8f9edfc
nvidia.com/gpu=MIG-acec7971-611e-54f5-ae21-4a0f5a5eb074
nvidia.com/gpu=all
```

No:

- `nvidia.com/mig-config=all`
- `nvidia.com/mig-monitor=all`

## Device Plugin Error

```text
error starting plugins: unable to add default resources to config: error visiting devices: error visiting device: error visiting device: error getting device memory info: Insufficient Permissions
```

Device-plugin pod (non-privileged):

```yaml
  env:
    - name: MPS_ROOT
      value: /run/nvidia/mps
    - name: MIG_STRATEGY
      value: mixed
    - name: NVIDIA_VISIBLE_DEVICES
      value: all
    - name: NVIDIA_DRIVER_CAPABILITIES
      value: compute,utility
    - name: NVIDIA_MIG_MONITOR_DEVICES
      value: all
    image: nvcr.io/nvidia/k8s-device-plugin:v0.18.0
    imagePullPolicy: IfNotPresent
    name: nvidia-device-plugin-ctr
    resources: {}
    securityContext:
      capabilities:
        add:
        - SYS_ADMIN
      privileged: false
      runAsGroup: 0
      runAsUser: 0
```
 (this config works with 1.17.x nvidia-container-toolkit, but not with any 1.18.x)

**Expected behavior**

device-plugin pod shouldn't need to be granted with "privileged=true" , for non-privileged `MIG_STRATEGY=mixed`, what is the officially supported configuration?

From NVIDIA MIG docs https://docs.nvidia.com/datacenter/tesla/mig-user-guide/device-nodes-and-capabilities.html, `mig-config` maps to `nvidia-cap1` and requires cgroup-granted access in non-privileged containers. In our case, node device nodes exist, but non-privileged container cgroup permissions for cap1/cap2 are missing.

It appears default CDI refresh only gives `nvidia.com/gpu=*` entries and does not expose MIG management CDI classes that would inject corresponding device nodes and cgroup rules into non-privileged management containers.

**Environment**
- nvidia-container-toolkit version: 1.19.0-1 (nvidia-ctk --version: 1.19.0, commit ec7b4e2fa2caecad6d89be4a26029b831fe7503a)
  - NVIDIA Driver Version: 580.126.09
  - Host OS: Ubuntu 22.04.5 LTS
  - Kernel Version: 5.15.0-1102-azure
  - Container Runtime Version: containerd 1.7.30-2 (containerd://1.7.30-2)
  - CPU Architecture: x86_64 (amd64 in Kubernetes node info)
  - GPU Model(s): NVIDIA H100 NVL

**Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/)** (optional if deemed irrelevant)

Question

## Mitigation with non-privileged device-plugin pod  (Requesting Confirmation) 

1. Runtime config:

```bash
nvidia-ctk config --config-file=/etc/nvidia-container-runtime/config.toml --in-place \
  --set accept-nvidia-visible-devices-as-volume-mounts=false \
  --set accept-nvidia-visible-devices-envvar-when-unprivileged=true
```

2. Generate management CDI specs:

```bash
nvidia-ctk cdi generate --mode=management --vendor=nvidia.com --class=mig-config \
  --output=/var/run/cdi/nvidia-mig-config.yaml
nvidia-ctk cdi generate --mode=management --vendor=nvidia.com --class=mig-monitor \
  --output=/var/run/cdi/nvidia-mig-monitor.yaml
```

henry118:
Fixes: #1740 

Same as legacy behavior, MIG config/monitor devices requires "privileged" container (`SYS_ADMIN` included in the bounding set).

#### w/ SYS_ADMIN

```
$ docker run --rm -it --runtime nvidia --cap-add=SYS_ADMIN --gpus=all -e NVIDIA_MIG_CONFIG_DEVICES=all -e NVIDIA_MIG_MONITOR_DEVICES=all ubuntu ls -l /dev/nvidia-caps
total 0
cr-------- 1 root root 236, 1 Sep  1 21:10 nvidia-cap1
cr--r--r-- 1 root root 236, 2 Sep  1 21:10 nvidia-cap2
```

#### w/o SYS_ADMIN

```
$ docker run --rm -it --runtime nvidia --gpus=all -e NVIDIA_MIG_CONFIG_DEVICES=all -e NVIDIA_MIG_MONITOR_DEVICES=all ubuntu ls -l /dev/nvidia-caps
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: could not apply required modification to OCI specification: error modifying OCI spec: failed to create the automatic CDI modifier: invalid device request: cannot set NVIDIA_MIG_CONFIG_DEVICES or NVIDIA_MIG_MONITOR_DEVICES in a non-privileged container
```

### DIFF
--- a/internal/modifier/cdi.go
+++ b/internal/modifier/cdi.go
@@ -18,8 +18,10 @@ package modifier
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
+	"github.com/NVIDIA/go-nvlib/pkg/nvlib/device"
 	"tags.cncf.io/container-device-interface/pkg/parser"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/config/image"
@@ -82,6 +84,7 @@ func (f *Factory) newJitCDIModifier(automaticDevices []string) (oci.SpecModifier
 	if f.image != nil {
 		automaticDevices = append(automaticDevices, withUniqueDevices(gatedDevices(*f.image)).DeviceRequests()...)
 		automaticDevices = append(automaticDevices, withUniqueDevices(imexDevices(*f.image)).DeviceRequests()...)
+		automaticDevices = append(automaticDevices, withUniqueDevices(migCapsDevices(*f.image)).DeviceRequests()...)
 	}
 	return f.newAutomaticCDISpecModifier(automaticDevices)
 }
@@ -155,6 +158,51 @@ func (d imexDevices) DeviceRequests() []string {
 		devices = append(devices, "mode=imex,id="+channelID)
 	}
 	return devices
+}
+
+type migCapsDevices image.CUDA
+
+// DeviceRequests returns the MIG management capabilities requested through the
+// NVIDIA_MIG_CONFIG_DEVICES and NVIDIA_MIG_MONITOR_DEVICES envvars.
+//
+// Only the value "all" is meaningful here: it maps to the global MIG config /
+// monitor capability nodes (nvidia-cap<config>, nvidia-cap<monitor>).
+func (d migCapsDevices) DeviceRequests() []string {
+	i := (image.CUDA)(d)
+
+	var devices []string
+	if strings.EqualFold(i.Getenv(image.EnvVarNvidiaMigConfigDevices), "all") {
+		devices = append(devices, "mode=mig-caps,id=config")
+	}
+	if strings.EqualFold(i.Getenv(image.EnvVarNvidiaMigMonitorDevices), "all") {
+		devices = append(devices, "mode=mig-caps,id=monitor")
+	}
+	if len(devices) == 0 {
+		return nil
+	}
+	// The global MIG config/monitor capability grants GPU-wide access and would
+	// expose sibling MIG instances on the same GPU. Do not grant it to a container
+	// that is scoped to specific MIG devices.
+	if requestsSpecificMigDevices(i) {
+		return nil
+	}
+	return devices
+}
+
+// requestsSpecificMigDevices reports whether the visible device list selects one
+// or more specific MIG devices rather than whole GPUs or "all".
+func requestsSpecificMigDevices(i image.CUDA) bool {
+	visibleDevices := i.VisibleDevices()
+	if slices.Contains(visibleDevices, "all") {
+		return false
+	}
+	for _, d := range visibleDevices {
+		id := device.Identifier(d)
+		if id.IsMigUUID() || id.IsMigIndex() {
+			return true
+		}
+	}
+	return false
 }
 
 // filterAutomaticDevices searches for "automatic" device names in the input slice.
--- a/internal/modifier/cdi.go
+++ b/internal/modifier/cdi.go
@@ -18,8 +18,10 @@ package modifier
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
+	"github.com/NVIDIA/go-nvlib/pkg/nvlib/device"
 	"tags.cncf.io/container-device-interface/pkg/parser"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/config/image"
@@ -82,6 +84,12 @@ func (f *Factory) newJitCDIModifier(automaticDevices []string) (oci.SpecModifier
 	if f.image != nil {
 		automaticDevices = append(automaticDevices, withUniqueDevices(gatedDevices(*f.image)).DeviceRequests()...)
 		automaticDevices = append(automaticDevices, withUniqueDevices(imexDevices(*f.image)).DeviceRequests()...)
+
+		migCaps := migCapsDevices(*f.image)
+		if err := migCaps.Validate(); err != nil {
+			return nil, fmt.Errorf("invalid MIG capability request: %w", err)
+		}
+		automaticDevices = append(automaticDevices, withUniqueDevices(migCaps).DeviceRequests()...)
 	}
 	return f.newAutomaticCDISpecModifier(automaticDevices)
 }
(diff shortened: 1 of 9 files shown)


internal/modifier/cdi.go
```diff
@@ -157,6 +160,51 @@ func (d imexDevices) DeviceRequests() []string {
...
+	if strings.EqualFold(i.Getenv(image.EnvVarNvidiaMigMonitorDevices), "all") {
+		devices = append(devices, "mode=mig-caps,id=monitor")
+	}
+	if len(devices) == 0 {
+		return nil
+	}
+	// The global MIG config/monitor capability grants GPU-wide access and would
+	// expose sibling MIG instances on the same GPU. Do not grant it to a container
+	// that is scoped to specific MIG devices.
+	if requestsSpecificMigDevices(i) {
```
cdesiniotis, line 189: Question -- is this existing behavior from the legacy toolkit?

pkg/nvcdi/lib-mig-caps.go
```diff
@@ -0,0 +1,100 @@
...
+	deviceNodes := discover.NewCharDeviceDiscoverer(
+		l.logger,
+		l.driver.DevRoot,
+		[]string{devicePath},
+	)
+
+	// The MIG capability nodes are nested under /dev/nvidia-caps. Add a hook to
+	// set the permissions of the parent folder so that non-root users in the
+	// container can access the injected node.
+	folderPermissionHooks := (*nvcdilib)(l).newDeviceFolderPermissionHookDiscoverer(deviceNodes)
```
cdesiniotis: Note, the implementation in the legacy `nvidia-container-runtime-hook` would only grant access to the `config` / `monitor` MIG caps if the container process had `CAP_SYS_ADMIN`, see https://github.com/NVIDIA/nvidia-container-toolkit/blob/36e84731a032badf60e91aff32fad76225a3d897/cmd/nvidia-container-runtime-hook/container_config.go#L190-L204 and [external document](https://docs.google.com/document/d/13LyeFxuWLSwDjCXKx7RyWWRZEAqA-mghtK3GRPvGv1s/edit). First, let's align with the legacy implementation. Second, if the container process has `CAP_SYS_ADMIN`, is this extra hook still required?

internal/modifier/cdi.go
henry118: Yes - here is the logic

https://github.com/NVIDIA/libnvidia-container/blob/70f648f8816a77093f033eee67afe655f74abe55/src/cli/common.c#L356-L359

cdesiniotis: Ack, thanks!

cdesiniotis: I see the legacy implementation returns an error in this case. Should we consider doing the same?

henry118: yes i'll update

pkg/nvcdi/lib-mig-caps.go
henry118: yeah, i missed that constraint. will address in the next revision

henry118:
@cdesiniotis I just pushed a new revision which should have addressed your comments. PTAL when you get a chance. Thanks!

internal/modifier/cdi.go
```diff
@@ -78,10 +80,17 @@ func (f *Factory) newCDIModifier(isJitCDI bool) (oci.SpecModifier, error) {
 }
 
 // newJitCDIModifier creates a modifier that for a generated in-memory CDI spec for the specified CDI devices.
-func (f *Factory) newJitCDIModifier(automaticDevices []string) (oci.SpecModifier, error) {
+func (f *Factory) newJitCDIModifier(automaticDevices []string) (modifier oci.SpecModifier, err error) {
+	defer func() {
+		if r := recover(); r != nil {
```
tariq1890: curious: why do we anticipate a `panic`? Instead of recovering from a panic, shouldn't we have the origin throw an error instead since this is an expected failure path?

internal/modifier/cdi.go
```diff
@@ -157,6 +166,59 @@ func (d imexDevices) DeviceRequests() []string {
...
+
+	migConfig := i.Getenv(image.EnvVarNvidiaMigConfigDevices)
+	migMonitor := i.Getenv(image.EnvVarNvidiaMigMonitorDevices)
+
+	if migConfig == "" && migMonitor == "" {
+		return nil
+	}
+	// Only the value "all" is currently supported for either envvar.
+	if migConfig != "" && !strings.EqualFold(migConfig, "all") {
+		panic(fmt.Sprintf("invalid NVIDIA_MIG_CONFIG_DEVICES %q: only \"all\" is supported", migConfig))
```
tariq1890: I am guessing there is a specific reason why `panic`s were chosen in this method. My gut reaction is that we should just be throwing an error.

See [this](https://go.dev/wiki/CodeReviewComments#dont-panic)

internal/modifier/cdi.go
```diff
@@ -157,6 +166,59 @@ func (d imexDevices) DeviceRequests() []string {
 	return devices
 }
 
+type migCapsDevices image.CUDA
+
+// DeviceRequests returns the MIG management capabilities requested through the
+// NVIDIA_MIG_CONFIG_DEVICES and NVIDIA_MIG_MONITOR_DEVICES envvars.
+func (d migCapsDevices) DeviceRequests() []string {
```
tariq1890: If we are panic'ing because of wanting to work with this method signature, can consider adding a separate validation stage (which will throw an error) prior to calling `DeviceRequests`?

henry118: Good idea. Added `Validate` method

cdesiniotis (approved):

tariq1890 (approved):

PR #1578, opened 2026-01-14
internal/discover: use explicit driver version for matching libraries

Ronitsabhaya75:
Previous Behavior
The discovery logic used a generic *.* wildcard pattern to identify driver libraries. This was less precise and could potentially match unrelated files or incorrect versions.
```
pattern := strings.TrimSuffix(libraryName, ".") + ".*.*"
```
Current Behavior
The update modifies  graphicsDriverLibraries to explicitly use the detected driver version when constructing the match pattern. This ensures that only libraries matching the active driver version are processed, preventing false positives and ensuring consistency.

```
pattern := strings.TrimSuffix(libraryName, ".") + "." + d.driverVersion
```

PR #1948, opened 2026-07-22
use the driver version explicitly to match libraries

tariq1890:
This is intended to replace the PR https://github.com/NVIDIA/nvidia-container-toolkit/pull/1578. 

It addresses some of the minor review comments and adds tests. The original author has been attributed and added as co-author.

> Previous Behavior
The discovery logic used a generic *.* wildcard pattern to identify driver libraries. This was less precise and could potentially match unrelated files or incorrect versions.
```
 pattern := strings.TrimSuffix(libraryName, ".") + ".*.*"
```
> Current Behavior
The update modifies  graphicsDriverLibraries to explicitly use the detected driver version when constructing the match pattern. This ensures that only libraries matching the active driver version are processed, preventing false positives and ensuring consistency.

```
pattern := strings.TrimSuffix(libraryName, ".") + "." + d.driverVersion
```

### DIFF
--- a/internal/discover/graphics.go
+++ b/internal/discover/graphics.go
@@ -111,14 +111,17 @@ func newVulkanConfigsDiscover(logger logger.Interface, driver *root.Driver) Disc
 
 type graphicsDriverLibraries struct {
 	Discover
-	logger      logger.Interface
-	hookCreator HookCreator
+	logger        logger.Interface
+	hookCreator   HookCreator
+	driverVersion string
 }
 
 var _ Discover = (*graphicsDriverLibraries)(nil)
 
 func newGraphicsLibrariesDiscoverer(logger logger.Interface, driver *root.Driver, hookCreator HookCreator) (Discover, error) {
-	cudaVersionPattern, err := driver.Version()
+	// We use the driver version as a suffix for matching libraries that are
+	// part of the driver.
+	driverVersion, err := driver.Version()
 	if err != nil {
 		return nil, fmt.Errorf("failed to get driver version: %w", err)
 	}
@@ -143,8 +146,8 @@ func newGraphicsLibrariesDiscoverer(logger logger.Interface, driver *root.Driver
 			// * libnvidia-allocator.so.RM_VERSION
 			// * libnvidia-vulkan-producer.so.RM_VERSION
 			// but need to be handled for the legacy case too.
-			"libnvidia-allocator.so." + cudaVersionPattern,
-			"libnvidia-vulkan-producer.so." + cudaVersionPattern,
+			"libnvidia-allocator.so." + driverVersion,
+			"libnvidia-vulkan-producer.so." + driverVersion,
 		},
 	)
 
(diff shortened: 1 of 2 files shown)


internal/discover/graphics_test.go
```diff
@@ -108,6 +108,25 @@ func TestGraphicsLibrariesDiscoverer(t *testing.T) {
...
+							Path: "/usr/lib64/libnvidia-allocator.so.999.99.99",
+						},
+					}
+					return mounts, nil
+				},
+			},
+			expectedMounts: []Mount{
+				{
+					Path: "/usr/lib64/libnvidia-allocator.so.999.99.99",
+				},
```
cdesiniotis, line 126: Why is this mount expected? `999.99.99` does not equal the driver version `123.45.67` configured for the test.

tariq1890: Fair point

internal/discover/graphics.go
```diff
@@ -234,10 +238,7 @@ func (d graphicsDriverLibraries) Hooks() ([]Hook, error) {
 
 // isDriverLibrary checks whether the specified filename is a specific driver library.
 func (d graphicsDriverLibraries) isDriverLibrary(filename string, libraryName string) bool {
-	// TODO: Instead of `.*.*` we could use the driver version.
-	pattern := strings.TrimSuffix(libraryName, ".") + ".*.*"
-	match, _ := filepath.Match(pattern, filename)
-	return match
+	return filename == strings.Join([]string{libraryName, d.driverVersion}, ".")
```
cdesiniotis: Prior to this change we were sanitizing the `librarName` input by trimming any possible version suffix, e.g. `strings.TrimSuffix(libraryName,".")`. Should we continue to do that?

internal/discover/graphics_test.go
tariq1890: This is actually a negative test case. If you see the [discoverer logic](https://github.com/NVIDIA/nvidia-container-toolkit/blob/253f1ea57af167ecbb8cd97c70904f5d7a42d292/internal/discover/graphics.go#L186) the way it is written today, it excludes the mount to libnvidia-allocator if the entry is of the name - `libnvidia-allocator.so.${DRIVER_VERSION}`

internal/discover/graphics.go
tariq1890: Yes, good catch. I will make the change

cdesiniotis: Can you also add a unit test function for this method? Just a few test cases should suffice.

internal/discover/graphics_test.go
cdesiniotis: Got it. I am not sure how practical this test case is since the discoverer should only return a `libnvidia-allocator.so` that matches the driver version.

With that said, we can keep this test case and update the description for clarity. Maybe `libnvidia-allocator not filtered out when version does not equal driver version`.

internal/discover/graphics.go
tariq1890: Please check now

tariq1890: > Can you also add a unit test function for this method? Just a few test cases should suffice.

Will do

cdesiniotis (approved):

PR #1965, opened 2026-07-30
[e2e] avoid use of unpinned ubuntu base image

tariq1890:
This PR fixes the failing e2e tests observed in the main branch,
The root cause is the use of the unpinned "ubuntu" base image which has
moved to `resolute` (Ubuntu 26.04). `resolute` has introduced new assumptions when it comes
to setting up a C build environment.


cdesiniotis (approved):
Thanks @tariq1890 !

Issue #563, opened 2024-06-24
Dynamically expose NVIDIA X.Org X11 display server libraries and configure the container correctly


ehfd:
Also refer: https://github.com/NVIDIA/libnvidia-container/issues/118

This issue is because X11 graphical libraries are not pushed into the Windows Subsystem for Linux (WSL) and thus a full-fledged X11 server is not possible to be deployed.

Because `RM_VERSION` in WSL and regular Linux tend to be different, it is also not possible to download the driver libraries inside the container and unpack them.

This will also benefit regular Linux container environments.

---

On top of PR #548, it would be ideal if it's possible to push `nvidia_drv.so` and `libglxserver_nvidia.so.*` into `/usr/lib/x86_64-linux-gnu/nvidia/xorg/` or (in a different notation) `libRoot + /nvidia/xorg/` inside the container, regardless of where it was found, and the container toolkit to generate a `/usr/share/X11/xorg.conf.d/10-nvidia.conf` file with the following content regardless of whether `10-nvidia.conf` exists in the host (mind the module path between distributions and architectures):

```bash
Section "OutputClass"
    Identifier "nvidia"
    MatchDriver "nvidia-drm"
    Driver "nvidia"
    Option "AllowEmptyInitialConfiguration"
    ModulePath "/usr/lib/x86_64-linux-gnu/nvidia/xorg"
EndSection
```

The above is the default behavior for the `nvidia-driver-550` from Ubuntu APT and I generally like this approach a lot.

---

`nvidia-xconfig` (and also possibly `nvidia-config`, combined with the NVIDIA GTK libraries and `libnvidia-wayland-client.so` which are dependencies) should also be pushed into the container for a full X11 experience.

Then, the X11 aspect will be solved and we developers can call it a day.

Also, injecting 32-bit libraries is definitely desirable for usage with Wine/Proton/etc.

---

Notes:

```bash
  --x-prefix=X-PREFIX
      The prefix under which the X components of the NVIDIA driver will be installed; the default is '/usr/X11R6'
      unless nvidia-installer detects that X.Org >= 7.0 is installed, in which case the default is '/usr'.  Only under
      rare circumstances should this option be used.

  --xfree86-prefix=XFREE86-PREFIX
      This is a deprecated synonym for --x-prefix.

  --x-module-path=X-MODULE-PATH
      The path under which the NVIDIA X server modules will be installed.  If this option is not specified,
[...]

PR #1968, opened 2026-08-02
fix(cdi): include 32-bit libraries in discovery

elibosley:
## Summary

CDI generation omitted the host's 32-bit NVIDIA driver stack on multilib systems; this combines predefined-path and linker-cache discovery so generated specs include both native and compat32 libraries.

## Why This Exists

`ldcache.List()` returns separate 32-bit and 64-bit library sets, but the CDI lookup path discarded the 32-bit set. The default library locator also stopped after finding libraries in a predefined native path, which prevented the linker cache from contributing libraries installed in a separate compat32 directory.

The resulting CDI spec could expose the 64-bit NVIDIA stack while leaving a 32-bit Vulkan application such as a Steam or Proton title without the matching vendor libraries. In that state, the application can select software rendering even though compatible ELF32 NVIDIA libraries are installed on the host.

This is related to #563, which tracks broader graphics-library coverage.

## Resolution

The default library locator now merges predefined-path and linker-cache results and deduplicates the combined list. Linker-cache discovery consumes both architecture lists, with native libraries first so existing version-inference precedence remains stable.

This keeps architecture discovery in the host linker cache rather than adding distribution-specific compat32 paths to CDI generation.

## Reviewer Considerations

- Please confirm that an exported `lookup.Merge` combinator is the preferred shape for combining locator results; it mirrors the existing exported `lookup.First` behavior while retaining successful matches when another locator fails.
- The default locator now returns all unique matches instead of the first successful source. This is intentional because CDI needs every driver-library directory, but callers that infer a driver version still see native libraries first.
- Multilib hosts will generate larger CDI specs because valid ELF32 driver libraries are now included. Hosts without 32-bit linker-cache entries should be unchanged.
- This PR is limited to discovery and regression coverage. It does not add distribution-specific paths or modify runtime selection policy.

## Behavior Changes

- CDI specs generated on multilib hosts include matching 32-bit NVIDIA driver libraries exposed through the host linker cache.
- Native library ordering is preserved for callers that use the first match.
- Duplicate paths found through predefined directories and the linker cache are emitted once.

## Implementation Summary

- Consume both values returned by `ldcache.List()`.
- Add a locator combinator that merges successful results across lookup sources.
- Use the merged, unique locator for default library discovery.
- Add regression tests for mixed 32-bit/64-bit linker-cache entries, ordering, result merging, and partial lookup failure.

## Verification

- `make test` passed.
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run --timeout 5m ./...` passed with 0 issues.
- `go vet ./pkg/lookup` passed.
- `git diff --check` passed.
- Real-host comparison on an x86_64 linux-based system with NVIDIA 595.84:
  - toolkit 1.19.1 generated 40 `/usr/lib64` library mounts and 0 `/usr/lib` library mounts;
  - patched `main` generated 41 `/usr/lib64` library mounts and 25 `/usr/lib` library mounts;
  - the newly discovered `/usr/lib` entries were verified as ELF32 and included `libcuda`, `libnvidia-ml`, and `libGLX_nvidia`.
- The host's live CDI specification was not replaced during validation.
- The commit has a matching DCO sign-off and a GitHub-verified SSH signature.

## Risk

Low to moderate. The discovery set intentionally grows on multilib hosts, and the primary compatibility risk is a caller assuming only one architecture is returned. Native-first ordering, deduplication, package tests, the full upstream test suite, and real-host CDI generation cover that behavior.

### DIFF
--- a/pkg/lookup/ldcache.go
+++ b/pkg/lookup/ldcache.go
@@ -43,16 +43,21 @@ func (f *Factory) newLdcacheLocator() Locator {
 		f.logger.Warningf("Failed to load ldcache: %v", err)
 		return notFound
 	}
+	return f.newLdcacheLocatorFrom(cache)
+}
 
+func (f *Factory) newLdcacheLocatorFrom(cache ldcache.LDCache) Locator {
 	var libraries []string
-	_, libs64 := cache.List()
-	for _, library := range libs64 {
-		chain, err := symlinks.ResolveChain(library)
-		if err != nil {
-			f.logger.Warningf("Failed to resolve symlink chain for library %q: %v", library, err)
-			continue
+	libs32, libs64 := cache.List()
+	for _, libs := range [][]string{libs64, libs32} {
+		for _, library := range libs {
+			chain, err := symlinks.ResolveChain(library)
+			if err != nil {
+				f.logger.Warningf("Failed to resolve symlink chain for library %q: %v", library, err)
+				continue
+			}
+			libraries = append(libraries, chain...)
 		}
-		libraries = append(libraries, chain...)
 	}
 
 	l := &ldcacheLocator{
--- a/pkg/lookup/ldcache_test.go
+++ b/pkg/lookup/ldcache_test.go
@@ -1,12 +1,15 @@
 package lookup
 
 import (
+	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	testlog "github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/require"
 
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/ldcache"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/test"
 )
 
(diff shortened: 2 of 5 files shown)


myeolenv:
Thanks for the contribution. We will review this PR.

henry118:
Hi @elibosley, thanks for the contribution, the idea makes sense. 

One thing I'd suggest is to gate this behind an opt-in flag rather than making it default behavior.

`libnvidia-container`'s legacy path already treats this as opt-in (`--compat32`) feature, worth preserving that behavior rather than switching to always-on.

We need to gate it because on musl-based containers (e.g. Alpine+gcompat), both the 32-bit and 64-bit library directories get appended into a single `/etc/ld-musl-<arch>.path` with no arch tagging. if musl's dynamic linker finds a library (by name) with the wrong arch, the workload would fail.

ehfd:
I agree. This should be gated with `NVIDIA_DRIVER_CAPABILITIES="compat32"`.

henry118:
@elibosley just checking in — are you still planning to continue working on this PR? If not, we can close it for now and reopen it when you're ready to pick it back up.

ehfd:
Opened https://github.com/NVIDIA/nvidia-container-toolkit/pull/2035.

elibosley:
@henry118 apologies, I am terrible at checking my GitHub notifications there are just too many 🥲 - sounds like you guys have it from here! Thank you for checking this out I appreciate it!

henry118:
Thanks @elibosley . I'll close this PR and track of work in #2035 instead.

PR #1970, opened 2026-08-04
Update .nvidia-ci.yml

nivaas219:
Hi maintainers, I noticed that the .integration template references a test stage, but test is not currently defined in the stages list. Is this template still expected to be used? If so, should the test stage be added?

### DIFF
--- a/.nvidia-ci.yml
+++ b/.nvidia-ci.yml
@@ -44,6 +44,7 @@ variables:
 
 stages:
   - pull
+  - test
   - scan
   - release
   - ngc-publish


rajatchopra:
> Hi maintainers, I noticed that the .integration template references a test stage, but test is not currently defined in the stages list. Is this template still expected to be used? If so, should the test stage be added?

We mark that integration tests don't need to be run in the internal ci - it is meant for security scans and pre-release image compilation. Hence 'test' does not need to be addressed.

PR #1971, opened 2026-08-06
oci/state: use crun's non-standard root state field to avoid bundle dir access

### LINKED ISSUE #648: nvidia-cdi-hook fails when running a container with --userns nomap
Running a container with `--gpus all` and either `--userns host` or `--userns keep-id` works, but `--userns nomap` fails. Note that this was tested on Arch Linux with nvidia-container-toolkit 1.16.1.

First with podman 5.2.0 and crun 1.16:

```
$ podman run --rm -it --gpus all --runtime /usr/bin/crun --userns host quay.io/archlinux/archlinux:base
[root@11d41940527d /]#
exit
$ podman run --rm -it --gpus all --runtime /usr/bin/crun --userns keep-id quay.io/archlinux/archlinux:base
[lahwaacz@17ff7ae4f275 /]$
exit
$ podman run --rm -it --gpus all --runtime /usr/bin/crun --userns nomap quay.io/archlinux/archlinux:base
Error: OCI runtime error: /usr/bin/crun: {"msg":"error executing hook `/usr/bin/nvidia-cdi-hook` (exit code: 1)","level":"error","time":"2024-08-13T07:48:19.461304Z"}
```

With podman 5.2.0 and runc 1.1.13:

```
$ podman run --rm -it --gpus all --runtime /usr/bin/runc --userns host quay.io/archlinux/archlinux:base
[root@6b07c771066e /]#
exit
$ podman run --rm -it --gpus all --runtime /usr/bin/runc --userns keep-id quay.io/archlinux/archlinux:base
[klinkovsky@6083384b42e7 /]$
exit
$ podman run --rm -it --gpus all --runtime /usr/bin/runc --userns nomap quay.io/archlinux/archlinux:base
Error: /usr/bin/runc: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: time="2024-08-13T09:49:32+02:00" level=error msg="failed to determined container root: failed to open OCI spec file: open /home/lahwaacz/.local/share/containers/storage/overlay-containers/981df8bef771d05f1e2f2a907cf6a0d0698ab2f24dedb34201ef14d91c6959cc/userdata/config.json: permission denied": OCI permission denied
```

The `failed to determined container root: failed to open OCI spec file` message (notable for its grammar error) comes from `update-ldcache.go`:

https://github.com/NVIDIA/nvidia-container-toolkit/blob/4f0de9f1ef07abd6c1b3dd54995b275fb74d14dc/cmd/nvidia-cdi-hook/update-ldcache/update-ldcache.go#L103-L106

(Unrelated rant at crun and runc: neither of their output is complete, but they complement each other - crun shows which hook failed and runc shows the actual error message...)

henry118:
Partially fixes #648.

The OCI runtime spec defines what a runtime passes to hooks on stdin: a state JSON object containing a container's bundle path. To find the container root, the toolkit currently opens `<bundle>/config.json` and reads `root.path` from it. This requires read access to the bundle directory.

When Podman runs a container with `--userns nomap` or `--userns auto`, the invoking user's UID is intentionally excluded from the container's user namespace UID map. The hook process runs inside the container's user namespace with an unmapped UID, which has no permission to read the bundle directory owned by the host user. As a result, `GetContainerRoot()` fails opening `config.json` with permission denied, and every CDI hook that calls it exits with an error.

`crun` includes a non-standard `root` field in the state JSON it sends to hooks, populated from config.json's `root.path`:

https://github.com/containers/crun/blob/c3a49b822cc45c1ae9e0459eff534efe61c4a5be/src/libcrun/container.c#L746

This provides the `rootfs` path directly, without requiring any access to the bundle directory. `GetContainerRoot()` can now check `s.Root` first. If present, it can be returned directly without accessing the config JSON file.

This fix is limited to `crun` only. `runc` does not include a `root` field in its state JSON. And this fix requires `crun` ≥ 1.27 because 1.27 included a change which allows ignoring `chdir` permission errors before container hook is executed (https://github.com/containers/crun/pull/2035).

Given #648 is reported on Podman and `crun` is podman's preferred and default runtime, this fix should be suffice for majority of the use cases.

### DIFF
--- a/internal/oci/state.go
+++ b/internal/oci/state.go
@@ -27,7 +27,14 @@ import (
 )
 
 // State stores an OCI container state. This includes the spec path and the environment
-type State specs.State
+type State struct {
+	specs.State
+	// Root is a non-standard extension included in the container state JSON by some
+	// OCI runtimes (e.g., crun). When present it provides the rootfs path directly,
+	// avoiding the need to open config.json — which may be permission-denied when
+	// running with user namespaces such as --userns=nomap (issue #648).
+	Root string `json:"root,omitempty"`
+}
 
 // LoadContainerState loads the container state from the specified filename. If the filename is empty or '-' the state is loaded from STDIN
 func LoadContainerState(filename string) (*State, error) {
@@ -59,6 +66,13 @@ func ReadContainerState(reader io.Reader) (*State, error) {
 // GetContainerRoot returns the root for the container from the associated spec. If the spec is not yet loaded, it is
 // loaded and cached.
 func (s *State) GetContainerRoot() (string, error) {
+	if s.Root != "" {
+		if filepath.IsAbs(s.Root) {
+			return s.Root, nil
+		}
+		return filepath.Join(s.Bundle, s.Root), nil
+	}
+
 	spec, err := s.loadMinimalSpec()
 	if err != nil {
 		return "", err
--- a/internal/oci/state.go
+++ b/internal/oci/state.go
@@ -27,7 +27,14 @@ import (
 )
 
 // State stores an OCI container state. This includes the spec path and the environment
-type State specs.State
+type State struct {
+	specs.State
+	// Root is a non-standard extension included in the container state JSON by some
+	// OCI runtimes (e.g., crun). When present it provides the rootfs path directly,
+	// avoiding the need to open config.json — which may be permission-denied when
+	// running with user namespaces such as --userns=nomap (issue #648).
+	Root string `json:"root,omitempty"`
+}
 
 // LoadContainerState loads the container state from the specified filename. If the filename is empty or '-' the state is loaded from STDIN
 func LoadContainerState(filename string) (*State, error) {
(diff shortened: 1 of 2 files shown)


internal/oci/state.go
elezar: What about adding a `State.getRoot() (string, error)` helper to determine the root so that we don't have to duplicate the `IsAbs` and `Join` calls?


```suggestion
func (s *State) getRoot() (string, error) {
	if s.Root != "" {
		return s.Root, nil
	}

	spec, err := s.loadMinimalSpec()
	if err != nil {
		return "", err
	}

	if spec.Root != nil {
		return spec.Root.Path, nil
	}

	return "", nil
}

// GetContainerRoot returns the root for the container from the associated spec. If the spec is not yet loaded, it is
// loaded and cached.
func (s *State) GetContainerRoot() (string, error) {
	containerRoot, err := s.getRoot()
	if err != nil {
		return "", err
	}
```

henry118: Updated.

cdesiniotis (approved):

cdesiniotis (approved):

Issue #1477, opened 2020-11-28
Automatically provisioning X11 and Wayland devices of GPU inside container?


ehfd:
Redirected from https://github.com/NVIDIA/k8s-device-plugin/issues/206 to a more suitable repository.

### 1. Issue or feature description
In docker and kubernetes, people have had to have manual host setup to provision the X server using host path directive `/tmp/.X11-unix`. This is quite tedious for sysadmins and at the same time a security threat as people can spoof the host.

To mitigate this, there have been attempts (https://github.com/ehfd/docker-nvidia-glx-desktop which is based on https://github.com/ryought/glx-docker-headless-gpu) to execute an X server and use GLX inside the container after getting provisioned the GPU using libnvidia-container.

An alternative was created by the developers at VirtualGL (used widely in HPC to enable GPU-based rendering in VNC virtual display environments) have been able to develop a feature that uses the EGL API to enable 3D GL rendering such as Blender, Matlab, and Unity, previously only possible with GLX and thus an X server. As you guys know well, nvidia-docker does not support GLX but has introduced the EGL API just below two years ago. 
See EGL config section of https://github.com/VirtualGL/virtualgl/issues/113#issuecomment-693127236

EGL is also required to start a Wayland compositor inside a container with the EGLStreams specification in NVIDIA GPUs, which is the way forward after X11 development has stopped.

**These use cases require access to the devices `/dev/dri/cardX` corresponding to each GPU provisioned using libnvidia-container. However, it does not seem like libnvidia-container provisions this automatically. I would like to ask you whether this is possible, and how this can be configured.**

### 2. Steps to reproduce the issue
Provision one GPU inside container `nvidia/cudagl:11.0-devel-ubuntu20.04` or `nvidia/opengl:1.2-glvnd-devel-ubuntu20.04` in Docker CE 19.03 (or using one `nvidia.com/gpu: 1` with k8s-device-plugin v0.7.0 with default configurations in Kubernetes v1.18.6).

Do: ```ls /dev```

Result: Inside the container you see `/dev/nvidiaX`, `/dev/nvidia-modeset`, `/dev/nvidia-uvm`, `/dev/nvidia-uvm-tools`, HOWEVER directory `/dev/dri` does not exist.
Wayland compositors are unlikely to start inside a container without DRM devices. VirtualGL does not work through any devices other than `/dev/dri/cardX` as well.

[...]

Issue #1735, opened 2026-03-16
[Bug]: nvidia-cdi-refresh.service excessively broad ordering constraint


brycehalling:
**Describe the bug**
In the 1.19.0 release, a new `After=multi-user.target` requirement was added to `nvidia-cdi-refresh.service`. This new ordering dependency makes it impossible to start any unit as part of `multi-user.target` that depends on `nvidia-cdi-refresh.service` as it creates an ordering cycle in systemd. I expect this is a fairly common use-case where systemd is used to start containers that require the toolkit on boot.

It was added as part of a "fix" for https://github.com/NVIDIA/nvidia-container-toolkit/issues/1611. In that issue, it is notable that there doesn't seem to be any explanation as to _why_ this dependency resolves the problem. The reporter's hastily self-tested workaround was just taken at face value and added to the unit file.

In general, ordering dependencies shipped in a package should be as minimal as possible to make the service work. A user can always add a drop-in for the unit that adds additional ordering constraints if their system requires it, but users cannot create a drop-in to remove ordering constraints that are shipped in the main unit file. 

**To Reproduce**
Update to the 1.19.0 version
```
► rpm -q nvidia-container-toolkit{,-base}
nvidia-container-toolkit-1.19.0-1.x86_64
nvidia-container-toolkit-base-1.19.0-1.x86_64
```

Create and enable a service that has an `After=nvidia-cdi-refresh.service` dependency, and whose `Install` section specifies `WantedBy=multi-user.target`. For example:
```
► sudo systemctl cat podman-compose.service 
# /etc/systemd/system/podman-compose.service
[Unit]
Description=podman-compose in /home/bryce/Software/podman-compose
Requires=nvidia-cdi-refresh.service
After=nvidia-cdi-refresh.service

[Service]
Type=simple
ExecStartPre=-/usr/bin/podman compose -f /home/bryce/Software/podman-compose/compose.yaml up --no-start 
ExecStartPre=/usr/bin/podman compose -f /home/bryce/Software/podman-compose/compose.yaml start
ExecStart=/usr/bin/podman compose -f /home/bryce/Software/podman-compose/compose.yaml wait
ExecStop=/usr/bin/podman compose -f /home/bryce/Software/podman-compose/compose.yaml down -t 30

[Install]
WantedBy=multi-user.target
```

On boot, systemd will flag an ordering cycle and one of the units will not be started in an effort to break that cycle.
```
[...]

PR #1979, opened 2026-08-12
fix: Create NVIDIA control device nodes in nvidia-cdi-refresh.service

### LINKED ISSUE #1982: [Bug]: nvidia-cdi-refresh service at boot causing container Vulkan errors
Similar to #1611 the nvidia-cdi-refresh service seems to run too early (?) and creates a cdi file that doesn't quite work - in this particular case it is the vulkan drivers that fail in the container as they can't create a surface (e.g. trying to run vkcube).

I'm running Fedora 44 with cosmic desktop.

For now I disable the service 
```sh 
$ sudo systemctl disable nvidia-cdi-refresh.service
Removed '/etc/systemd/system/multi-user.target.wants/nvidia-cdi-refresh.service'.
Disabling 'nvidia-cdi-refresh.service', but its triggering units are still active:
nvidia-cdi-refresh.path
```

But have to change the location of the generated cdi (by modifying `/etc/systemd/system/nvidia-cdi-refresh.service`)

```sh
#Environment=NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/var/run/cdi/nvidia.yaml
Environment=NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/etc/cdi/nvidia.yaml
```

The log below is from the service running at boot up and then running again (manual restart).  The line `level=warning msg="Could not locate /dev/nvidia-modeset: /dev/nvidia-modeset: not found"` stands out but I'm not sure what that means?

Is there a better workaround/fix that will survive nvidia-ctk updates?

```
journalctl -u nvidia-cdi-refresh.service -b                                                                                                                                      [11:03:55]
Aug 12 09:58:44 lnx0021 systemd[1]: Starting nvidia-cdi-refresh.service - Refresh NVIDIA CDI specification file...
Aug 12 09:58:45 lnx0021 sh[1970]: GPU 0: NVIDIA GeForce RTX 3080 Ti Laptop GPU (UUID: GPU-8b41da4e-1fb9-c561-350c-2a6584c3b378)
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Using /usr/lib64/libnvidia-sandboxutils.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Using /usr/lib64/libnvidia-ml.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Auto-detected mode as 'nvml'"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Using driver version 610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate /dev/nvidia-modeset: /dev/nvidia-modeset: not found"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /dev/nvidia-uvm-tools as /dev/nvidia-uvm-tools"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /dev/nvidia-uvm as /dev/nvidia-uvm"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /dev/nvidiactl as /dev/nvidiactl"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-egl-gbm.so.1.1.3 as /usr/lib64/libnvidia-egl-gbm.so.1.1.3"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-egl-wayland.so.1.1.21 as /usr/lib64/libnvidia-egl-wayland.so.1.1.21"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-egl-wayland2.so.1.0.1 as /usr/lib64/libnvidia-egl-wayland2.so.1.0.1"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-allocator.so.610.57.04 as /usr/lib64/libnvidia-allocator.so.610.57.>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate libnvidia-vulkan-producer.so.610.57.04: libnvidia-vulkan-producer.so.610.57.04>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/xorg/modules/drivers/nvidia_drv.so as /usr/lib64/xorg/modules/drivers/nvidia_>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate libglxserver_nvidia.so.610.57.04: libglxserver_nvidia.so.610.57.04: not found"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/share/glvnd/egl_vendor.d/10_nvidia.json as /usr/share/glvnd/egl_vendor.d/10_nvidia.>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json as /usr/share/egl/egl_external>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/share/egl/egl_external_platform.d/10_nvidia_wayland.json as /usr/share/egl/egl_exte>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/share/egl/egl_external_platform.d/09_nvidia_wayland2.json as /usr/share/egl/egl_ext>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/share/nvidia/nvoptix.bin as /usr/share/nvidia/nvoptix.bin"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate X11/xorg.conf.d/10-nvidia.conf: X11/xorg.conf.d/10-nvidia.conf: not found"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate X11/xorg.conf.d/nvidia-drm-outputclass.conf: X11/xorg.conf.d/nvidia-drm-output>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate vulkan/icd.d/nvidia_icd.json: vulkan/icd.d/nvidia_icd.json: not found\nvulkan/>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate vulkan/icd.d/nvidia_layers.json: vulkan/icd.d/nvidia_layers.json: not found\nv>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/share/vulkan/implicit_layer.d/nvidia_layers.json as /etc/vulkan/implicit_layer.d/nv>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/share/vulkan/icd.d/nvidia_icd.x86_64.json as /etc/vulkan/icd.d/nvidia_icd.x86_64.js>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libEGL_nvidia.so.610.57.04 as /usr/lib64/libEGL_nvidia.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libGLESv1_CM_nvidia.so.610.57.04 as /usr/lib64/libGLESv1_CM_nvidia.so.610.57.>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libGLESv2_nvidia.so.610.57.04 as /usr/lib64/libGLESv2_nvidia.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libGLX_nvidia.so.610.57.04 as /usr/lib64/libGLX_nvidia.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libcuda.so.610.57.04 as /usr/lib64/libcuda.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libcudadebugger.so.610.57.04 as /usr/lib64/libcudadebugger.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvcuvid.so.610.57.04 as /usr/lib64/libnvcuvid.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-allocator.so.610.57.04 as /usr/lib64/libnvidia-allocator.so.610.57.>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-cfg.so.610.57.04 as /usr/lib64/libnvidia-cfg.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-eglcore.so.610.57.04 as /usr/lib64/libnvidia-eglcore.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-encode.so.610.57.04 as /usr/lib64/libnvidia-encode.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-fbc.so.610.57.04 as /usr/lib64/libnvidia-fbc.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-glcore.so.610.57.04 as /usr/lib64/libnvidia-glcore.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-glsi.so.610.57.04 as /usr/lib64/libnvidia-glsi.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-glvkspirv.so.610.57.04 as /usr/lib64/libnvidia-glvkspirv.so.610.57.>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-gpucomp.so.610.57.04 as /usr/lib64/libnvidia-gpucomp.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-gtk3.so.610.57.04 as /usr/lib64/libnvidia-gtk3.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-ml.so.610.57.04 as /usr/lib64/libnvidia-ml.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-ngx.so.610.57.04 as /usr/lib64/libnvidia-ngx.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-nvvm.so.610.57.04 as /usr/lib64/libnvidia-nvvm.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-opencl.so.610.57.04 as /usr/lib64/libnvidia-opencl.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-opticalflow.so.610.57.04 as /usr/lib64/libnvidia-opticalflow.so.610>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-pkcs11-openssl3.so.610.57.04 as /usr/lib64/libnvidia-pkcs11-openssl>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-present.so.610.57.04 as /usr/lib64/libnvidia-present.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-ptxjitcompiler.so.610.57.04 as /usr/lib64/libnvidia-ptxjitcompiler.>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-rtcore.so.610.57.04 as /usr/lib64/libnvidia-rtcore.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-sandboxutils.so.610.57.04 as /usr/lib64/libnvidia-sandboxutils.so.6>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-tileiras.so.610.57.04 as /usr/lib64/libnvidia-tileiras.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-tls.so.610.57.04 as /usr/lib64/libnvidia-tls.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-vksc-core.so.610.57.04 as /usr/lib64/libnvidia-vksc-core.so.610.57.>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvidia-wayland-client.so.610.57.04 as /usr/lib64/libnvidia-wayland-client.>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/libnvoptix.so.610.57.04 as /usr/lib64/libnvoptix.so.610.57.04"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/lib64/vdpau/libvdpau_nvidia.so.610.57.04 as /usr/lib64/vdpau/libvdpau_nvidia.so.610>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate /nvidia-persistenced/socket: /nvidia-persistenced/socket: not found"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate /nvidia-fabricmanager/socket: /nvidia-fabricmanager/socket: not found"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate /tmp/nvidia-mps: /tmp/nvidia-mps: not found"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /lib/firmware/nvidia/610.57.04/gsp_ga10x.bin as /lib/firmware/nvidia/610.57.04/gsp_ga10x>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /lib/firmware/nvidia/610.57.04/gsp_tu10x.bin as /lib/firmware/nvidia/610.57.04/gsp_tu10x>
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/bin/nvidia-smi as /usr/bin/nvidia-smi"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/bin/nvidia-debugdump as /usr/bin/nvidia-debugdump"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/bin/nvidia-persistenced as /usr/bin/nvidia-persistenced"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/bin/nvidia-cuda-mps-control as /usr/bin/nvidia-cuda-mps-control"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Selecting /usr/bin/nvidia-cuda-mps-server as /usr/bin/nvidia-cuda-mps-server"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate nvidia-imex: nvidia-imex: not found"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=warning msg="Could not locate nvidia-imex-ctl: nvidia-imex-ctl: not found"
Aug 12 09:58:45 lnx0021 nvidia-ctk[2159]: time="2026-08-12T09:58:45+01:00" level=info msg="Generated CDI spec with version 0.7.0"
Aug 12 09:58:45 lnx0021 systemd[1]: nvidia-cdi-refresh.service: Deactivated successfully.
Aug 12 09:58:45 lnx0021 systemd[1]: Finished nvidia-cdi-refresh.service - Refresh NVIDIA CDI specification file.
Aug 12 09:58:45 lnx0021 systemd[1]: nvidia-cdi-refresh.service: Consumed 334ms CPU time over 643ms wall clock time, 90.1M memory peak.
Aug 12 10:00:42 lnx0021 systemd[1]: Starting nvidia-cdi-refresh.service - Refresh NVIDIA CDI specification file...
Aug 12 10:00:42 lnx0021 sh[10844]: GPU 0: NVIDIA GeForce RTX 3080 Ti Laptop GPU (UUID: GPU-8b41da4e-1fb9-c561-350c-2a6584c3b378)
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Using /usr/lib64/libnvidia-sandboxutils.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Using /usr/lib64/libnvidia-ml.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Auto-detected mode as 'nvml'"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Using driver version 610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /dev/nvidia-modeset as /dev/nvidia-modeset"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /dev/nvidia-uvm-tools as /dev/nvidia-uvm-tools"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /dev/nvidia-uvm as /dev/nvidia-uvm"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /dev/nvidiactl as /dev/nvidiactl"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-egl-gbm.so.1.1.3 as /usr/lib64/libnvidia-egl-gbm.so.1.1.3"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-egl-wayland.so.1.1.21 as /usr/lib64/libnvidia-egl-wayland.so.1.1.2>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-egl-wayland2.so.1.0.1 as /usr/lib64/libnvidia-egl-wayland2.so.1.0.>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-allocator.so.610.57.04 as /usr/lib64/libnvidia-allocator.so.610.57>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate libnvidia-vulkan-producer.so.610.57.04: libnvidia-vulkan-producer.so.610.57.0>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/xorg/modules/drivers/nvidia_drv.so as /usr/lib64/xorg/modules/drivers/nvidia>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate libglxserver_nvidia.so.610.57.04: libglxserver_nvidia.so.610.57.04: not found"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/share/glvnd/egl_vendor.d/10_nvidia.json as /usr/share/glvnd/egl_vendor.d/10_nvidia>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json as /usr/share/egl/egl_externa>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/share/egl/egl_external_platform.d/10_nvidia_wayland.json as /usr/share/egl/egl_ext>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/share/egl/egl_external_platform.d/09_nvidia_wayland2.json as /usr/share/egl/egl_ex>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/share/nvidia/nvoptix.bin as /usr/share/nvidia/nvoptix.bin"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate X11/xorg.conf.d/10-nvidia.conf: X11/xorg.conf.d/10-nvidia.conf: not found"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate X11/xorg.conf.d/nvidia-drm-outputclass.conf: X11/xorg.conf.d/nvidia-drm-outpu>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate vulkan/icd.d/nvidia_icd.json: vulkan/icd.d/nvidia_icd.json: not found\nvulkan>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate vulkan/icd.d/nvidia_layers.json: vulkan/icd.d/nvidia_layers.json: not found\n>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/share/vulkan/implicit_layer.d/nvidia_layers.json as /etc/vulkan/implicit_layer.d/n>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/share/vulkan/icd.d/nvidia_icd.x86_64.json as /etc/vulkan/icd.d/nvidia_icd.x86_64.j>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libEGL_nvidia.so.610.57.04 as /usr/lib64/libEGL_nvidia.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libGLESv1_CM_nvidia.so.610.57.04 as /usr/lib64/libGLESv1_CM_nvidia.so.610.57>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libGLESv2_nvidia.so.610.57.04 as /usr/lib64/libGLESv2_nvidia.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libGLX_nvidia.so.610.57.04 as /usr/lib64/libGLX_nvidia.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libcuda.so.610.57.04 as /usr/lib64/libcuda.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libcudadebugger.so.610.57.04 as /usr/lib64/libcudadebugger.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvcuvid.so.610.57.04 as /usr/lib64/libnvcuvid.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-allocator.so.610.57.04 as /usr/lib64/libnvidia-allocator.so.610.57>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-cfg.so.610.57.04 as /usr/lib64/libnvidia-cfg.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-eglcore.so.610.57.04 as /usr/lib64/libnvidia-eglcore.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-encode.so.610.57.04 as /usr/lib64/libnvidia-encode.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-fbc.so.610.57.04 as /usr/lib64/libnvidia-fbc.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-glcore.so.610.57.04 as /usr/lib64/libnvidia-glcore.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-glsi.so.610.57.04 as /usr/lib64/libnvidia-glsi.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-glvkspirv.so.610.57.04 as /usr/lib64/libnvidia-glvkspirv.so.610.57>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-gpucomp.so.610.57.04 as /usr/lib64/libnvidia-gpucomp.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-gtk3.so.610.57.04 as /usr/lib64/libnvidia-gtk3.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-ml.so.610.57.04 as /usr/lib64/libnvidia-ml.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-ngx.so.610.57.04 as /usr/lib64/libnvidia-ngx.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-nvvm.so.610.57.04 as /usr/lib64/libnvidia-nvvm.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-opencl.so.610.57.04 as /usr/lib64/libnvidia-opencl.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-opticalflow.so.610.57.04 as /usr/lib64/libnvidia-opticalflow.so.61>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-pkcs11-openssl3.so.610.57.04 as /usr/lib64/libnvidia-pkcs11-openss>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-present.so.610.57.04 as /usr/lib64/libnvidia-present.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-ptxjitcompiler.so.610.57.04 as /usr/lib64/libnvidia-ptxjitcompiler>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-rtcore.so.610.57.04 as /usr/lib64/libnvidia-rtcore.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-sandboxutils.so.610.57.04 as /usr/lib64/libnvidia-sandboxutils.so.>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-tileiras.so.610.57.04 as /usr/lib64/libnvidia-tileiras.so.610.57.0>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-tls.so.610.57.04 as /usr/lib64/libnvidia-tls.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-vksc-core.so.610.57.04 as /usr/lib64/libnvidia-vksc-core.so.610.57>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvidia-wayland-client.so.610.57.04 as /usr/lib64/libnvidia-wayland-client>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/libnvoptix.so.610.57.04 as /usr/lib64/libnvoptix.so.610.57.04"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/lib64/vdpau/libvdpau_nvidia.so.610.57.04 as /usr/lib64/vdpau/libvdpau_nvidia.so.61>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate /nvidia-persistenced/socket: /nvidia-persistenced/socket: not found"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate /nvidia-fabricmanager/socket: /nvidia-fabricmanager/socket: not found"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate /tmp/nvidia-mps: /tmp/nvidia-mps: not found"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /lib/firmware/nvidia/610.57.04/gsp_ga10x.bin as /lib/firmware/nvidia/610.57.04/gsp_ga10>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /lib/firmware/nvidia/610.57.04/gsp_tu10x.bin as /lib/firmware/nvidia/610.57.04/gsp_tu10>
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/bin/nvidia-smi as /usr/bin/nvidia-smi"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/bin/nvidia-debugdump as /usr/bin/nvidia-debugdump"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/bin/nvidia-persistenced as /usr/bin/nvidia-persistenced"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/bin/nvidia-cuda-mps-control as /usr/bin/nvidia-cuda-mps-control"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Selecting /usr/bin/nvidia-cuda-mps-server as /usr/bin/nvidia-cuda-mps-server"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate nvidia-imex: nvidia-imex: not found"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=warning msg="Could not locate nvidia-imex-ctl: nvidia-imex-ctl: not found"
Aug 12 10:00:42 lnx0021 nvidia-ctk[10846]: time="2026-08-12T10:00:42+01:00" level=info msg="Generated CDI spec with version 0.7.0"
Aug 12 10:00:42 lnx0021 systemd[1]: nvidia-cdi-refresh.service: Deactivated successfully.
Aug 12 10:00:42 lnx0021 systemd[1]: Finished nvidia-cdi-refresh.service - Refresh NVIDIA CDI specification file.
```

ehfd:
Reviewers: @elezar @myeolenv 

Associated with https://github.com/NVIDIA/libnvidia-container/pull/388.

Closes #1982

## Summary

**Re-enables Wayland and associated Vulkan and OpenGL capabilities on the container toolkit without workarounds.**

CDI spec generation only discovered NVIDIA control device nodes that already existed on the host; on systems where no host component triggers their creation, `/dev/nvidia-modeset` (and, after a reboot, the UVM nodes) were silently omitted from the generated specifications. This change creates missing control device nodes ahead of discovery — the programmatic equivalent of running `nvidia-ctk system create-device-nodes --control-devices --load-kernel-modules` at the points where specifications are produced. The legacy-mode counterpart (mounting the node for the `graphics` capability) is NVIDIA/libnvidia-container#388, keeping each half in the component that owns the behavior.

## Why This Exists

The NVIDIA driver does not register its device nodes with devtmpfs; nodes such as `/dev/nvidia-modeset` are created on demand by userspace driver components. On headless Docker hosts (typical Debian, Ubuntu, and Arch installations) nothing ever triggers modeset creation, so:

- `nvidia-ctk cdi generate`, the `nvidia-cdi-refresh` service, and jit-cdi mode (the default resolution of `mode = "auto"`) generated specs without the node, and containers requesting any capability set never received it. Wayland/KMS clients in containers fail as a result.
- In legacy mode, `nvidia-container-cli` creates the node under `--load-kmods` but mounts it only for the `display` capability; NVIDIA/libnvidia-container#388 extends that mount to `graphics`.
- The Kubernetes toolkit installer already pre-creates control device nodes, which explains why GPU Operator nodes behaved differently from plain Docker hosts for identical images.

This closes the device half of the container Wayland/X11 enablement tracked in #1477 and #563.

### DIFF
--- a/deployments/systemd/nvidia-cdi-refresh.service
+++ b/deployments/systemd/nvidia-cdi-refresh.service
@@ -29,6 +29,13 @@ Environment=NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/var/run/cdi/nvidia.yaml
 EnvironmentFile=-/etc/nvidia-container-toolkit/nvidia-cdi-refresh.env
 ExecCondition=/bin/sh -c '/usr/bin/grep -qE "/(nvidia|nvidia-current)[.]ko" /lib/modules/%v/modules.dep || [ -e /dev/dxg ]'
 ExecStart=/bin/sh -c '/usr/bin/nvidia-smi -L || /usr/sbin/nvidia-smi -L || /usr/lib/wsl/lib/nvidia-smi -L'
+# Create any missing NVIDIA control device nodes (loading the kernel modules
+# they require) before generating the CDI specification; nothing else is
+# guaranteed to create them. Best-effort, and skipped on WSL where /dev/dxg is
+# used instead, as well as on Tegra platforms using the nvgpu kernel module.
+# If module loading fails (for example while a driver upgrade has replaced the
+# on-disk modules) still create the nodes for the modules that are loaded.
+ExecStart=-/bin/sh -c '[ -e /dev/dxg ] || /usr/bin/grep -q "^nvgpu " /proc/modules || /usr/bin/nvidia-ctk system create-device-nodes --control-devices --load-kernel-modules || /usr/bin/nvidia-ctk system create-device-nodes --control-devices'
 ExecStart=/usr/bin/nvidia-ctk cdi generate
 CapabilityBoundingSet=CAP_SYS_MODULE CAP_SYS_ADMIN CAP_MKNOD
 # We set the service to restart on failure to ensure that a CDI spec is
--- a/deployments/udev/99-nvidia-cdi-refresh.rules
+++ b/deployments/udev/99-nvidia-cdi-refresh.rules
@@ -1,0 +1,22 @@
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The NVIDIA control device nodes (/dev/nvidiactl, /dev/nvidia-modeset, and
+# /dev/nvidia-uvm*) are created on demand by userspace and are not provided by
+# devtmpfs, so loading the NVIDIA kernel module does not guarantee that they
+# exist. Trigger nvidia-cdi-refresh.service on module load to create the
+# missing nodes and refresh the CDI specification.
+# Debian packages the kernel module as nvidia-current, which appears in sysfs
+# as nvidia_current.
+ACTION=="add", SUBSYSTEM=="module", KERNEL=="nvidia|nvidia_current", TAG+="systemd", ENV{SYSTEMD_WANTS}+="nvidia-cdi-refresh.service"
--- a/deployments/systemd/nvidia-cdi-refresh.env
+++ b/deployments/systemd/nvidia-cdi-refresh.env
@@ -18,3 +18,12 @@
 # For example, to change the for the generated CDI specification update and
 # uncomment the following line:
 # NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/var/run/cdi/nvidia.yaml
+
+# The service also runs
+#   nvidia-ctk system create-device-nodes --control-devices --load-kernel-modules
+# before generating the CDI specification. Its driver and device roots can be
+# overridden by uncommenting the following lines (nvidia-ctk cdi generate uses
+# the separate NVIDIA_CTK_DRIVER_ROOT and NVIDIA_CTK_DEV_ROOT variables):
+#
+# NVIDIA_DRIVER_ROOT=/
+# NVIDIA_DEV_ROOT=/
--- a/docker/Dockerfile.debian
+++ b/docker/Dockerfile.debian
@@ -56,6 +56,7 @@ RUN make PREFIX=${DIST_DIR} cmds
 WORKDIR $DIST_DIR
 COPY packaging/debian ./debian
 COPY deployments/systemd/ .
+COPY deployments/udev/ .
 
 RUN dch --create --package="${PKG_NAME}" \
         --newversion "${REVISION}" \
--- a/docker/Dockerfile.opensuse-leap
+++ b/docker/Dockerfile.opensuse-leap
@@ -48,6 +48,7 @@ RUN make PREFIX=${DIST_DIR} cmds
 WORKDIR $DIST_DIR/..
 COPY packaging/rpm .
 COPY deployments/systemd/ ${DIST_DIR}/
+COPY deployments/udev/ ${DIST_DIR}/
 
 CMD arch=$(uname -m) && \
     rpmbuild --clean --target=$arch -bb \
--- a/docker/Dockerfile.rpm-yum
+++ b/docker/Dockerfile.rpm-yum
@@ -76,6 +76,7 @@ RUN make PREFIX=${DIST_DIR} cmds
 WORKDIR $DIST_DIR/..
 COPY packaging/rpm .
 COPY deployments/systemd/ ${DIST_DIR}/
+COPY deployments/udev/ ${DIST_DIR}/
 
 CMD arch=$(uname -m) && \
     rpmbuild --clean --target=$arch -bb \
--- a/docker/Dockerfile.ubuntu
+++ b/docker/Dockerfile.ubuntu
@@ -54,6 +54,7 @@ RUN make PREFIX=${DIST_DIR} cmds
 WORKDIR $DIST_DIR
 COPY packaging/debian ./debian
 COPY deployments/systemd/ .
+COPY deployments/udev/ .
 
 RUN dch --create --package="${PKG_NAME}" \
         --newversion "${REVISION}" \
--- a/packaging/debian/nvidia-container-toolkit-base.install
+++ b/packaging/debian/nvidia-container-toolkit-base.install
@@ -4,3 +4,4 @@ nvidia-cdi-hook /usr/bin
 nvidia-cdi-refresh.service /lib/systemd/system/
 nvidia-cdi-refresh.path /lib/systemd/system/
 nvidia-cdi-refresh.env /etc/nvidia-container-toolkit/
+99-nvidia-cdi-refresh.rules /lib/udev/rules.d/
--- a/packaging/debian/nvidia-container-toolkit-base.postinst
+++ b/packaging/debian/nvidia-container-toolkit-base.postinst
@@ -5,6 +5,12 @@ set -e
 case "$1" in
     configure)
         /usr/bin/nvidia-ctk --quiet config --config-file=/etc/nvidia-container-runtime/config.toml --in-place
+
+        # Reload udev rules so that the nvidia-cdi-refresh rules take effect
+        # without a reboot.
+        if command -v udevadm >/dev/null 2>&1; then
+            udevadm control --reload-rules >/dev/null 2>&1 || true
+        fi
     ;;
 
     abort-upgrade|abort-remove|abort-deconfigure)
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -16,3 +16,4 @@ override_dh_fixperms:
 	chmod 755 debian/$(shell dh_listpackages)/usr/bin/nvidia-cdi-hook || true
 	chmod 644 debian/$(shell dh_listpackages)/lib/systemd/system/nvidia-cdi-refresh.service || true
 	chmod 644 debian/$(shell dh_listpackages)/lib/systemd/system/nvidia-cdi-refresh.path || true
+	chmod 644 debian/$(shell dh_listpackages)/lib/udev/rules.d/99-nvidia-cdi-refresh.rules || true
--- a/packaging/rpm/SPECS/nvidia-container-toolkit.spec
+++ b/packaging/rpm/SPECS/nvidia-container-toolkit.spec
@@ -21,12 +21,17 @@ Source7: nvidia-cdi-refresh.service
 Source8: nvidia-cdi-refresh.path
 Source9: nvidia-cdi-refresh.env
 Source10: 90-nvidia-container-toolkit.preset
+Source11: 99-nvidia-cdi-refresh.rules
 
 %if 0%{?rhel} == 7 || 0%{?amzn} == 2
 BuildRequires: systemd
 %else
 BuildRequires: systemd-rpm-macros
 %endif
+
+# Distributions that don't define the udev rules directory macro fall back to
+# the default location.
+%{!?_udevrulesdir: %global _udevrulesdir %{_prefix}/lib/udev/rules.d}
 
 Obsoletes: nvidia-container-runtime <= 3.5.0-1, nvidia-container-runtime-hook <= 1.4.0-2
 Provides: nvidia-container-runtime
@@ -38,12 +43,13 @@ Requires: nvidia-container-toolkit-base == %{version}-%{release}
 Provides tools and utilities to enable GPU support in containers.
 
 %prep
-cp %{SOURCE0} %{SOURCE1} %{SOURCE2} %{SOURCE3} %{SOURCE4} %{SOURCE5} %{SOURCE6} %{SOURCE7} %{SOURCE8} %{SOURCE9} %{SOURCE10} .
+cp %{SOURCE0} %{SOURCE1} %{SOURCE2} %{SOURCE3} %{SOURCE4} %{SOURCE5} %{SOURCE6} %{SOURCE7} %{SOURCE8} %{SOURCE9} %{SOURCE10} %{SOURCE11} .
 
 %install
 mkdir -p %{buildroot}%{_bindir}
 mkdir -p %{buildroot}%{_unitdir}
 mkdir -p %{buildroot}%{_presetdir}
+mkdir -p %{buildroot}%{_udevrulesdir}
 mkdir -p %{buildroot}%{_sysconfdir}/nvidia-container-toolkit
 
 install -m 755 -t %{buildroot}%{_bindir} nvidia-container-runtime-hook
@@ -55,6 +61,7 @@ install -m 755 -t %{buildroot}%{_bindir} nvidia-cdi-hook
 install -m 644 -t %{buildroot}%{_unitdir} nvidia-cdi-refresh.service
 install -m 644 -t %{buildroot}%{_unitdir} nvidia-cdi-refresh.path
 install -m 644 -t %{buildroot}%{_presetdir} 90-nvidia-container-toolkit.preset
+install -m 644 -t %{buildroot}%{_udevrulesdir} 99-nvidia-cdi-refresh.rules
 install -m 644 -t %{buildroot}%{_sysconfdir}/nvidia-container-toolkit nvidia-cdi-refresh.env
 
 %post
@@ -129,6 +136,12 @@ for unit in nvidia-cdi-refresh.path nvidia-cdi-refresh.service; do
   fi
 done
 
+# Reload udev rules so that the nvidia-cdi-refresh rules take effect without a
+# reboot.
+if command -v udevadm >/dev/null 2>&1; then
+  udevadm control --reload-rules >/dev/null 2>&1 || :
+fi
+
 # Trigger CDI refresh on running systemd hosts without making install depend on
 # the current system state.
 if command -v systemctl >/dev/null 2>&1; then
@@ -148,6 +161,7 @@ fi
 %{_unitdir}/nvidia-cdi-refresh.service
 %{_unitdir}/nvidia-cdi-refresh.path
 %{_presetdir}/90-nvidia-container-toolkit.preset
+%{_udevrulesdir}/99-nvidia-cdi-refresh.rules
 %config(noreplace) %{_sysconfdir}/nvidia-container-toolkit/nvidia-cdi-refresh.env
 
 # The OPERATOR EXTENSIONS package consists of components that are required to enable GPU support in Kubernetes.
--- a/tests/e2e/nvidia-cdi-refresh_test.go
+++ b/tests/e2e/nvidia-cdi-refresh_test.go
@@ -93,6 +93,13 @@ EOF
 	fi
 	`
 
+	nvidiaCdiRefreshUdevRulesInstalledTemplate = `
+	if [ ! -f /lib/udev/rules.d/99-nvidia-cdi-refresh.rules ]; then
+		echo "99-nvidia-cdi-refresh.rules is not installed"
+		exit 1
+	fi
+	`
+
 	nvidiaCdiRefreshFileExistsTemplate = `
 	# is /var/run/cdi/nvidia.yaml exists? and exit with 0 if it does not exist
 	if [ ! -f /var/run/cdi/nvidia.yaml ]; then
@@ -209,6 +216,11 @@ var _ = Describe("nvidia-cdi-refresh", Ordered, ContinueOnFailure, Label("system
 			Expect(err).ToNot(HaveOccurred())
 		})
 
+		It("should install the nvidia-cdi-refresh udev rules", func(ctx context.Context) {
+			_, _, err := systemdRunner.Run(nvidiaCdiRefreshUdevRulesInstalledTemplate)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
 		It("should generate the nvidia.yaml file", func(ctx context.Context) {
 			_, _, err := systemdRunner.Run(nvidiaCdiRefreshFileExistsTemplate)
 			Expect(err).ToNot(HaveOccurred())
@@ -255,6 +267,11 @@ var _ = Describe("nvidia-cdi-refresh", Ordered, ContinueOnFailure, Label("system
 
 		It("should load the nvidia-cdi-refresh.service unit", func(ctx context.Context) {
 			_, _, err := systemdRunner.Run(nvidiaCdiRefreshServiceLoadedTemplate)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should install the nvidia-cdi-refresh udev rules", func(ctx context.Context) {
+			_, _, err := systemdRunner.Run(nvidiaCdiRefreshUdevRulesInstalledTemplate)
 			Expect(err).ToNot(HaveOccurred())
 		})
 


internal/modifier/graphics.go
```diff
@@ -56,11 +57,36 @@ func (f *Factory) newGraphicsModifier() (oci.SpecModifier, error) {
 
 	d := discover.Merge(
 		drmNodes,
+		f.newModesetDeviceNodeDiscoverer(devRoot),
 		mounts,
```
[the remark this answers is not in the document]
ehfd: The runtime constructs the driver with DevRoot pinned to Root (internal/runtime/runtime.go), so the split-root omission can't occur on this path, and Root-as-devRoot was the function's pre-existing idiom — but since this change also creates a node, switched to f.driver.DevRoot as hardening. Behavior is unchanged in all current configurations.

ehfd:
@cdesiniotis For after v1.20.0.

henry118:
@ehfd  Thanks for putting this PR together. I definitely see why this change is needed.

As an alternative, do you think it would be simpler to just add the following directly in `nvidia-cdi-refresh.service`?

```
nvidia-ctk system create-device-nodes --control-devices --load-kernel-modules
```

ehfd:
@henry118 It could address it partially, but has some rough spots.

Most representatively, the first container injection won't work with `nvidia-ctk system create-device-nodes --control-devices --load-kernel-modules`; it will work from the second time.

ehfd:
@henry118 The footprint is now reduced greatly. The logic is baked into `ExecStart=/usr/bin/nvidia-ctk cdi generate` instead of `nvidia-ctk system create-device-nodes --control-devices --load-kernel-modules`.

ehfd:
Now, this PR also depends on https://github.com/NVIDIA/libnvidia-container/pull/388 merged to be fully fixed in both CDI and legacy pathways.

henry118:
> @henry118 It could address it partially, but has some rough spots.
> 
> Most representatively, the first container injection won't work with `nvidia-ctk system create-device-nodes --control-devices --load-kernel-modules`; it will work from the second time.

Could you please explain this a bit more? Is it because of startup races between docker and nvidia-cdi-refresh?

ehfd:
@henry118 I will answer in-depth tomorrow.

ehfd:
At boot, it is exactly a docker to nvidia-cdi-refresh race, but the race is structural, not an ordering oversight. Node absence is not a boot-only condition so that no systemd unit ordering can close it.

The below is an AI analysis.

 ### The mechanism                                                                                            
                                                                                                              
1. The driver never registers control nodes with devtmpfs. internal/system/devicenodes/devicenodes.go:57-60 : /dev/nvidia-modeset, /dev/nvidia-uvm* exist only if some userspace agent created them since the last boot or module unload. /dev resets every boot; nodes also vanish on any modprobe -r / driver reload.
2. The refresh service is unordered against every container engine. deployments/systemd/nvidia-cdi-refresh.service: Type=oneshot, WantedBy=multi-user.target, no After=/Requires= on docker/containerd/crio. A GPU container started early at boot (e.g. restart=always, or an immediate docker run) reaches the runtime before the oneshot has run any ExecStart — henry118's extra ExecStart included. First container loses; second (started after the oneshot) works. That is the boot instance of the race.
3. Default containers don't even use the on-disk spec. internal/info/auto.go:89 — mode = "auto" resolves to jit-cdi on NVML platforms. The spec is generated in-memory per container at container-start (internal/modifier/cdi.go:44-87 → newAutomaticCDISpecModifier), so /var/run/cdi/nvidia.yaml — the only artifact henry118's edit fixes — is irrelevant to the default path. Only node existence on the host matters to jit-cdi, and that's guaranteed only if the service's ExecStart already ran.
4. The path unit doesn't fire on the events that actually remove the nodes. deployments/systemd/nvidia-cdi-refresh.path watches only /lib/modules/%v/modules.dep{,.bin} and /usr/bin/nvidia-ctk — install/upgrade/kernel-change events. A driver reload (modprobe -r nvidia_uvm nvidia_modeset, driver update that reloads modules) deletes the nodes but touches none of those paths, so the service never re-runs. With henry118's approach, every container started from that moment until reboot/upgrade fails — not just the first.
5. Discovery is fail-open, so the failure is silent. internal/discover/mounts.go:66-68 — a missing node produces `Could not locate /dev/nvidia-modeset` at warning level and the spec is emitted without it. That's precisely pauljeary's #1982 log: boot run at 09:58:45 misses it (nothing had created it yet → broken Wayland/Vulkan in containers), manual restart at 10:00:42 finds it (the COSMIC session had triggered creation in between). Also note #1982's own workaround was systemctl disable nvidia-cdi-refresh.service — any fix living solely in that unit dies with it.                                                          
6. Why it must be in-process, not a prior command. devicenodes.go:46-49 — the UVM major is read from /proc/devices at constructor time and "only becomes available once the associated kernel modules have been loaded". Creation must be ordered after module load and immediately before discovery reads /dev. Baking it into nvidia-ctk cdi generate (generate.go:320) and the jit/graphics modifiers (cdi.go:187, graphics.go:37) makes that ordering guaranteed by construction; an external actor can only make the race window smaller, never zero.

Timeline under henry118's original proposal:
 ```                                                                                                          
boot ─┬─ multi-user.target ─┬─ nvidia-cdi-refresh.service (oneshot):
      │                     │     create-device-nodes → cdi generate  ← spec+nodes OK *after this point*
      │                     └─ docker.service (unordered, parallel)
      │                            └─ GPU container #1 (restart=always)  ← jit-cdi spec built NOW: no
modeset → FAIL
      │                                  GPU container #2 (later)         ← nodes exist → OK
      └─ driver reload mid-lifetime ──→ nodes gone, path unit NOT fired   ← containers #3..N all FAIL
 ```                                                                                                          
                                                                                                              
Not only docker↔refresh-service — that's the boot-time instance of it. The general condition is: any spec generation or container start that happens before some agent has created the control nodes since the driver was last (re)loaded gets a silently degraded spec.

Concretely, with the unit-only approach:                                                                   
1. nvidia-cdi-refresh.service is Type=oneshot, WantedBy=multi-user.target, with no ordering against docker/containerd/crio. A GPU container with restart=always (or an early docker run) can start before the oneshot's extra ExecStart runs. Since mode = "auto" resolves to jit-cdi on NVML platforms, that container builds its spec in-memory at start time and misses /dev/nvidia-modeset — first injection fails, the second one (after the oneshot) works. There's no systemd edge that fixes this: the service must also re-run on install/upgrade events, so engines can't simply be ordered After= it, and that wouldn't cover crio/podman/manual runs anyway.
2. Node absence isn't boot-only. The driver doesn't register control nodes with devtmpfs, so they also disappear on every driver reload/modprobe -r. The path trigger only watches modules.dep and the nvidia-ctk binary — reloads don't fire it — so from that moment all containers lose the node until reboot or upgrade, not just the first.
3. The unit is optional: tarball/non-systemd installs, and users who disable it (the exact workaround in #1982), get no coverage at all. Moving creation into cdi generate and the jit-cdi/graphics modifiers makes it happen in the same process, right before discovery reads /dev — and after the module load, which is what makes the dynamic UVM major readable from /proc/devices. It's idempotent (skips existing nodes) and best-effort (warn, never abort), so ordering no longer depends on any external actor having run first.

One caveat worth adding in review: the boot race for static-spec consumers (docker --device nvidia.com/gpu=…, mode = "cdi") additionally depends on henry118's ExecStart being ordered before the cdi generate line — silent breakage if appended after. In-process creation can't be misordered.

henry118:
> nvidia-cdi-refresh.service is Type=oneshot, WantedBy=multi-user.target, with no ordering against docker/containerd/crio. A GPU container with restart=always (or an early docker run) can start before the oneshot's extra ExecStart runs. Since mode = "auto" resolves to jit-cdi on NVML platforms, that container builds its spec in-memory at start time and misses /dev/nvidia-modeset — first injection fails, the second one (after the oneshot) works. There's no systemd edge that fixes this: the service must also re-run on install/upgrade events, so engines can't simply be ordered After= it, and that wouldn't cover crio/podman/manual runs anyway.

True, and this specific race is fixable with `Before=/After=` ordering or an RPM postinst hook. That's a packaging fix I'd rather make explicit than solve by moving node creation into the toolkit code.

> Node absence isn't boot-only. The driver doesn't register control nodes with devtmpfs, so they also disappear on every driver reload/modprobe -r. The path trigger only watches modules.dep and the nvidia-ctk binary — reloads don't fire it — so from that moment all containers lose the node until reboot or upgrade, not just the first.

Agreed this is real. And I'd rather close it with a udev rule that re-triggers node creation + CDI refresh. Although I'm not sure your real use case, but that's the automated equivalent of "whoever reloads the driver is responsible" without doing it manually.

> The unit is optional: tarball/non-systemd installs, and users who disable it (the exact workaround in https://github.com/NVIDIA/nvidia-container-toolkit/issues/1982), get no coverage at all. Moving creation into cdi generate and the jit-cdi/graphics modifiers makes it happen in the same process, right before discovery reads /dev — and after the module load, which is what makes the dynamic UVM major readable from /proc/devices. It's idempotent (skips existing nodes) and best-effort (warn, never abort), so ordering no longer depends on any external actor having run first.

IMO this is a packaging gap. Packagers must ensure the refresh path runs before container launches than fold node creation into cdi generate/jit-cdi. 

The main reason I'm not favoring the approach in this PR is that control devices are platform-dependent. Creating them universally means baking per-platform capability checks into the toolkit code. Keeping node creation as its own step (systemd unit + udev rule) lets distro maintainers customize the device creation logic per platform.

ehfd:
@henry118 Okay, I understand. I'll investigate further.

ehfd:
@henry118 Redesigned the PR based on your suggestions. However, the dependency on https://github.com/NVIDIA/libnvidia-container/pull/388 is now stronger.

deployments/systemd/nvidia-cdi-refresh.service
```diff
@@ -29,6 +32,13 @@ Environment=NVIDIA_CTK_CDI_OUTPUT_FILE_PATH=/var/run/cdi/nvidia.yaml
 EnvironmentFile=-/etc/nvidia-container-toolkit/nvidia-cdi-refresh.env
 ExecCondition=/bin/sh -c '/usr/bin/grep -qE "/(nvidia|nvidia-current)[.]ko" /lib/modules/%v/modules.dep || [ -e /dev/dxg ]'
 ExecStart=/bin/sh -c '/usr/bin/nvidia-smi -L || /usr/sbin/nvidia-smi -L || /usr/lib/wsl/lib/nvidia-smi -L'
+# Create any missing NVIDIA control device nodes (loading the kernel modules
+# they require) before generating the CDI specification; nothing else is
+# guaranteed to create them. Best-effort, and skipped on WSL where /dev/dxg is
+# used instead. If module loading fails (for example while a driver upgrade
+# has replaced the on-disk modules), still create the nodes for the modules
+# that are already loaded.
+ExecStart=-/bin/sh -c '[ -e /dev/dxg ] || /usr/bin/nvidia-ctk system create-device-nodes --control-devices --load-kernel-modules || /usr/bin/nvidia-ctk system create-device-nodes --control-devices'
```
henry118:
```suggestion
ExecStart=-/bin/sh -c '[ -e /dev/dxg ] || grep -q "^nvgpu " /proc/modules || /usr/bin/nvidia-ctk system create-device-nodes --control-devices --load-kernel-modules'
```

henry118: The suggested `nvgpu` check gate the command on some older Tegra platforms.

Also, why is the extra fallback without kernel module loading required here?

henry118 (commented):
@ehfd This approach LGTM overall.

@cdesiniotis @tariq1890 PTAL.

ehfd:
Applied, thank you. @henry118 

I used `/usr/bin/grep` to match the `ExecCondition=` line above; otherwise it's your suggestion as written. Verified on a live host: with nvgpu absent the grep exits 1 and the chain proceeds normally, and with an nvgpu module loaded the step short-circuits before `nvidia-ctk` and the unit still exits 0.

On the fallback: `--load-kernel-modules` calls `nvmodules.LoadAll()`, which modprobes `nvidia`, `nvidia-uvm` and `nvidia-modeset` and returns on the first failure, and `run()` propagates that error before `CreateNVIDIAControlDevices()` is reached. So a single failing modprobe means no control device nodes are created at all, including for the modules that are loaded and working. That is reachable in practice because modprobe returns 1 for a module that is loaded but no longer resolvable on disk (`FATAL: Module nvidia not found in directory /lib/modules/<kver>` — I reproduced this against a loaded module missing from `modules.dep`). The second invocation then creates whichever nodes have a major in `/proc/devices`.

However, if you think this should still be removed after my explanation and the case I am stating is niche, I can remove it.

ehfd:
Additional considerations across all three PRs: https://github.com/NVIDIA/libnvidia-container/pull/388#issuecomment-5352340514

deployments/systemd/nvidia-cdi-refresh.service
```diff
@@ -18,6 +18,9 @@ ConditionPathExists=|/usr/bin/nvidia-smi
 ConditionPathExists=|/usr/sbin/nvidia-smi
 ConditionPathExists=|/usr/lib/wsl/lib/nvidia-smi
 ConditionPathExists=/usr/bin/nvidia-ctk
+# Order container engines after this service so that the first container
+# started at boot does not race the device node creation and CDI refresh.
+Before=docker.service containerd.service crio.service
```
henry118: Just call out that we've historically had issues with adding dependencies, see #1735 .

So for this case, since it introduced a `Before=` rather than `After=`, in order to form a cycle among, say docker, us, other, either
1. docker has a `Before=us`; or
2. other has `After=docker` and `Before=us`;

The question is how likely are these two cases in practice. They seem unlikely to me.

cdesiniotis: I think these two scenarios are unlikely, so I would be open to adding this ordering dependency. However, can we propose this change in a separate PR? I would prefer if this PR was scoped to just addressing the race condition where CDI spec generation occurs before all NVIDIA device nodes exist and / or all NVIDIA driver kernel modules are loaded.

deployments/udev/99-nvidia-cdi-refresh.rules
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
cdesiniotis: nit: remove the year
```suggestion
# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
```

ehfd: Done.

deployments/systemd/nvidia-cdi-refresh.service
ehfd: Removed as well.

cdesiniotis (approved):

deployments/systemd/nvidia-cdi-refresh.service
ehfd: Following up on https://github.com/NVIDIA/nvidia-container-toolkit/pull/2019.

PR #1980, opened 2026-08-12
fix: NVIDIA X.Org driver modules mount fixes

### LINKED ISSUE #563: Dynamically expose NVIDIA X.Org X11 display server libraries and configure the container correctly

### LINKED ISSUE #1477: Automatically provisioning X11 and Wayland devices of GPU inside container?

ehfd:
Reviewers: @elezar @cdesiniotis @henry118

Fixes #1477
Fixes #563

## Summary

Graphics config files for the NVIDIA driver were mounted into the container at their host paths, so on hosts that do not install them under `/usr/share` (or `/etc` for the OpenCL ICD) they land where the loaders in the container never look. This mounts them at the canonical container locations, mirroring the existing Vulkan ICD handling, and injects the EGL X11 platform libraries and `nvidia-xconfig`.

The `create-xorg-config` CDI hook from the previous revision of this PR is gone. Per @cdesiniotis' review, the host's `xorg.conf.d` snippet is what carries the `ModulePath` when the driver installs the X.Org modules outside the default X.Org module path; mounting that snippet where the container's X server reads it is sufficient, and a host without one is treated as a host-side misconfiguration.

## Why This Exists

EGL vendor, EGL external platform and GLVND config files were mounted at their host paths. On hosts with non-FHS layouts these land where container glvnd never looks, forcing workarounds such as `__EGL_VENDOR_LIBRARY_FILENAMES`. The same applies to the `xorg.conf.d` snippet that makes the injected X.Org modules loadable, and to the OpenCL ICD.

The EGL X11 platform libraries (`libnvidia-egl-xcb`, `libnvidia-egl-xlib`) and their external platform config files were not discovered at all, even though the GBM and Wayland platform libraries next to them are.

## Resolution

- EGL/GLVND/X11 config files are mounted at `/usr/share/...` and the OpenCL ICD at `/etc/OpenCL/vendors`, mirroring the existing Vulkan ICD handling.
- The X.Org driver modules keep being mounted at their host paths, so an absolute `ModulePath` in the mounted snippet stays valid in the container, and modules installed to the default X.Org module path on the host are found at the same path in the container.
- `libnvidia-egl-xcb.so.*.*`, `libnvidia-egl-xlib.so.*.*` and their external platform JSON files are discovered and injected.
- `nvidia-xconfig` is injected when present on the host.
- All the mount discoverers in `graphics.go` are wrapped in `WithCache` (per @henry118's earlier review); the X.Org modules were previously discovered twice because `graphicsDriverLibraries` reads `Mounts()` from both `Mounts()` and `Hooks()`.

## Behavior Changes

- Containers requesting `graphics` or `display` receive the config files at canonical paths, the EGL X11 platform libraries and their config files, and `nvidia-xconfig` where installed.
- No new hooks are generated. The generated spec differs from `main` only by those mounts.
- Hosts without the relevant files see per-file discovery warnings, as with existing optional components.

## Implementation Summary

- Split the graphics config discovery out into `newGraphicsConfigsDiscoverer`, using `mountsToContainerPath` for `/usr/share` and `/etc`.
- Add the EGL X11 platform libraries to the graphics library discoverer and `nvidia-xconfig` to a binaries discoverer.
- Add `TestGraphicsConfigsDiscoverer`, covering both standard and non-standard host locations.

## Verification

`go test ./...` failing set is identical to the unmodified `main` baseline on the validation host (all failures are pre-existing environment failures in `nvidia-ctk-installer`, `internal/modifier` CSV mode and `pkg/nvcdi`); `golangci-lint run ./...` reports 0 issues; `gofmt` is clean.

Real-host validation on 2x Tesla P100, driver 580.178.04, Ubuntu 26.04, Docker 29.7.2, container `ubuntu:24.04` with only `xserver-xorg-core` and `mesa-utils` installed:

1. **Host with modules in the default X.Org module path** (`/usr/lib/xorg/modules/{drivers,extensions}`, Ubuntu packaging, snippet without a `ModulePath`): X.Org loads `nvidia_drv.so` and `libglxserver_nvidia.so` from `/usr/lib/xorg/modules/...`, `glxinfo -B` reports `Tesla P100-SXM2-16GB/PCIe/SSE2`. Unchanged from `main`.

2. **Host with modules outside the default X.Org module path** (`/usr/lib/x86_64-linux-gnu/nvidia/xorg`, with the host `10-nvidia.conf` carrying the matching `ModulePath`): the mounted snippet is what makes it work.

   ```
   (==) ModulePath set to "/usr/lib/xorg/modules"
   (**) OutputClass "nvidia" ModulePath extended to "/usr/lib/x86_64-linux-gnu/nvidia/xorg,/usr/lib/xorg/modules"
   (II) LoadModule: "nvidia"
   (II) Loading /usr/lib/x86_64-linux-gnu/nvidia/xorg/nvidia_drv.so
   (II) LoadModule: "glxserver_nvidia"
   (II) Loading /usr/lib/x86_64-linux-gnu/nvidia/xorg/libglxserver_nvidia.so
   (II) NVIDIA GLX Module  580.178.04
   ```

   `glxinfo -B` and the EGL X11 platform both report `Tesla P100-SXM2-16GB/PCIe/SSE2`.

   Note: `MatchDriver "nvidia-drm"` only matches when the DRM device nodes are present in the container, so this path requires `nvidia-drm` loaded on the host and the `display`/`graphics` capability. With `nvidia_drm` unloaded the OutputClass never matches and no `ModulePath` is applied.

3. `nvidia-xconfig` in the container generates a working BusID `xorg.conf`; `libEGL_nvidia` dlopens the injected `libnvidia-egl-xcb.so.1`/`libnvidia-egl-xlib.so.1` when their config files are present (verified with `LD_DEBUG=libs`). On 580 the built-in X11 platform still services `EGL_PLATFORM_X11`, so injecting them is about keeping the container's EGL platform set matching the host's rather than fixing an observed failure today.

4. Legacy-mode parity (`mode = "legacy"`): `graphics` receives the same set as CDI mode; `compute,utility` receives none of it, unchanged from `main`.

### DIFF
--- a/internal/discover/graphics.go
+++ b/internal/discover/graphics.go
@@ -55,29 +55,58 @@ func NewGraphicsMountsDiscoverer(logger logger.Interface, driver *root.Driver, h
 		return nil, fmt.Errorf("failed to construct discoverer for graphics libraries: %w", err)
 	}
 
-	configs := NewMounts(
+	binaries := NewMounts(
 		logger,
-		driver.Configs(),
+		lookup.NewExecutableLocator(logger, driver.Root),
 		driver.Root,
 		[]string{
-			"glvnd/egl_vendor.d/10_nvidia.json",
-			"egl/egl_external_platform.d/15_nvidia_gbm.json",
-			"egl/egl_external_platform.d/10_nvidia_wayland.json",
-			"egl/egl_external_platform.d/09_nvidia_wayland2.json",
-			"nvidia/nvoptix.bin",
-			"X11/xorg.conf.d/10-nvidia.conf",
-			"X11/xorg.conf.d/nvidia-drm-outputclass.conf",
-			"OpenCL/vendors/nvidia.icd",
+			"nvidia-xconfig",
 		},
 	)
 
 	discover := Merge(
 		libraries,
-		configs,
+		binaries,
+		newGraphicsConfigsDiscoverer(logger, driver),
 		newVulkanConfigsDiscover(logger, driver),
 	)
 
 	return discover, nil
+}
+
+// newGraphicsConfigsDiscoverer creates a discoverer for graphics-related config
+// files such as the EGL vendor and external platform ICD files.
+// The config files are mounted at the standard locations in the container so
+// that they are discovered by the loaders in the container even if they are
+// installed at non-standard locations on the host.
+func newGraphicsConfigsDiscoverer(logger logger.Interface, driver *root.Driver) Discover {
+	shareConfigs := WithCache(&mountsToContainerPath{
+		logger:  logger,
+		locator: driver.Configs(),
+		required: []string{
+			"glvnd/egl_vendor.d/10_nvidia.json",
+			"egl/egl_external_platform.d/15_nvidia_gbm.json",
+			"egl/egl_external_platform.d/10_nvidia_wayland.json",
+			"egl/egl_external_platform.d/09_nvidia_wayland2.json",
+			"egl/egl_external_platform.d/20_nvidia_xcb.json",
+			"egl/egl_external_platform.d/20_nvidia_xlib.json",
+			"nvidia/nvoptix.bin",
+			"X11/xorg.conf.d/10-nvidia.conf",
+			"X11/xorg.conf.d/nvidia-drm-outputclass.conf",
+		},
+		containerRoot: "/usr/share",
+	})
+
+	etcConfigs := WithCache(&mountsToContainerPath{
+		logger:  logger,
+		locator: driver.Configs(),
+		required: []string{
+			"OpenCL/vendors/nvidia.icd",
+		},
+		containerRoot: "/etc",
+	})
+
+	return Merge(shareConfigs, etcConfigs)
 }
 
 // newVulkanConfigsDiscover creates a discoverer for vulkan ICD files.
@@ -101,12 +130,12 @@ func newVulkanConfigsDiscover(logger logger.Interface, driver *root.Driver) Disc
 	case "arm64":
 		required = append(required, "vulkan/icd.d/nvidia_icd.aarch64.json")
 	}
-	return &mountsToContainerPath{
+	return WithCache(&mountsToContainerPath{
 		logger:        logger,
 		locator:       locator,
 		required:      required,
 		containerRoot: "/etc",
-	}
+	})
 }
 
 type graphicsDriverLibraries struct {
@@ -135,11 +164,14 @@ func newGraphicsLibrariesDiscoverer(logger logger.Interface, driver *root.Driver
 		driver.Libraries(),
 		driver.Root,
 		[]string{
-			// The libnvidia-egl-gbm and libnvidia-egl-wayland libraries do not
-			// have the RM version. Use the *.* pattern to match X.Y.Z versions.
+			// The EGL platform libraries such as libnvidia-egl-gbm and
+			// libnvidia-egl-wayland do not have the RM version. Use the *.*
+			// pattern to match X.Y.Z versions.
 			"libnvidia-egl-gbm.so.*.*",
 			"libnvidia-egl-wayland.so.*.*",
 			"libnvidia-egl-wayland2.so.*.*",
+			"libnvidia-egl-xcb.so.*.*",
+			"libnvidia-egl-xlib.so.*.*",
 			// We include the following libraries to have them available for
 			// symlink creation below:
 			// If CDI injection is used, these should already be detected as:
@@ -151,6 +183,10 @@ func newGraphicsLibrariesDiscoverer(logger logger.Interface, driver *root.Driver
 		},
 	)
 
+	// The X.Org driver modules are mounted at their host paths. This keeps any
+	// ModulePath specified in the xorg.conf.d config file that is mounted into
+	// the container valid and means that modules installed to the default X.Org
+	// module path on the host are also found there in the container.
 	xorgLibraries := NewMounts(
 		logger,
 		lookup.NewFileLocator(
@@ -231,7 +267,7 @@ func (d graphicsDriverLibraries) Hooks() ([]Hook, error) {
 		return nil, nil
 	}
 
-	hook := d.hookCreator.Create("create-symlinks", links...)
+	hook := d.hookCreator.Create(CreateSymlinksHook, links...)
 
 	return hook.Hooks()
 }
--- a/internal/discover/graphics_test.go
+++ b/internal/discover/graphics_test.go
@@ -17,6 +17,7 @@
 package discover
 
 import (
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -25,6 +26,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/devices"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/lookup/root"
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/test"
 )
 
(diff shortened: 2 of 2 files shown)


ehfd:
@cdesiniotis For after v1.20.0.

internal/discover/graphics.go
```diff
@@ -55,29 +55,58 @@ func NewGraphicsMountsDiscoverer(logger logger.Interface, driver *root.Driver, h
...
+	return discover, nil
+}
+
+// newGraphicsConfigsDiscoverer creates a discoverer for graphics-related config
+// files such as the EGL vendor and external platform ICD files.
+// The config files are mounted at the standard locations in the container so
+// that they are discovered by the loaders in the container even if they are
+// installed at non-standard locations on the host.
+func newGraphicsConfigsDiscoverer(logger logger.Interface, driver *root.Driver) Discover {
+	shareConfigs := &mountsToContainerPath{
```
henry118: Can we wrap all the mounts with https://github.com/NVIDIA/nvidia-container-toolkit/blob/02ccb62b0d828daaf3f4714f5d636773bfcf06aa/internal/discover/cache.go#L34?

cmd/nvidia-cdi-hook/create-xorg-config/create-xorg-config.go
```diff
@@ -0,0 +1,154 @@
...
+    ModulePath "%v"
+EndSection
+`, modulePath)
+}
+
+// createConfigInRoot creates the specified file in the specified root with the
+// specified contents.
+// Parent directories that do not exist are created.
+func createConfigInRoot(root *os.Root, path string, contents string) error {
+	relativePath := strings.TrimPrefix(path, "/")
```
henry118: what if `path == "/"`?

internal/discover/graphics.go
```diff
@@ -151,29 +191,43 @@ func newGraphicsLibrariesDiscoverer(logger logger.Interface, driver *root.Driver
...
+// xorgModulesContainerDir returns the container directory into which the X.Org
+// driver modules for the NVIDIA driver are mounted.
+// The directory is located below the primary driver library directory so that
+// the modules are available at a predictable path in the container regardless
+// of where they are installed on the host.
+func xorgModulesContainerDir(libRoots []string) string {
+	if len(libRoots) == 0 {
+		return "/usr/lib/nvidia/xorg"
+	}
+	return filepath.Join(libRoots[0], "nvidia", "xorg")
```
henry118: any particular reason for picking the 1st element (but ignore the rest) here?

henry118 (commented):
Thanks @ehfd. overall lgtm. left a few comments, ptal.

internal/discover/graphics.go
ehfd: The modules have to land in a single directory because that directory is what we pass to X.Org as ModulePath. The first entry is the primary driver library directory, so the container path mirrors the host layout used by the driver packages (`{{ libdir }}/nvidia/xorg`) — on a Debian-style host it's an identity mapping. The remaining entries are alternative library directories (32-bit compat, for example) that don't contain X.Org modules, so mounting into them would just duplicate the same files.

cmd/nvidia-cdi-hook/create-xorg-config/create-xorg-config.go
ehfd: It failed with a confusing `openat : empty` path. Now rejected in `validateFlags`, and `createConfigInRoot` also cleans the path and refuses one that doesn't name a file.

Fixed.

internal/discover/graphics.go
ehfd: All four instances in graphics.go are wrapped now; the xorgLibraries one in particular was being discovered twice since graphicsDriverLibraries reads Mounts() from both Mounts() and Hooks().

Fixed.

henry118:
This LGTM now. @cdesiniotis can you also ptal?

cdesiniotis:
@ehfd thanks for your contribution. I am not too keen on introducing a CDI hook for this. Wouldn't the host's Xorg conf file (i.e. `/usr/share/X11/xorg.conf.d/10-nvidia.conf`) handle the scenario where the NVIDIA-specific libraries are installed at non-standard locations? Would the issue you are encountering go away if the `xorg.conf.d` file was also mounted into the container?

You mention a use case where the `xorg.conf.d` file does not exist on the host, but in my opinion, that is a misconfiguration and this type of scenario is out-of-scope for the NVIDIA Container Toolkit to address. A prerequisite is that the host has a functioning (and complete) driver installation. If a specific application does not work on the host, then we cannot expect it to work in a container.

ehfd:
@cdesiniotis I understand your perspective. I will find a solution which works for all parties.

ehfd:
@cdesiniotis @henry118 

I've reworked the PR around your suggestion. The `create-xorg-config` hook is gone, and the PR now relies on the host's `xorg.conf.d` snippet being mounted where the container's X server reads it.

I validated this on a 2x Tesla P100 host (driver 580.178.04, Ubuntu 26.04, Docker 29.7.2) with a stock `ubuntu:24.04` container that only has `xserver-xorg-core` and `mesa-utils` installed. With the X.Org modules deliberately installed outside the default X.Org module path (`/usr/lib/x86_64-linux-gnu/nvidia/xorg`) and the host `10-nvidia.conf` carrying the matching `ModulePath`, the mounted snippet is exactly what makes it work:

```
(==) ModulePath set to "/usr/lib/xorg/modules"
(**) OutputClass "nvidia" ModulePath extended to "/usr/lib/x86_64-linux-gnu/nvidia/xorg,/usr/lib/xorg/modules"
(II) LoadModule: "nvidia"
(II) Loading /usr/lib/x86_64-linux-gnu/nvidia/xorg/nvidia_drv.so
(II) LoadModule: "glxserver_nvidia"
(II) Loading /usr/lib/x86_64-linux-gnu/nvidia/xorg/libglxserver_nvidia.so
(II) NVIDIA GLX Module  580.178.04
```

`glxinfo -B` and the EGL X11 platform both report the P100. One prerequisite worth recording for anyone who finds this later: `MatchDriver "nvidia-drm"` only matches when the DRM device nodes are in the container, so this needs `nvidia-drm` loaded on the host and the `display`/`graphics` capability requested. With `nvidia_drm` unloaded the OutputClass never matches and no `ModulePath` is applied — which lines up with your point that the host needs a functioning driver installation.

What's left in the PR:

- Config files (EGL vendor, EGL external platform, GLVND, `xorg.conf.d`, OpenCL ICD) mounted at the canonical container locations rather than the host paths, mirroring the existing Vulkan ICD handling. This is what makes the `xorg.conf.d` snippet land where the container's X server reads it on hosts that do not install it under `/usr/share`, and it removes the need for `__EGL_VENDOR_LIBRARY_FILENAMES`-style workarounds for the same reason.
- X.Org modules still mounted at their host paths — unchanged from `main` — so an absolute `ModulePath` in the mounted snippet stays valid.
- `libnvidia-egl-xcb` / `libnvidia-egl-xlib` and their external platform JSON files injected alongside the GBM and Wayland platform libraries. `libEGL_nvidia` does `dlopen` them once the JSON files are present (confirmed with `LD_DEBUG=libs`), so this is about the container's EGL platform set matching the host's.
- `nvidia-xconfig` injected when present on the host (preferred in #563; the container loses configuration-controlling mechanisms without this).

henry118 (commented):
LGTM

ehfd:
Additional considerations across all three PRs: https://github.com/NVIDIA/libnvidia-container/pull/388#issuecomment-5352340514

ehfd:
@cdesiniotis Can you review this too?

cdesiniotis (approved):
LGTM. @ehfd thanks for your contribution and patience during reviews.

ehfd:
@cdesiniotis @henry118 Thank you for your time investment for the community!

PR #1995, opened 2026-08-13
add new CDI hook to set CUDA memory limits [DO-NOT-MERGE]

tariq1890:
This commit introduces a new CDI hook binary called `apply-cuda-memory-limits`. The purpose of this CDI hook is to apply the desired memory limits to a GPU device. 

It reads the following envars from a container spec to fetch the input parameters for the CDI hook

- NVIDIA_GPU_MEMORY_REQUEST(S)
- NVIDIA_GPU_MEMORY_LIMIT(S)

Under the covers, the CDI hooks invokes the new `NVML SetMemoryLimits_v1` method (introduced in CUDA 13.4) to apply the desired CUDA memory limit values to the cgroup of the GPU container in question

### DIFF
--- a/cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
+++ b/cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
@@ -1,0 +1,185 @@
+/**
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package cudamemorylimits
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+	"github.com/urfave/cli/v3"
+
+	cgroupinfo "github.com/NVIDIA/nvidia-container-toolkit/internal/info/cgroup"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/logger"
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/oci"
+	"github.com/NVIDIA/nvidia-container-toolkit/pkg/lookup"
+)
+
+type command struct {
+	logger logger.Interface
+}
+
+type config struct {
+	driverRoot    string
+	gpuIds        []string
+	containerSpec string
+}
+
+func NewCommand(logger logger.Interface) *cli.Command {
+	c := command{
+		logger: logger,
+	}
+	return c.build()
+}
+
+func (m command) build() *cli.Command {
+	cfg := config{}
+
+	c := cli.Command{
+		Name:  "apply-cuda-memory-limits",
+		Usage: "Set the soft and hard limits of CUDA memory usage on a GPU device in the container.",
+		Before: func(ctx context.Context, cmd *cli.Command) (context.Context, error) {
+			return ctx, m.validateFlags(cmd, &cfg)
+		},
+		Action: func(ctx context.Context, cmd *cli.Command) error {
+			return m.run(cmd, &cfg)
+		},
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:        "driver-root",
+				Usage:       "Specify the driver root",
+				Destination: &cfg.driverRoot,
+			},
+			&cli.StringSliceFlag{
+				Name:        "gpu-id",
+				Usage:       "Specify the UUID of the GPU",
+				Destination: &cfg.gpuIds,
+			},
+			&cli.StringFlag{
+				Name:        "container-spec",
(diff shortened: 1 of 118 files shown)


cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
```diff
@@ -0,0 +1,185 @@
+/**
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cdesiniotis, line 2: nit: remove the year from the copyright header

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
```diff
@@ -0,0 +1,185 @@
...
+			return ctx, m.validateFlags(cmd, &cfg)
+		},
+		Action: func(ctx context.Context, cmd *cli.Command) error {
+			return m.run(cmd, &cfg)
+		},
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:        "driver-root",
+				Usage:       "Specify the driver root",
+				Destination: &cfg.driverRoot,
```
cdesiniotis, line 68: Question -- do we need to define a default value for this field, or validate that it is specified in `validateFlags()`?

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
```diff
@@ -0,0 +1,185 @@
...
+	if !ok1 {
+		memReqStr, ok1 = fs.LookupEnv("NVIDIA_GPU_MEMORY_REQUEST")
+	}
+	memLimitStr, ok2 := fs.LookupEnv("NVIDIA_GPU_MEMORY_LIMITS")
+	if !ok2 {
+		memLimitStr, ok2 = fs.LookupEnv("NVIDIA_GPU_MEMORY_LIMIT")
+	}
+	if !ok1 || !ok2 {
+		return nil
+	}
```
cdesiniotis, line 118: Question -- does this mean that it is invalid to only specify one of these envvars (but not the other)?

internal/info/cgroup/cgroup_path.go
```diff
@@ -0,0 +1,137 @@
+/**
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cdesiniotis, line 2: nit: let's remove the year from the copyright header

internal/info/cgroup/cgroup_path.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return s.Type == v2FsMagicNumber
+}
+
+func GetAbsolutePath(spec specs.Spec) (string, error) {
+	c := &cgroups.Cgroup{}
+	myCgroupPath := spec.Linux.CgroupsPath
+	parts := strings.Split(myCgroupPath, ":")
+	if len(parts) != 3 {
+		return "", fmt.Errorf("expected cgroupsPath to be of format \"slice:prefix:name\" for systemd cgroups, got %q instead", myCgroupPath)
```
cdesiniotis, line 51: Question -- does this only work when systemd is managing cgroups? What about cgroupfs?

internal/info/cgroup/cgroup_path.go
cdesiniotis, line 1: Let's add some unit tests for this file.

pkg/nvcdi/full-gpu-nvml.go
```diff
@@ -173,11 +173,17 @@ func (l *fullGPUDeviceSpecGenerator) newFullGPUDiscoverer(d device.Device) (disc
 		deviceNodes,
 	)
 
+	cudaMemoryLimitsHook, err := (*nvcdilib)(l.nvmllib).newCudaMemoryLimits(d)
```
cdesiniotis, line 176: Just curious, is the NVML API for setting memory limits only supported on full GPUs? What about MIG devices?

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
tariq1890: Good question. What would you suggest here?

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
tariq1890: I am open to changing this. We can just limit it to 2 out of the 4 env vars here (maybe drop the plurals?) for the sake of simplicity here.

cdesiniotis: My question was more so out of curiosity. Is it valid to only specify a memory limit and not a request?

tariq1890: Yes, I believe it is. If unspecified, the value stays the same as it was

internal/info/cgroup/cgroup_path.go
tariq1890: Yes, this PR only supports `systemd` as the cgroup manager in its current state. I think `cgroupfs` needs to be supported too.

internal/info/cgroup/cgroup_path.go
tariq1890: will do

pkg/nvcdi/full-gpu-nvml.go
tariq1890: I will need to find out

cdesiniotis:
Closing as superseded by https://github.com/NVIDIA/nvidia-container-toolkit/pull/2075

PR #2012, opened 2026-08-19
Ensure backport commits have verified signatures.

henry118:
Branch protection requires verified commit signatures, but commits made via `git push` from a workflow are never signed. 

Re-create each new commit through the Git Data API instead: GitHub automatically signs commits created this way, so the resulting chain shows as "Verified" without needing to manage a GPG/SSH key for the bot.

Verified on testing repo: https://github.com/henry118/playground/pull/4

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/.github/scripts/backport.js
+++ b/.github/scripts/backport.js
@@ -100,7 +100,42 @@ for (const targetBranch of branches) {
     // Push the backport branch (force to handle updates)
     core.info(`Pushing ${backportBranch} to origin`);
     execSync(`git push --force-with-lease origin ${backportBranch}`, { stdio: 'inherit' });
-    
+
+    // Re-create each new commit through the Git Data API so the resulting chain shows as "Verified"
+    core.info(`Re-creating commits via the Git Data API to get verified signatures`);
+    const newCommitShas = execSync(`git log --format=%H ${targetBranch}..${backportBranch}`, { encoding: 'utf-8' })
+      .trim().split('\n').filter(Boolean).reverse(); // oldest -> newest
+
+    const { data: baseRef } = await github.rest.git.getRef({
+      owner: context.repo.owner,
+      repo: context.repo.repo,
+      ref: `heads/${targetBranch}`
+    });
+    let parentSha = baseRef.object.sha;
+
+    for (const sha of newCommitShas) {
+      const treeSha = execSync(`git rev-parse ${sha}^{tree}`, { encoding: 'utf-8' }).trim();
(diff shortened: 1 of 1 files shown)


tariq1890:
Thanks @henry118 for making this change.

This could also be applied to other repos that use the cherry-pick bot

.github/scripts/backport.js
```diff
@@ -97,10 +97,60 @@ for (const targetBranch of branches) {
...
+    // commit through the Git Data API instead: GitHub automatically signs
+    // commits created this way, so the resulting chain shows as "Verified"
+    // without needing to manage a GPG/SSH key for the bot.
+    //
+    // IMPORTANT: GitHub only auto-signs bot commits created via this endpoint
+    // when the request has NO custom author, committer, or signature info.
+    // Setting author/committer to the original PR author would silently
+    // disable signing, so these commits are attributed to github-actions[bot]
+    // instead; original authorship is preserved via the `-x` cherry-pick
+    // trailer already in the commit message, and in the PR body.
```
tariq1890: I've noticed that agent-generated code tends to be very verbose with the comments. I imagine this is to help the prompter understand the code better.

In general, I think it's better to keep comments to a minimum as the code should be self-explanatory. If there are important details/context that needs to be stored, we can also add them to the commit description so that context is preserved.

.github/scripts/backport.js
```diff
@@ -97,10 +97,60 @@ for (const targetBranch of branches) {
         }
       }
     }
-    // Push the backport branch (force to handle updates)
+    // Push the backport branch (force to handle updates). This uploads the
+    // tree/blob objects to GitHub; the commits themselves are unsigned at
+    // this point and will be replaced below.
```
tariq1890:
```
     // This uploads the
    // tree/blob objects to GitHub; the commits themselves are unsigned at
    // this point and will be replaced below.
```
I suggest removing this. This is more or less implied.

.github/scripts/backport.js
henry118: Updated

.github/scripts/backport.js
henry118: Updated

.github/scripts/backport.js
tariq1890: Thanks

tariq1890 (approved):

cdesiniotis (approved):

PR #2015, opened 2026-08-21
use Masterminds/semver instead of x/mod

tariq1890:
This PR replaces `golang.org/x/mod` with the `github.com/Masterminds/semver/v3` module for all semver processing in the repo.

`Masterminds/semver` is a more lightweight dependency and it also supports the parsing of semantic version strings with/without the `"v"` prefix. This removes the need for any explicit prefixing of version strings just so that they are usable in `x/mod`'s semver methods.

### DIFF
--- a/cmd/nvidia-container-runtime-hook/container_config.go
+++ b/cmd/nvidia-container-runtime-hook/container_config.go
@@ -7,8 +7,8 @@ import (
 	"os"
 	"path"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/opencontainers/runtime-spec/specs-go"
-	"golang.org/x/mod/semver"
 
 	"github.com/NVIDIA/nvidia-container-toolkit/internal/config/image"
 )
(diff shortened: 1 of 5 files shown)


cmd/nvidia-container-runtime-hook/container_config.go
```diff
@@ -107,8 +107,12 @@ func (s *Spec) GetCapabilities() []string {
 	var caps []string
 	// If v1.0.0-rc1 <= OCI version < v1.0.0-rc5 parse s.Process.Capabilities as:
 	// github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/specs-go/config.go#L30-L54
-	rc1cmp := semver.Compare("v"+*s.Version, "v1.0.0-rc1")
```
henry118, line 110: The new package can handle inputs w/ or w/o "v".  I think all the work we do on taking care of "v" can be dropped?

cmd/nvidia-container-runtime-hook/container_config.go
```diff
@@ -107,8 +107,12 @@ func (s *Spec) GetCapabilities() []string {
...
-	rc5cmp := semver.Compare("v"+*s.Version, "v1.0.0-rc5")
+	sv, err := semver.NewVersion("v" + *s.Version)
+	if err != nil {
+		log.Panicln("could not parse OCI spec version:", err)
```
henry118: Is it safe t panic here? it seems the old version would not panic on parse failure

tariq1890: You're right that it won't panic on parse failure but instead return `-1`. It does technically introduce a new failure path.

cmd/nvidia-container-runtime-hook/container_config.go
tariq1890: You're right. let me see what changes I can make here.

cmd/nvidia-container-runtime-hook/container_config.go
tariq1890: Let me see how this pans out after I address your other comment.

tariq1890: @henry118 Instead of panic'ing, I will default to `0.0.0` version. This will mirror the old behaviour where `x/mod/semver.Compare` considers the invalid semver as "lesser than".

henry118 (approved):
LGTM on green

cmd/nvidia-container-runtime-hook/container_config.go
tariq1890: I've now removed any of the version input pre-processing which has been made redundant by `Mastermimds/semver`

PR #2018, opened 2026-08-21
Update golang toolchain

henry118:
Go 1.25 is [EOL](https://endoflife.date/go). 

* Bump the minimum required toolchain to 1.26
* Bump CI version (build/test/release) to 1.27

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/NVIDIA/nvidia-container-toolkit
 
-go 1.25.0
+go 1.26.0
 
 require (
 	github.com/Masterminds/semver/v3 v3.5.0
--- a/cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
+++ b/cmd/nvidia-cdi-hook/cudacompat/cudacompat.go
@@ -155,7 +155,7 @@ func (m command) getContainerForwardCompatDir(containerRoot *root, o *options) (
 
 	libs, err := containerRoot.globFiles(filepath.Join(o.cudaCompatContainerRoot, "libcuda.so.*.*"))
 	if err != nil {
-		m.logger.Warningf("Failed to find CUDA compat library: %w", err)
+		m.logger.Warningf("Failed to find CUDA compat library: %v", err)
 		return "", nil
 	}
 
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/NVIDIA/nvidia-container-toolkit
 
-go 1.25.0
+go 1.26.0
```
tariq1890, line 3: My understanding is that this represents the minimum go version required to compile the module. This is usually updated during a dependency update as the latest of the min-go versions among the dependencies would end up being the minimum go version needed to compile.

henry118: It looks like the following repos are all on 1.26.x now:
* https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/main/go.mod#L3
* https://github.com/NVIDIA/gpu-operator/blob/main/go.mod#L3
* https://github.com/NVIDIA/k8s-device-plugin/blob/main/go.mod#L3

Anything else we should check?

henry118: [This link](https://github.com/search?q=org%3ANVIDIA+path%3A%2Fgo%5C.mod%24%2F+%22github.com%2FNVIDIA%2Fnvidia-container-toolkit%22&type=code) gives all repos.

tariq1890: So, if you look at this [commit](https://github.com/NVIDIA/gpu-operator/commit/3d4b4c04c6937f992bd7eda31992d031b0eb64e3) in gpu-operator, you'll see that it was updated as part of the `operator-framework` dependency update. The minimum version required to compile that `operator-framework` module version for 1.26.3 and gpu-operator inherited that. The updates to the go versions in the `go.mod` is always associated with a dependency update

tariq1890: Same goes for the k8s-device-plugin. From this [commit](https://github.com/NVIDIA/k8s-device-plugin/commit/0ee1f79287fa1378300524b052cc25b76a3f1082), you'll see that the go version in go.mod was updated after the dependency update of `k8s.io`

tariq1890: I'd argue that lower the min-go version, the better. It's a sign of how the toolkit isn't as influenced by external deps as the other repos are, and that the go code is quite backwards compatible.

henry118: The two example you showed are opened by dependabot. It has to update the repo's toolchain because their dependency forced them.

We are not the case here since we're lower level. If you look at the dependencies of the other two cases, it's a regular manual bump too, e.g. https://github.com/operator-framework/api/pull/497

I agree lower min-go version is better. Hence we use N-1 for our toolchain and N for CI. Note that Golang only supports two versions (N, N-1) at a time. Older versions are not maintained and tend to have CVEs and bugs. It's generally a good practice to keep the toolchain up-to-date.

tariq1890: Makes sense.

tariq1890 (approved):

cdesiniotis (approved):

PR #2031, opened 2026-08-28
fix: update all CDI hooks to fail for invalid container root paths

cdesiniotis:


henry118 (commented):
What about the following?

1. `"//"`, `"/."`, `"/foo/.."`
2. `/container/root` -> `/` (link)
3. `mount --bind / /container/root` (bind mount)

also runc validates the rootfs too, but it does not fail on root though.
https://github.com/opencontainers/runc/blob/cb470207c6f48904e11418a8fef3c37aeb7e482e/libcontainer/configs/validate/validator.go#L55-L72

cdesiniotis:
> What about the following?
> 
> 1. `"//"`, `"/."`, `"/foo/.."`
> 2. `/container/root` -> `/` (link)
> 3. `mount --bind / /container/root` (bind mount)
> 
> also runc validates the rootfs too, but it does not fail on root though. https://github.com/opencontainers/runc/blob/cb470207c6f48904e11418a8fef3c37aeb7e482e/libcontainer/configs/validate/validator.go#L55-L72

Fair point. If runc does not fail on `/` then one could make the argument we shouldn't either. If anything, we should update https://github.com/NVIDIA/nvidia-container-toolkit/blob/7df27d08ebe6de4215bcf85dfb54166f8bcf69f0/internal/oci/state.go#L82-L95 to check for symlinks (like runc does in https://github.com/opencontainers/runc/blob/cb470207c6f48904e11418a8fef3c37aeb7e482e/libcontainer/configs/validate/validator.go#L55-L72. WDYT?

The `""` case we already handle when checking for an absolute vs relative path.

henry118:
runc performs the validation when a container is created, before invoking oci hooks. 

but i am not against the extra validation either, esp if we don't want to depend on runtime's implementation details.

henry118:
> If runc does not fail on / then one could make the argument we shouldn't either.

Agreed

cdesiniotis:
Closing as no longer planned. May open a follow up based on the discussion here.

Issue #2037, opened 2026-08-31
[Feature]: Support injecting POSIX rlimits (e.g. RLIMIT_MEMLOCK for RDMA) into the OCI spec via nvidia-container-runtime config


liuxu623:
### Problem

RDMA workloads (GPUDirect RDMA, NCCL over IB/RoCE, UCX) need to pin buffer pools that run to gigabytes. Kubernetes never sets `RLIMIT_MEMLOCK`, so every container inherits it from the container runtime daemon — with containerd under systemd that is the 8 MiB `LimitMEMLOCK` default. `ibv_reg_mr` then fails with `ENOMEM` and NCCL silently falls back to TCP sockets.

The standard advice — add the `IPC_LOCK` capability to the pod — **only works for containers that run as root**. For a non-root container, capabilities added via `securityContext.capabilities.add` land in the bounding set but never in the effective/permitted set (Kubernetes has no ambient-capability support), so `CAP_IPC_LOCK` is silently inert.

Measured on Kubernetes 1.36 / containerd 2.3.3 / toolkit v1.19.1 (via GPU Operator), same image, same node:

| container user | `CapEff` | `ulimit -l` | `mlock` test |
|---|---|---|---|
| uid 1000 + `IPC_LOCK` added | `0000000000000000` (cap lost) | 8192 KB | 64 MiB → `ENOMEM` |
| uid 0 + `IPC_LOCK` added | `a80465fb` (cap effective) | 8192 KB | 1 GiB → OK |

Since GPU images increasingly run as non-root, there is currently **no per-runtime way** to give GPU containers a usable memlock limit:

- `LimitMEMLOCK=infinity` on the containerd systemd unit works, but is node-wide — every container on the node gets it, not just GPU workloads.
- containerd's per-runtime `base_runtime_spec` could scope it to the nvidia runtime, but it replaces the entire default OCI spec (fragile across containerd upgrades), and the toolkit option that could set it — `RUNTIME_CONFIG_OVERRIDE` — has been non-functional since the v1.17 refactor (see "Related regression" below).

### Proposal

Let `nvidia-container-runtime` — which is already an OCI-spec-rewriting shim — optionally inject POSIX rlimits into the spec:

```toml
[nvidia-container-runtime]
rlimits = ["memlock=unlimited"]        # or e.g. "memlock=1073741824:unlimited"
```

Implementation sketch (happy to submit a PR, it is small and self-contained):

[...]

PR #2038, opened 2026-08-31
Add support for injecting POSIX rlimits via nvidia-container-runtime config

liuxu623:
Addresses #2037.

## Motivation

Kubernetes provides no way to set per-container POSIX rlimits (kubernetes/kubernetes#3595), and the common advice for RDMA workloads — adding the `IPC_LOCK` capability — is silently ineffective for non-root containers: capabilities added via `securityContext` land in the bounding set but never in a non-root container's effective set, so `ibv_reg_mr` still fails with `ENOMEM` against the inherited `RLIMIT_MEMLOCK` (typically the 8 MiB systemd default of the container runtime daemon). Raising the limit on the daemon instead changes it for every container on the host, and containerd's per-runtime `base_runtime_spec` replaces the entire default spec, which is fragile across containerd upgrades.

`nvidia-container-runtime` already rewrites the OCI spec per container and is naturally scoped to exactly the workloads that need pinned memory, so this adds an opt-in config option there. See #2037 for the full analysis.

## What this PR does

New `rlimits` option in the runtime config:

```toml
[nvidia-container-runtime]
rlimits = ["memlock=unlimited"]
```

- Entries have the form `NAME=SOFT[:HARD]`. Names are case-insensitive with an optional `RLIMIT_` prefix; values are non-negative integers or `unlimited`/`infinity` (mapped to `RLIM_INFINITY`). `HARD` defaults to `SOFT`; `SOFT > HARD` and duplicate types are rejected.
- Implemented as an `oci.SpecModifier` (`internal/modifier/rlimit.go`) that merges the configured entries into `spec.Process.Rlimits`: an entry replaces an existing rlimit of the same type, other entries in the spec are preserved.
- Applies in all runtime modes (legacy/csv/cdi/jit-cdi) and is deliberately **not** gated on requested devices: workloads that need the limits (e.g. RDMA-only containers using an `rdma/*` resource) do not necessarily request GPUs.
- The toolkit container accepts `--nvidia-container-runtime.rlimits` / env `NVIDIA_CONTAINER_RUNTIME_RLIMITS` and renders the option into the installed `config.toml`, so GPU Operator users can enable it with no operator changes:

```yaml
toolkit:
  env:
  - name: NVIDIA_CONTAINER_RUNTIME_RLIMITS
    value: memlock=unlimited
```

## Notes

- Default behavior is unchanged: the option is `omitempty` and the modifier is skipped when unset (`nvidia-ctk config default` output is identical).
- This only takes effect where `nvidia-container-runtime` is the runtime (RuntimeClass or default runtime). The hook-only path (`docker --gpus`) cannot alter rlimits, since a prestart hook runs after the process limits are set.
- Also adds a `[]string` case to the toolkit installer's optional-config switch; the existing `[]string`-typed options previously fell through to the "Unexpected type" warning.

## Testing

- Unit tests for entry parsing (valid/invalid forms, duplicates, soft>hard), spec merging (replace/append/nil-process), and modifier construction.
- `go test ./internal/modifier/... ./api/config/... ./cmd/nvidia-ctk-installer/toolkit/... ./internal/runtime/...` passes.
- Mechanism verified against a Kubernetes 1.36 / containerd 2.3.3 cluster where non-root GPU pods previously failed RDMA memory registration with the inherited 8 MiB limit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/api/config/v1/runtime.go
+++ b/api/config/v1/runtime.go
@@ -27,6 +27,15 @@ type RuntimeConfig struct {
 	Runtimes []string    `toml:"runtimes"`
 	Mode     string      `toml:"mode"`
 	Modes    modesConfig `toml:"modes"`
+	// Rlimits defines POSIX resource limits to apply to the OCI runtime
+	// specification of containers handled by the NVIDIA Container Runtime.
+	// Each entry has the form NAME=SOFT[:HARD], where NAME is an rlimit name
+	// with or without the RLIMIT_ prefix (case-insensitive, e.g. "memlock" or
+	// "RLIMIT_MEMLOCK") and the values are non-negative integers or
+	// "unlimited"/"infinity". If HARD is omitted, it is set to SOFT.
+	// A configured entry replaces an rlimit of the same type already present
+	// in the incoming OCI specification.
+	Rlimits []string `toml:"rlimits,omitempty"`
 }
 
 // modesConfig defines (optional) per-mode configs
--- a/cmd/nvidia-ctk-installer/toolkit/toolkit.go
+++ b/cmd/nvidia-ctk-installer/toolkit/toolkit.go
@@ -63,6 +63,8 @@ type Options struct {
 
 	ContainerRuntimeRuntimes []string
 
+	ContainerRuntimeRlimits []string
+
 	ContainerRuntimeHookSkipModeDetection bool
 
 	ContainerCLIDebug string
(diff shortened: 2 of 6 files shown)


henry118:
@liuxu623 Thanks for the contribution. 

I don't think this belongs in `nvidia-container-runtime`.

`ulimit` is a generic runtime/process setting, while the toolkit should stay focused on nvidia GPU-related configs. And CDI doesn't provide a way to configure `rlimits` either.

For this use case, an NRI plugin would be a better fit for the timebeing until KEP-5758 lands.

henry118:
I'm closing this one as explained previously. Happy to discuss the use cases and other potential opportunities. Thanks.

PR #2040, opened 2026-08-31
bump runc to v1.5.1 and remove deprecated methods

tariq1890:
In runc v1.5.x, the `WithProcfd` file is no longer available to import so any dependabot bump PRs will fail with compilation errors.

The `utils.WithProcfd` method is no longer used and has been further simplified to use `os.Root.Open()` to perform safe file operations and rendering procfd paths directly as `/prof/self/fd/<fd-number>`. h/t @henry118 for this suggestion

Bonus: This PR also inlines the recommended `devices.DeviceFromPath` method call as defined in the `moby/sys/devices` go module.


henry118 (commented):
Thanks for working on this.

IIUC we don't really need the runc implementation here.

With golang 1.24+, I think we can achieve the same with os.Root + manual /proc/self/fd/N

tariq1890:
> With golang 1.24+, I think we can achieve the same with os.Root + manual /proc/self/fd/N

That's fair. Even less runc code to copy then!

internal/utils/utils.go
```diff
@@ -0,0 +1,42 @@
...
+	"os"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
+)
+
+func WithProcfd(root, unsafePath string, fn func(procfd string) error) error {
+	fullPath, err := securejoin.SecureJoin(root, unsafePath)
+	if err != nil {
+		return fmt.Errorf("resolving path inside rootfs failed: %w", err)
+	}
```
tariq1890: I just realised that this is not required since the `os.OpenInRoot` already ensures that the file path is within the root making the `SecureJoin` method redundant.

I will remove this

tariq1890: Confirmed, e2e tests are now failing with the error mentioned below:
```
 STDERR: docker: Error response from daemon: failed to create task for container: failed to create shim task: 
OCI runtime create failed: runc create failed: unable to start container process: 
error during container init: error running createContainer hook #3: exit status 1, stdout: , 
stderr: 2026/09/01 00:19:09 Error updating ldcache: error mounting host ldconfig: 
error creating tmpfs: open o_path procfd: openat /var/lib/docker/overlay2/b9ae33912aa181b9ede54dbd97f57247c6d046cc0c1e1ba38beb77056a6632a2/merged/run/nvidia-ctk-hook/5aa2cc30-37a2-4239-a221-61925b6cc492: path escapes from parent
```

tariq1890:
> IIUC we don't really need the runc implementation here.

@henry118 I've removed invocations of methods from the `libcontainers/utils` package in the `runc` go module. 

> With golang 1.24+, I think we can achieve the same with os.Root + manual /proc/self/fd/N

I am resolving the procfs path through the `manual /proc/self/fd/N`, but I didn't have much luck using the `os.OpenInRoot` method.; it kept failing with the `path escapes from parent` error

henry118 (approved):
LGTM

cmd/nvidia-cdi-hook/disable-device-node-modification/params_linux.go
```diff
@@ -25,55 +25,47 @@ import (
...
-	"github.com/opencontainers/runc/libcontainer/utils"
 	"golang.org/x/sys/unix"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/internal/utils"
 )
 
 func createParamsFileInContainer(containerRoot *os.Root, contents []byte) error {
-	containerRootDirPath := containerRoot.Name()
 
 	hookScratchDirPath := "/run/nvidia-ctk-hook" + uuid.NewString()
```
cdesiniotis, line 35: Potentially out of scope for this PR, but shouldn't this be `hookScratchDirPath := "/run/nvidia-ctk-hook/" + uuid.NewString()`? I think there is a typo here. In a running container, I see the below:
```
root@d2af80a529ad:/# ls -ltr /run/nvidia-ctk-hook*
/run/nvidia-ctk-hook:
total 4
drwxr-xr-x 2 root root 4096 Sep  1 22:20 2c3a944a-b848-4af1-b4b9-d6da24166562

/run/nvidia-ctk-hookb870042d-5b35-4c44-8ac8-ca3ac990d3c9:
total 4
-r--r--r-- 1 root root 1021 Sep  1 22:20 nvct-params
```
when in reality I am guessing we want:
```
/run/nvidia-ctk-hook/<uid-1>
/run/nvidia-ctk-hook/<uid-2>
```

tariq1890: Thanks for pointing this out. Agreed that we should be doing a `filepath.Join` instead of a string concatenate. As discussed, I can make this change in a follow-up PR.

cdesiniotis (approved):

PR #2045, opened 2026-09-03
chore: add PR template

cdesiniotis:
This addresses SIG-CON-05 and SIG-CON-11 from OSS scorecard

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,15 @@
+## Description
+
+<!-- Brief description of the change, including context or motivation -->
+
+## Checklist
+
+- [ ] No secrets, sensitive information, or unrelated changes
+- [ ] Unit tests passing (`make test`)
+- [ ] Lint checks passing (`make lint`)
+- [ ] Test cases are added for new code paths
+- [ ] Commits are signed-off and [cryptographically signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)
+
+## Testing
+
+<!-- How was this tested? e.g., unit tests, manual testing on a GPU node -->


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
...
+
+<!-- Brief description of the change, including context or motivation -->
+
+## Checklist
+
+- [ ] No secrets, sensitive information, or unrelated changes
+- [ ] Unit tests passing (`make test`)
+- [ ] Lint checks passing (`make lint`)
+- [ ] Test cases are added for new code paths
+- [ ] Commits are signed-off and [cryptographically signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)
```
tariq1890, line 11: Should we add something along the lines of "clean commit history" to the checklist?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Testing
+
+<!-- How was this tested? e.g., unit tests, manual testing on a GPU node -->
```
tariq1890: Add EOL

henry118 (approved):

.github/PULL_REQUEST_TEMPLATE.md
cdesiniotis: How would you suggest we word it? The bullets here should be precise imo.

.github/PULL_REQUEST_TEMPLATE.md
cdesiniotis: Updated.

.github/PULL_REQUEST_TEMPLATE.md
tariq1890: Agreed that we need a concise bullet point to express this. I can't think of anything right now, we can revisit this later.

tariq1890 (approved):

PR #2059, opened 2026-09-09
update-ldcache hook: exec ldconfig from a sealed memfd

klueska:
## Summary

`UpdateLDCache` used to mount a fresh `/proc` inside the container root, just so it could bind-mount the host `ldconfig` binary in through a TOCTOU-safe `/proc/self/fd/<n>` target and then exec it by that same path once pivoted.

This breaks under gVisor. Its `createContainer` hook doesn't have permission to mount a new `/proc` inside its sandbox, so `mountProc` fails with `EPERM`. This constitutes a real regression, because this hook worked fine before the changes described above were introduced.

To address this, `UpdateLDCache` now clones the host `ldconfig` binary into a sealed, anonymous `memfd` (via `libcontainer/exeseal.CloneBinary`) before touching any namespace or pivoting, then execs it by descriptor with `execveat(fd, "", ..., AT_EMPTY_PATH)`. A `memfd` has no path and no mount-namespace membership, so it comes through `pivot_root` untouched. We don't need to bind-mount anything into the container, and we don't need to mount `/proc` anywhere.

This is also more isolated than before. Since `ldconfig` has no use for `/proc` or `/sys`, `prepareRoot` now masks both with an empty `tmpfs` before pivoting, if they're present in the container root, instead of leaving a real working `procfs` mounted there and trusting `ldconfig` never reads it.

Given that this is a regression, I recommend backporting to `1.18.x`, `1.19.x`, and `1.20.x`

## Test plan

- [x] `go build ./internal/ldconfig/... ./cmd/nvidia-cdi-hook/...`
- [x] `go vet ./internal/ldconfig/...`
- [x] `go test ./internal/ldconfig/...`
- [x] Verified against real GPU hardware through gVisor's `runsc` runtime, as well as standard `runc`: `docker run --runtime=runsc --device nvidia.com/gpu=0` against a non-Debian image resolves the driver libraries and runs `nvidia-smi` correctly, confirming the hook and the `/proc`/`/sys` masking both work under gVisor's restricted-namespace gofer process.


internal/ldconfig/ldconfig_linux.go
```diff
@@ -171,6 +171,31 @@ func msMoveRoot(rootfs string) error {
...
+func maskPseudoFilesystems(root *os.Root) error {
+	for _, name := range []string{"proc", "sys"} {
+		if _, err := root.Stat(name); err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				continue
+			}
+			return fmt.Errorf("error checking %v: %w", name, err)
+		}
+		target := filepath.Join(root.Name(), name)
+		if err := unix.Mount("tmpfs", target, "tmpfs", 0, ""); err != nil {
```
henry118: The `unix.Mount` targets a plain joined path string, so there's a potential TOCTOU window. Is it a concern? 

Should we harden it with the following (existing) pattern?
```
f, err := root.Open(name)
if err != nil {
    if errors.Is(err, os.ErrNotExist) {
        continue
    }
   return err
}
err = unix.Mount("tmpfs", utils.GetProcFdPath(f), "tmpfs", 0, "")
```

henry118 (commented):
@klueska LGTM. just one nit comment/question.

internal/ldconfig/ldconfig_linux.go
klueska: Yes. This is a valid concern and in need of fixing. Thanks!

klueska: Updated.

internal/ldconfig/ldconfig.go
```diff
@@ -187,36 +188,47 @@ func (l *Ldconfig) UpdateLDCache() error {
...
+	return SafeExec(memfd, args, nil)
 }
 
-func (l *Ldconfig) prepareRoot() (string, error) {
-	root, err := os.OpenRoot(l.inRoot)
+// prepareRoot copies the host ldconfig binary's contents into a sealed,
+// anonymous memfd (an in-memory file with no path of its own) and pivots
+// into the container root, returning the memfd for UpdateLDCache to exec by
+// descriptor (see SafeExec).
+func (l *Ldconfig) prepareRoot() (_ *os.File, rerr error) {
```
tariq1890, line 198: nit: Is there any benefit to declaring the return arg here? If not, we can avoid the "underscoring out" of the file return arg by sticking to the conventional means of specifying a function signature.

```suggestion
func (l *Ldconfig) prepareRoot() (*os.File, error) {
```

internal/ldconfig/ldconfig_linux.go
```diff
@@ -171,6 +172,33 @@ func msMoveRoot(rootfs string) error {
...
+// container image itself) can never be read by anything running in the
+// isolated namespaces below, because there is nothing real there to read.
+func maskPseudoFilesystems(root *os.Root) error {
+	for _, name := range []string{"proc", "sys"} {
+		f, err := root.Open(name)
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				continue
+			}
+			return fmt.Errorf("error opening %v: %w", name, err)
```
tariq1890:
```suggestion
			return fmt.Errorf("error opening %s: %w", name, err)
```

internal/ldconfig/ldconfig_linux.go
```diff
@@ -171,6 +172,33 @@ func msMoveRoot(rootfs string) error {
...
+		}
+		defer f.Close()
+
+		if err := unix.Mount("tmpfs", utils.GetProcFdPath(f), "tmpfs", 0, ""); err != nil {
+			return fmt.Errorf("error masking %v: %w", name, err)
```
tariq1890:
```suggestion
			return fmt.Errorf("error masking %s: %w", name, err)
```

internal/ldconfig/ldconfig.go
```diff
@@ -187,36 +188,47 @@ func (l *Ldconfig) UpdateLDCache() error {
...
+	// the memfd survives the pivot below untouched.
+	memfd, err := cloneLdconfigIntoMemfd(l.ldconfigPath)
 	if err != nil {
-		return "", fmt.Errorf("failed to open root: %w", err)
+		return nil, fmt.Errorf("error cloning host ldconfig: %w", err)
 	}
-	defer root.Close()
+	defer func() {
+		if rerr != nil {
+			memfd.Close()
```
tariq1890, line 207: question - why do we close this only on error? What would happen if we closed it unconditionally instead?

internal/ldconfig/ldconfig.go
```diff
@@ -140,10 +140,11 @@ ldconfig in the container. Such differences include system search paths.`)
 }
 
 func (l *Ldconfig) UpdateLDCache() error {
-	ldconfigPath, err := l.prepareRoot()
+	memfd, err := l.prepareRoot()
 	if err != nil {
 		return err
 	}
+	defer memfd.Close()
```
tariq1890, line 147: won't this defer statement be called regardless of the outcome in `l.prepareRoot`?

internal/ldconfig/ldconfig.go
klueska: Yes. It's needed so we can check inside the defer to only close the memfd on error of the function not on success. This is a standard pattern used often.

internal/ldconfig/ldconfig.go
klueska: We can't close it unconditionally. On success we return it so the caller can make use of it and then it's his responsibility to close it.

internal/ldconfig/ldconfig.go
klueska: The alternative would be to explicitly call close in the if statement in every error path, but this is error prone especially if this function gets expanded in the future.

internal/ldconfig/ldconfig.go
klueska: No, if there is an error in prepareRoot() it won't be called. It's only called on success when the memfd is actually passed back.

klueska: Note that within `prepareRoot()` itself, it also won't be called if the creation of the memfd itself fails and was never available to close.

internal/ldconfig/ldconfig_linux.go
klueska: Updated.

internal/ldconfig/ldconfig_linux.go
klueska: Updated.

henry118 (approved):

internal/ldconfig/ldconfig.go
klueska: Note also that this is every error *after* the first one (i.e. the one that would result from no memfd being created at all).

tariq1890: Understood, thanks

tariq1890 (approved):

internal/ldconfig/ldconfig_linux.go
```diff
@@ -171,6 +172,33 @@ func msMoveRoot(rootfs string) error {
...
+		f, err := root.Open(name)
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				continue
+			}
+			return fmt.Errorf("error opening %s: %w", name, err)
+		}
+		defer f.Close()
+
+		if err := unix.Mount("tmpfs", utils.GetProcFdPath(f), "tmpfs", 0, ""); err != nil {
```
cdesiniotis, line 195: Note to self: the `internal/utils` package is on `main` but not on `release-1.20`. Will need to backport https://github.com/NVIDIA/nvidia-container-toolkit/pull/2040 as well.

cdesiniotis:
A note on compatibility: `memfd_create` was introduced in Linux 3.17 and glic 2.27. In libnvidia-container, we have a runtime fallback for old systems that do not support memfd: https://github.com/NVIDIA/libnvidia-container/commit/8fe3262b485da8d84bcfefc65b1fffb21d3ddb76. Should we consider something similar here?

klueska:
> A note on compatibility: `memfd_create` was introduced in Linux 3.17 and glic 2.27. In libnvidia-container, we have a runtime fallback for old systems that do not support memfd: [NVIDIA/libnvidia-container@8fe3262](https://github.com/NVIDIA/libnvidia-container/commit/8fe3262b485da8d84bcfefc65b1fffb21d3ddb76). Should we consider something similar here?

I'm not sure if its worth the trouble.

`glibc` shouldn't matter since all of the relevant code is is pure `go` (i.e. not `CGo` which would rely on glibc):
https://github.com/opencontainers/runc/blob/main/libcontainer/exeseal/cloned_binary_linux.go

And Linux 3.17 is "ancient" at this point for all practical purposes (it was released in October of 2014).

Willing to revisit if you feel strongly though.

What motivated the fallback in `libnvidia-container`? Was it the kernel or glibc (or something else)?

cdesiniotis:
> glibc shouldn't matter since all of the relevant code is is pure go (i.e. not CGo which would rely on glibc):
https://github.com/opencontainers/runc/blob/main/libcontainer/exeseal/cloned_binary_linux.go

Perfect, thanks for confirming this. I don't have any concerns then.

> What motivated the fallback in libnvidia-container? Was it the kernel or glibc (or something else)?

Looking back, it had to do with the glibc version during build time. We added the fallback to resolve centos7 build errors for libnvidia-container. We still build libnvidia-container (and the rest of the toolkit) in a centos7 image (which contains a version of glibc < 2.27) to maintain compatibility with a wide range of glibc versions.

PR #2075, opened 2026-09-10
feat: add new CDI hook to set CUDA memory limits

tariq1890:
This commit introduces a new subcommand of the nvidia CDI hook called apply-cuda-memory-limits. The purpose of this CDI hook is to apply the desired memory limits to a GPU device.

It reads the following envars from a container spec to fetch the input parameters for the CDI hook

- NVIDIA_GPU_MEMORY_REQUEST
- NVIDIA_GPU_MEMORY_LIMIT

Under the covers, the CDI hooks invokes the new NVML `SetMemoryLimits_v1` method (introduced in CUDA 13.4) to apply the desired CUDA memory limit values to the cgroup of the GPU container in question


cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
```diff
@@ -0,0 +1,185 @@
...
+	if len(candidates) == 0 {
+		return fmt.Errorf("no libnvidia-ml.so.1 found")
+	}
+
+	m.logger.Infof("driver library found: %s", candidates[0])
+
+	nvmllib := nvml.New(nvml.WithLibraryPath(candidates[0]))
+	ret := nvmllib.Init()
+	if ret != nvml.SUCCESS {
+		return fmt.Errorf("failed to initialize nvml: %v", ret)
```
cdesiniotis, line 165: nit: 
```suggestion
		return fmt.Errorf("failed to initialize nvml: %w", ret)
```

cdesiniotis: There are two other instances in this method where we are not wrapping errors.

cmd/nvidia-ctk/cdi/generate/generate_test.go
```diff
@@ -98,11 +98,35 @@ devices:
...
+                - --gpu-id
+                - {{ .gpuID }}
+              env:
+                - NVIDIA_CTK_DEBUG=false
     - name: all
       containerEdits:
         deviceNodes:
             - path: /dev/nvidia0
               hostPath: {{ .driverRoot }}/dev/nvidia0
+        hooks:
```
cdesiniotis, line 118: Would it be possible to add a test case for multiple devices? I would like to confirm that the `all` device contains one `apply-cuda-memory-limits` hook per GPU.

internal/info/cgroup/cgroup_path.go
```diff
@@ -0,0 +1,137 @@
...
+	err := unix.Statfs("/sys/fs/cgroup", &s)
+	if err != nil {
+		panic(err)
+	}
+	return s.Type == v2FsMagicNumber
+}
+
+func GetAbsolutePath(spec specs.Spec) (string, error) {
+	c := &cgroups.Cgroup{}
+	myCgroupPath := spec.Linux.CgroupsPath
```
henry118, line 48: The `CgroupsPath` in the runtime spec is optional. I'm not sure we can get the cgroup path reliably this way.

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
```diff
@@ -0,0 +1,185 @@
...
+
+	for _, gpuID := range gpuIDs {
+		device, ret := nvmllib.DeviceGetHandleByUUID(gpuID)
+		if ret != nvml.SUCCESS {
+			return fmt.Errorf("failed to get GPU device handle with uuid %s: %v", gpuID, ret)
+		}
+		if device == nil {
+			return fmt.Errorf("empty GPU device handle: %s", gpuID)
+		}
+		ret = device.SetMemoryLimits_v1(cgroupPath, requests*1024*1024, limits*1024*1024)
```
henry118, line 179: Do we know what cgroup namespace the NVML expects (host or container)? The cgroup path here is the container's cgroup namespace.

internal/info/cgroup/cgroup_path.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	// an example of the final path in rootless:
+	// "/sys/fs/cgroup/user.slice/user-1001.slice/user@1001.service/user.slice/libpod-132ff0d72245e6f13a3bbc6cdc5376886897b60ac59eaa8dea1df7ab959cbf1c.scope"
+
+	return path, nil
+}
+
+func getSliceFull(c *cgroups.Cgroup) (string, error) {
+	slice := "system.slice"
```
henry118, line 88: Can we always assume systemd here?

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
tariq1890: Fair. I guess I could use `%w` considering that `nvml.Return` implements the `Error` interface of Go

cmd/nvidia-ctk/cdi/generate/generate_test.go
tariq1890: I can definitely look into this

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
tariq1890: > Do we know what cgroup namespace the NVML expects (host or container)? 

Do you mean "pod or container"?

> The cgroup path here is the container's cgroup namespace.

This is expected. We want the memory limits to be scoped to a container

internal/info/cgroup/cgroup_path.go
tariq1890: So I am still thinking if we need to support `cgroupfs` as the cgroup manager. I could either error out saying that `only systemd is supported as the cgroups manager` or implement support for `cgroupfs`.

fwiw, the quick research I've done suggests that the use of `cgroupfs` as the cgroup manager for cgroups v2 is extremely rare. Note that this CDI hook will only  support cgroups v2

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
henry118: I get that. But should NVML be called in the host's cgroup namespace or container's?

internal/info/cgroup/cgroup_path.go
henry118: we don't need to support cgroupv1 (it's already deprecated). i just wonder if it's a good idea to always assume systemd here

cdesiniotis: IMO it is okay to only support systemd cgroups initially and document it. We can always add cgroupfs support later in this hook if we deem it necessary. We aren't breaking existing workloads. A container will only fail if `NVIDIA_GPU_MEMORY_REQUEST` / `NVIDIA_GPU_MEMORY_LIMIT` envvars are set and `cgroupfs` is used instead of `systemd`.

henry118: IIRC docker's default cgroup driver is not systemd though. and containerd's default cgroup driver isn't systemd either.

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
henry118: Oh, a correction - the cgroup path value in the runtime spec is actually relative to the cgroup root as seen by the runtime process - so it's essentially the host's cgroup namespace (not the container's)

internal/info/cgroup/cgroup_path.go
tariq1890: That's a good point. Let me look into cgroupfs support

tariq1890: Actually containerd 2.x and newer have systemd as the cgroup manager by default.

I am not sure about docker and podman

henry118: Did I miss anything? 
https://github.com/containerd/containerd/blob/f6132dbe1f482cbe0aebc4bd3d8d7a184fb4a2aa/internal/cri/config/config_unix.go#L69-L70

cdesiniotis: From https://docs.docker.com/engine/containers/runmetrics/#running-docker-on-cgroup-v2:
```
The default cgroup driver (dockerd --exec-opt native.cgroupdriver) is systemd on v2, cgroupfs on v1.
```

From https://kubernetes.io/docs/concepts/architecture/cgroups/#requirements:
```
cgroup v2 has the following requirements:
. . . 
- The kubelet and the container runtime are configured to use the [systemd cgroup driver](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#systemd-cgroup-driver)
```

From what I can tell, using systemd for cgroup management is recommend (and stated as required in k8s) for cgroupsv2 hosts.

cdesiniotis: From https://docs.podman.io/en/v4.9.0/markdown/podman.1.html:
```
The CGroup manager to use for container cgroups. 
Supported values are cgroupfs or systemd. 
Default is systemd unless overridden in the containers.conf file.

```

tariq1890: Thanks @cdesiniotis !

tariq1890: @henry118 

I could be wrong, but from what I understand of this [PR](https://github.com/containerd/containerd/pull/8722), containerd will resolve the cgroup driver based on the environment it is in. Since Kubernetes/kubelet has `systemd` as the default cgroup manager, containerd will pick that up and resolve the cgroup manager for itself as the same.

internal/info/cgroup/cgroup_path.go
tariq1890: > The CgroupsPath in the runtime spec is optional

I was not aware that this is actually an optional field in the OCI spec. Using `.spec.Linux.CgroupsPath` was the only way for me to actually resolve the full cgroup namespace path which becomes the valid input for the `nvml.SetMemoryLimits_v1` symbol.

If there is a better approach, I am certainly open to it.

internal/info/cgroup/cgroup_path.go
henry118: Regardless I don't see why we need to be concerned with cgroup driver. All we need is the cgroup path to pass into the NVML API.

Note that in the current implementation, the hook runs as `createContainer` hook, which runs in the container's cgroup namespace. But the path we derive from the runtime spec is the from host cgroup namespace. 

There are a few problems:
* The cgroup path in the spec is optional and cannot be depend upon;
* If NVML expects to be invoked within the container's cgroup namespace, this is not a legit value to use;
* Otherwise `createContainer` does not sound like the right mechanism.

tariq1890: > Otherwise createContainer does not sound like the right mechanism.

It should be `createRuntime`. Agreed that `createContainer` is not the right phase. Looks like I forgot to make that change in this PR. Thanks for pointing that out

henry118: One more thing worth considering wrt cgroup ns. normally for cpu/memory, containers are not allowed to adjust their cgroup settings from within their own cgroup ns (for obvious reasons), hence it's the runtime that performs the configs. but i'm not sure about GPUs.

tariq1890: @henry118 It runs as a `createRuntime` hook. See [here](https://github.com/NVIDIA/nvidia-container-toolkit/blob/fcca9a9c8c6bb6800a34707a1dc889b6baf031b3/internal/discover/hooks.go#L228)

internal/info/cgroup/cgroup_path.go
tariq1890: How does runc resolve the cgroup namespace absolute path when `Spec.Linux.CgroupsPath` is empty?

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
```diff
@@ -100,10 +100,6 @@ func (m command) run(_ *cli.Command, cfg *config) error {
 	}
 	specFilePath := oci.GetSpecFilePath(s.Bundle)
 	fs := oci.NewFileSpec(specFilePath, false)
-	ctrSpec, err := fs.Load()
-	if err != nil {
-		return fmt.Errorf("failed to load OCI container spec: %w", err)
-	}
```
cdesiniotis: @tariq1890 do we still need to load the spec here?

tariq1890: Ah, good catch. I didn't expect that the method would both i) return a value and ii) populate the function receiver variable.

```
func (s *fileSpec) Load() (*specs.Spec, error) {
	specFile, err := os.Open(s.path)
	if err != nil {
		return nil, fmt.Errorf("error opening OCI specification file: %v", err)
	}
	defer specFile.Close()

	spec, err := s.loadFrom(specFile)
	if err != nil {
		return nil, fmt.Errorf("error loading OCI specification from file: %v", err)
	}
	s.Spec = spec
	return s.Spec, nil
}
```

cmd/nvidia-ctk/cdi/generate/generate_test.go
cdesiniotis: Not a blocker, confirmed that the `all` device contains one hook per GPU.

cdesiniotis (approved):

internal/info/cgroup/cgroup_path.go
tariq1890: Resolving this thread as the cgroup absolute path resolution logic no longer assumes `systemd` as the cgroups manager.

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
tariq1890: > But should NVML be called in the host's cgroup namespace or container's?

NVML expects the host's cgroup namespace. The new subcommand is intended to run as a `createRuntime` hook

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
```diff
@@ -0,0 +1,219 @@
...
+func (m command) run(_ *cli.Command, cfg *config) error {
+	s, err := oci.LoadContainerState(cfg.containerSpec)
+	if err != nil {
+		return fmt.Errorf("failed to load container state: %w", err)
+	}
+	specFilePath := oci.GetSpecFilePath(s.Bundle)
+	fs := oci.NewFileSpec(specFilePath, false)
+	_, err = fs.Load()
+	if err != nil {
+		m.logger.Warningf("Skipping CUDA memory limits: failed to load OCI container spec: %v", err)
```
tariq1890: @cdesiniotis @henry118 I've made this a silent failure so that it handles containers with `--userns nomap` (see issue #648). Without this, containers with `--userns nomap` will fail even if they don't set the gpu memory limit env vars.

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
```diff
@@ -0,0 +1,219 @@
...
+		return nil
+	}
+
+	memReqStr, hasRequest := fs.LookupEnv(GPUMemoryRequestEnvName)
+	memLimitStr, hasLimit := fs.LookupEnv(GPUMemoryLimitEnvName)
+	if !hasRequest && !hasLimit {
+		return nil
+	}
```
tariq1890, line 125: @cdesiniotis this logic has been updated since the last approve. In the latest iteration, setting both the request and limits are no longer mandatory; either one env var being set is enough for the the CDI hook to call the NVML Memory Limits API.

The new behaviour is as follows

i) If both env vars are set, it continues to work as before
ii) If only the request env var is set, the limit is set to the max-possible uint64 values
iii) If only the limit env var is set, then the request is set to the same value as the limit.

cmd/nvidia-cdi-hook/apply-cuda-memory-limits/apply-cuda-memory-limits.go
cdesiniotis: Note that `createRuntime` hooks should not be impacted by https://github.com/NVIDIA/nvidia-container-toolkit/issues/648 as they do not run in the container's namespace. So I do not think it is expected that this hook would fail even if a container runs in its own userns with uid/gids remapped.

tariq1890: That is fair.

tariq1890: I've reverted this back to failing fast.

cdesiniotis (approved):

internal/info/cgroup/cgroup_path.go
tariq1890: This is resolved as we determine the cgroup absolute path using the `/proc/<pid>/cgroup` file instead.

henry118 (approved):

PR #2087, opened 2026-09-13
internal/nvsandboxutils: Look up entry points with dlsym

rautyrauty:
## Description

The generated bindings reference the `libnvidia-sandboxutils` entry points at link time and link with `-Wl,--unresolved-symbols=ignore-in-object-files`, so every binary using them keeps five unresolved dynamic symbols, bound
lazily against the library `lib.go` dlopens with `RTLD_GLOBAL`.

Two consequences:

- Distributions that check packages for unresolved ELF symbols reject such binaries on platforms where the library is not packaged at all. This breaks LXD (which vendors `pkg/nvcdi`) on ALT Linux i586 and aarch64, where nothing
provides `libnvidia-sandboxutils.so.1`:
```
i586: NEW bad_elf_symbols detected:
lxd-6.9-alt1.i586.rpm /usr/bin/lxd U nvSandboxUtilsInit
```
- A call made before a successful `Init()` terminates the process with a dynamic linker symbol lookup error instead of returning a code.

The entry points are now resolved with `dlsym(RTLD_DEFAULT)` — the same global scope the dynamic linker searched, so a loaded library is found exactly as before, and an absent one yields `NVSANDBOXUTILS_ERROR_LIBRARY_LOAD`.

The wrappers are a plain `.c`/`.h` pair pulled into the preamble via `Includes` in the c-for-go configuration, so regenerating the bindings keeps them; the generated files change only as that configuration dictates. `nvml*`
symbols from `go-nvml` are untouched — the same treatment there would span several hundred generated functions and is better done separately.

## Checklist

- [x] No secrets, sensitive information, or unrelated changes
- [x] Unit tests passing (`make test`)
- [x] Lint checks passing (`make lint`)
- [ ] Test cases are added for new code paths
- [x] Commits are signed-off and [cryptographically signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

## Testing

- `make test` passes; `golangci-lint run ./internal/nvsandboxutils/...` reports 0 issues.
- `nm -D --undefined-only nvidia-ctk | grep nvSandboxUtils`: 5 symbols before, 0 after.
- Manually against a stub `libnvidia-sandboxutils.so.1`: `Init`, `GetDriverVersion` and `Shutdown` return the stub's values; with no library present `Init` returns `ERROR_LIBRARY_LOAD` instead of crashing.
- Regenerated with c-for-go to confirm the generated preamble and call sites match this diff.
- Rebuilt the ALT Linux `lxd` package with these bindings vendored: the repository-wide ELF symbols check, which previously reported five new `nvSandboxUtils*` symbols, now passes on x86_64, i586 and aarch64, and the resulting
i586 binary has no unresolved `nvSandboxUtils*` left.
- No GPU node was available; the runtime path was exercised with the stub library only.


rajatchopra (changes requested):
Useful PR.
Can we have a mock_nvsandbox.so and build a unit test case around the five calls?

rautyrauty:
Added in a follow-up commit: testdata/mock_nvsandbox.c is compiled and dlopen'ed by the test, so all five entry points go through the real dlsym path - on main it fails with undefined symbol: nvSandboxUtilsShutdown.

PR #2103, opened 2026-09-16
fix: [apply-cuda-memory-limits] exit gracefully on oci spec load failure

tariq1890:
With this fix, containers started with "usens --nomap" no longer fail as the OCI spec load operation is no longer non-blocking. This is required as the containers with "users --nomap" fail even if they don't set the cuda memory limits environment variables.


cdesiniotis (approved):

tariq1890:
@cdesiniotis Thanks for the approve. Just fixed some typos in the commit description.

henry118 (commented):
curious what led this change? previously we didn't expect the permission issue...

tariq1890:
After running a podman container with `--userns nomap`, we observed this error. The failure is especially bad because the container does not set the `NVIDIA_GPU_MEMORY_REQUEST|LIMIT` env vars.

```
~$ podman run --rm --security-opt=label=disable --device=nvidia.com/gpu=all --userns nomap ubuntu nvidia-smi
Error: OCI runtime error: crun: error executing hook `/usr/bin/nvidia-cdi-hook` (exit code: 1)

~$ cat nvidia-cdi-hook.log
time="2026-09-16T16:51:39Z" level=error msg="failed to load OCI container spec: error opening OCI specification file: open /config.json: no such file or directory"
```

henry118:
> open /config.json: no such file or directory"

This appears to be a wrong path rather than permission issues?

henry118:
Opened PR https://github.com/containers/crun/pull/2261 which should address this issue.

PR #2107, opened 2026-09-18
Reject config paths that descend into non-struct fields

ajavanma:
## Description

Reject `config --set` keys that continue past a non-struct field. A key such as `disable-require.extra=true` currently reaches a boolean during recursive TOML-field lookup and panics. Checking the current type before enumerating fields returns the existing wrapped invalid-option error instead.

Add parser and real command-handler regressions for boolean, string, slice, and pointer fields. Tests also verify that invalid input preserves the configuration file and that valid nested paths, list values, and optional boolean values still work.

## Checklist

- [x] No secrets, sensitive information, or unrelated changes
- [x] Unit tests passing (`make test`)
- [x] Lint checks passing (`make lint`)
- [x] Test cases are added for new code paths
- [x] Commits are signed-off and [cryptographically signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

## Testing

- Saved audit reproduction fails on unchanged main `3a6c050d9c0c7b2516ce17e2190fb7e71b741f82`.
- `go test ./cmd/nvidia-ctk/config -run '^TestConfigCommandSet$' -count=1`: four malformed-path cases fail on unchanged production code; valid controls pass.
- `go test ./cmd/nvidia-ctk/config -count=1`: passes with the fix.
- `make fmt`, `make build`, `make test`: pass on Linux ARM64 with Go 1.26.1.
- `make lint`: passes with golangci-lint 2.13.2, zero issues.
- Built `nvidia-ctk` binary against temporary files: all four malformed paths return exit 1 with invalid-option errors, no panic, and unchanged file contents. A valid multi-set command preserves nested strings, serializes lists, and removes an optional boolean set to nil.
- No GPU/container-runtime end-to-end checks run; this change affects configuration-key validation only. No dependency changes.


myeolenv:
Thanks for the contribution. We will review this PR soon.

